PR #3, opened 2019-11-13
wording fixes and setting scaffolding

jeffhollan:
Let me know your thoughts on this @tomkerkhove - made a few wording changes but also making space for a contributor guide and replacing getting started and merging with deploy


jeffhollan:
Actually in hindsight I do think there should be a top level Overview doc that explains a bit of what KEDA is, and how it has a few different ways to work (Deployments and Jobs)

tomkerkhove:
> Actually in hindsight I do think there should be a top level Overview doc that explains a bit of what KEDA is, and how it has a few different ways to work (Deployments and Jobs)

That is correct, but I'm still working on that - Just wanted to get the foundational structure in first.

Will open issues for what's still open later today

content/_index/scalers.md
```diff
@@ -4,5 +4,5 @@ weight = "200"
 
 background = "light"
 title = "Scalers"
-subtitle = "Syna comes with a lot of fragments, allowing you to design your pages as you want them"
+subtitle = "Scalers represent event sources that KEDA can scale based on"
```
tomkerkhove: Thanks 😅

content/contributing.md
```diff
@@ -1,6 +1,6 @@
 +++
-title = "Getting Started KEDA"
-date = "2017-10-05"
+title = "Contributing to KEDA"
```
tomkerkhove: This document is actually going to introduce new people to KEDA - What it is, how it could help, etc.

In terms of how to contribute, I would rather keep it out of this documentation page and just point to the contributing.md in our GitHub repo and keep them next to each other.

content/_index/samples.md
```diff
@@ -0,0 +1,12 @@
++++
+fragment = "content"
+#disabled = true
+date = "2019-11-11"
+weight = 300
+background = "light" # can influence the text color
+particles = true
++++
+
+### Samples
```
tomkerkhove: I'm not seeing this pop-up, did you run this locally?

If so, where does it show up?

jeffhollan: It’s now showing near bottom of home page above the community call section but if you’re working on an overview would make sense to be there

content/contributing.md
jeffhollan: Yep sounds good

jeffhollan:
Ok great. I’ll keep changed but revert the getting started one. I do wonder if we should rename getting started to overview as I think of getting started like “how do I deploy”

tomkerkhove:
> Ok great. I’ll keep changed but revert the getting started one. I do wonder if we should rename getting started to overview as I think of getting started like “how do I deploy”

Well it would be part of it, let's think of it like this:
1. Installing KEDA
2. Deploying deployment
3. Deploying the trigger auth
4. Deploying the scaled object
5. Monitor with KEDA dashboard

content/_index/samples.md
tomkerkhove: Ah now I see it, that's fine for me!

Over time we could curate a list of 4-6 entries which are highlighted.

content/contributing.md
tomkerkhove: Thanks! Once that is changed it looks good to me!

jeffhollan:
Ok sounds good. And are you working on that one?

tomkerkhove:
I'm happy to but not there yet, opened https://github.com/kedacore/keda-docs/issues/10 for it.

jeffhollan:
Ok if I stumble upon time tonight and you haven’t done more I’ll take a stab at at. And are you thinking it’s concept or actually step by step QuickStart “run this, then look at that” or just “when monitoring you’d do Kubectl on the KEDA operator (by default running in the KEDA namespace”

tomkerkhove:
Frankly, haven't put too much though in it but for deployment I'd point to the deployment docs for example and otherwise show then how to run things.

This is a good example imo: https://docs.konghq.com/1.4.x/getting-started/quickstart/

That said, I would provide a dedicated page for scaledobject and triggerauth specifically but have to do this later this week(end)/next week.

PR #4, opened 2019-11-13
Provide first batch of scaler documentation

tomkerkhove:
Provide first batch of scaler documentation.

Rest will follow and KEDA repo will be cleaned up to get rid of deprecated data.

### DIFF
--- a/content/scalers/aws-cloudwatch.md
+++ b/content/scalers/aws-cloudwatch.md
@@ -1,0 +1,68 @@
++++
+fragment = "content"
+weight = 100
+title = "AWS Cloudwatch"
+background = "light"
++++
+
+Scale applications based on a AWS Cloudwatch.
+
+**Availability:** v1.0+ | **Maintainer:** Community
+
+<!--more-->
+
+### Trigger Specification
+
+This specification describes the `aws-cloudwatch` trigger that scales based on a AWS Cloudatch.
+
+```yaml
+triggers:
+  - type: aws-cloudwatch
+    metadata:
+      # Required: namespace
+      namespace: AWS/SQS
+      # Required: Dimension Name
+      dimensionName: QueueName
+      dimensionValue: keda
+      metricName: ApproximateNumberOfMessagesVisible
+      targetMetricValue: "2"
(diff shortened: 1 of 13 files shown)


content/scalers/aws-cloudwatch.md
```diff
@@ -0,0 +1,68 @@
...
+      # Required: Dimension Name
+      dimensionName: QueueName
+      dimensionValue: keda
+      metricName: ApproximateNumberOfMessagesVisible
+      targetMetricValue: "2"
+      minMetricValue: "0"
+      # Required: region
+      awsRegion: "eu-west-1"
+      # Optional: AWS Access Key ID
+      awsAccessKeyID: AWS_ACCESS_KEY_ID # default AWS_ACCESS_KEY_ID
```
zach-dunton-sf: This is still the old authentication method. Is the TriggerAuthentication going to be merged before this?

jeffhollan: @zach-dunton-sf great callout. I believe so.  Let me check.

//cc @ahmelsayed

jeffhollan: Oh and to clarify I believe the way the TriggerAuthentication was built is that it can be backwards compatible. So can either define inline or in the seperate CRD.  We can decide if we want to include samples for both flavors in each scaler doc

tomkerkhove: @zach-dunton-sf You are correct, I'm doing it step by step and will send a PR to include the trigger auth part.

@jeffhollan I don't think that's the case and frankly I think that's the best approach to only use CRDs as it makes it simpler to manage. Just to clarify, do we support backwards compat @ahmelsayed ?

tomkerkhove: Actually I do remember this now vaguely so you are most probably right @jeffhollan  🤔

tomkerkhove: @zach-dunton-sf Submitted a PR https://github.com/kedacore/keda-docs/pull/11

PR #11, opened 2019-11-13
Provide auth parameters for AWS CloudWatch

tomkerkhove:
Provide auth parameters for AWS CloudWatch.

Relates to https://github.com/kedacore/keda/issues/333
Relates to https://github.com/kedacore/keda/issues/293
Relates to https://github.com/kedacore/keda/issues/303


tomkerkhove:
@jeffhollan This would be the approach I'd like to take on every scaled detail page

PR #12, opened 2019-11-13
Proposal for CODEOWNERS

tomkerkhove:
Let me know if you agree so we can automatically assign reviews.

@ahmelsayed I don't want to force you to be part of it but you're the main contributor so might make sense.


ahmelsayed:
Thanks @tomkerkhove! Looks good to me

PR #14, opened 2019-11-13
WIP - Provide documentation on scaling deployment, jobs & what TriggerAuthentication is

tomkerkhove:
Introduces :
- Deployment `keda.sh/workloads/deployments/`
- Jobs - `keda.sh/workloads/jobs/`
- General page to be added

PR #17, opened 2019-11-14
Creating doc structure

jeffhollan:


jeffhollan:
@tomkerkhove I refactored the site a bit to have a more structured docs section.  My thinking is things like Getting Started and Architecture would link there, and also give us pages to go into things like scaling deployments and jobs.  Check it out and let me know your thoughts.  This would replace https://github.com/kedacore/keda-docs/pull/14

tomkerkhove:
I like this approach:
image

Since the whole website are docs, maybe we should rename it to `Concepts` or `Basics` as that's more what it covers rather than "docs", no?

tomkerkhove:
Nitpick - Can we change the URL from `https://deploy-preview-17--vigorous-franklin-8edd65.netlify.com/docs/scaling_deployments/` to `https://deploy-preview-17--vigorous-franklin-8edd65.netlify.com/docs/scaling-deployments/`?

jeffhollan:
@tomkerkhove makes sense. Let’s go with “concepts” instead of docs and also good with the hyphenated url

PR #28, opened 2019-11-15
WIP - Display list of supported scalers on workload docs

tomkerkhove:
Display list of supported scalers on workload docs to make them more easily discoverable as they are only on our landing page for now.

### DIFF
--- a/content/concepts/scaling-deployments/scaler-header.md
+++ b/content/concepts/scaling-deployments/scaler-header.md
@@ -1,0 +1,8 @@
++++
+fragment = "header"
+weight = "200"
+
+background = "light"
+title = "Scalers"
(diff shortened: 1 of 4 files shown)


tomkerkhove:
I'll have to get this sorted:
image

jeffhollan:
Yeah I played with this a bit before and I don’t reallly want scalers showing at bottom of every docs page. If we create a dedicated section for scalers like we have for auth I’m cool with it but couldn’t figure out how to do that. Other than that would prefer to just like to root page list.

jeffhollan:
Ah looks like you only are trying to show in 2 sections. I’m cool with that if you can get it working

tomkerkhove:
It's just an idea, having them on landing page is ok for me as well!

Let's maybe focus on other things instead of this then?

PR #32, opened 2019-11-15
Update helm instructions for 1.0.0

ahmelsayed:


tomkerkhove:
And this broke our hub.helm.sh integration :)

Tracking via https://github.com/kedacore/charts/issues/3

Issue #31, opened 2019-11-15
Add nav bar and footer for samples 


jeffhollan:
Thinking we add one more link for samples that links to the samples repo on GitHub?

Thoughts @tomkerkhove

tomkerkhove:
Sounds good to me.

I was also wondering if we could "highlight" 3-6 samples underneath "Scalers" on the landing page with a similar flow?

PR #35, opened 2019-11-16
samples link and list

jeffhollan:
resolves https://github.com/kedacore/keda-docs/issues/31

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -45,7 +45,7 @@ lastmod = ["lastmod", ":git", "date"]
 
   [params.colors]
     primary = "#00838F"
-    secondary = "#868e96"
+    secondary = "#D3DBE3"
     success = "#008f54"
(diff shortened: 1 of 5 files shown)


tomkerkhove:
LGTM

One question though - Should we add a. NET sample without functions to show you can do that? Not a must though

jeffhollan:
@tomkerkhove thought about that too. Know I want the rabbitmq cause it’s simple, functions cause it highlights that, 3rd one I wasn’t sure. I went with the one that shows how you can do jobs, but maybe .NET is more applicable? I’m ok either way.

tomkerkhove:
Let's just stick with this then but maybe add a line before the samples where all the samples are aggregated (even though it's in the nav)

Happy to PR if you want

jeffhollan:
Great idea. As like a subtitle

Issue #25, opened 2019-11-15
Provide docs on how to authenticate with TriggerAuthentication


tomkerkhove:
Provide docs on how to authenticate with TriggerAuthentication

PR #39, opened 2019-11-17
adding auth spec and auth concepts

jeffhollan:
resolves https://github.com/kedacore/keda-docs/issues/25

### DIFF
--- a/content/concepts/authentication/content.md
+++ b/content/concepts/authentication/content.md
@@ -69,7 +69,7 @@ If you have multiple containers in a deployment, you will need to include the na
 
 > TIP: If creating a deployment yaml that references a secret, be sure the secret is created before the deployment that references it, and the scaledObject after both of them to avoid race conditions.
 
-While this method works for many scenarios, there are some downsides.  With the container reference approach, it means every reference the `ScaledObject` needs must be included on the container.  It also doesn't support a model where other types of authentication may work - namely "pod identity" where access to a source could be acquired with no secrets or connection strings.  For these and other reasons, we also provide a `TriggerAuthentication` resource to define authentication as a separate resource to a `ScaledObject`, which can reference secrets directly or supply configuration like pod identiy.
+While this method works for many scenarios, there are some downsides.  With the container reference approach, it means every reference the `ScaledObject` needs must be included on the container.  It also makes sharing identity across many ScaledObjects more difficult.  It also doesn't support a model where other types of authentication may work - namely "pod identity" where access to a source could be acquired with no secrets or connection strings.  For these and other reasons, we also provide a `TriggerAuthentication` resource to define authentication as a separate resource to a `ScaledObject`, which can reference secrets directly or supply configuration like pod identiy.
 
 ## TriggerAuthentication
 
@@ -95,6 +95,19 @@ spec:
 ```
 
 Based on the requirements you can mix and match the reference types providers in order to configure all required parameters.
+
+The parameters you define in the `TriggerAuthentication` definition do not need to be included in the `metadata` of the `ScaledObject.spec.triggers` definition.  To reference a `TriggerAuthentication` from a `ScaledObject` you add the `authenticationRef` to the trigger.
+
+```yaml
+# some Scaled Object
+# ...
+  triggers:
+  - type: {scaler-type}
+    metadata:
+      param1: {some-value}
+    authenticationRef:
+      name: {trigger-authencation-name} # this may define other params not defined in metadata
+```
 
 ### Environment variable(s)
 
--- a/content/concepts/authentication/content.md
+++ b/content/concepts/authentication/content.md
@@ -8,10 +8,142 @@ title = "Authentication"
   sticky = true
 +++
 
-### Referencing secrets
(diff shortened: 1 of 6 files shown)


content/concepts/authentication/content.md
```diff
@@ -8,10 +8,155 @@ title = "Authentication"
   sticky = true
 +++
 
-### Referencing secrets
+Often a scaler will require authentication or secrets and config in order for KEDA to check for events and monitor the event rate.  KEDA provides a few secure patterns to add those secrets to a `ScaledOjbect`.
```
jeffhollan:
```suggestion
Often a scaler will require authentication or secrets and config to check for events.  KEDA provides a few secure patterns to add authentication to a `ScaledOjbect`.
```

content/concepts/authentication/content.md
```diff
@@ -8,10 +8,155 @@ title = "Authentication"
...
 
-TODO
+## Referencing secrets and config maps
 
-### TriggerAuthentication
+Some metadata parameters will not allow resolving from a literal value, and will instead require a reference to a secret, config map, or environment variable defined on the target deployment.
```
jeffhollan:
```suggestion
Some metadata parameters will not allow resolving from a literal value, and will instead require a reference to a secret, config map, or environment variable defined on the target container.
```

content/concepts/authentication/content.md
```diff
@@ -8,10 +8,155 @@ title = "Authentication"
...
 
-TODO
\ No newline at end of file
+For example, if using the [RabbitMQ scaler](https://keda.sh/scalers/rabbitmq-queue/), the `host` parameter may include passwords so is required to be a reference.  You can create a secret with the value of the `host` string, mount that secret to the deployment, and reference it from the `ScaledObject` like below:
```
jeffhollan:
```suggestion
For example, if using the [RabbitMQ scaler](https://keda.sh/scalers/rabbitmq-queue/), the `host` parameter may include passwords so is required to be a reference.  You can create a secret with the value of the `host` string, reference that secret in the deployment, and map it to the `ScaledObject` metadata parameter like below:
```

content/concepts/authentication/content.md
```diff
@@ -8,10 +8,155 @@ title = "Authentication"
...
+  - type: rabbitmq
+    metadata:
+      queueName: hello
+      host: {secret-key-name}
+      queueLength  : '5'
+```
+
+If you have multiple containers in a deployment, you will need to include the name of the container that has the references in the `ScaledObject`.  If you do not include a `containerName` it will default to the first container.  KEDA will attempt to resolve references from secrets, config maps, and environment variables of the container.
+
+> TIP: If creating a deployment yaml that references a secret, be sure the secret is created before the deployment that references it, and the scaledObject after both of them to avoid race conditions.
```
jeffhollan:
```suggestion
> TIP: If creating a deployment yaml that references a secret, be sure the secret is created before the deployment that references it, and the scaledObject after both of them to avoid invalid references.
```

content/concepts/authentication/content.md
```diff
@@ -8,10 +8,155 @@ title = "Authentication"
...
+
+While this method works for many scenarios, there are some downsides.  With the container reference approach, it means every reference the `ScaledObject` needs must be included on the container.  It also makes sharing identity across many ScaledObjects more difficult.  It also doesn't support a model where other types of authentication may work - namely "pod identity" where access to a source could be acquired with no secrets or connection strings.  For these and other reasons, we also provide a `TriggerAuthentication` resource to define authentication as a separate resource to a `ScaledObject`, which can reference secrets directly or supply configuration like pod identiy.
```
jeffhollan:
```suggestion
While this method works for many scenarios, there are some downsides.  This method makes it difficult to efficiently share auth config across `ScaledObjects`.  It also doesn’t support referencing a secret directly, only secrets that are referenced by the container.  This method also doesn't support a model where other types of authentication may work - namely "pod identity" where access to a source could be acquired with no secrets or connection strings.  For these and other reasons, we also provide a `TriggerAuthentication` resource to define authentication as a separate resource to a `ScaledObject`, which can reference secrets directly or supply configuration like pod identity.
```

content/concepts/authentication/content.md
```diff
@@ -8,10 +8,155 @@ title = "Authentication"
...
+
+## TriggerAuthentication
+
+`TriggerAuthentication` allows you to define once how to authenticate and use it for multiple triggers across different teams, without them knowing where the secrets are.  It also enables more advanced methods of authentication like "pod identity."
```
jeffhollan:
```suggestion
`TriggerAuthentication` allows you to describe authentication parameters separate from the `ScaledObject` and the deployment containers.  It also enables more advanced methods of authentication like "pod identity."
```

content/concepts/authentication/content.md
```diff
@@ -8,10 +8,155 @@ title = "Authentication"
...
+
+```yaml
+apiVersion: keda.k8s.io/v1alpha1
+kind: TriggerAuthentication
+metadata:
+  name: {trigger-auth-name}
+  namespace: keda
```
jeffhollan:
```suggestion
  namespace: default # must be same namespace as the ScaledObject
```

PR #40, opened 2019-11-18
removing references to dashboard

jeffhollan:
I'm fairly confident with the latest changes of v1.0 and operator SDK the dashboard is non-functional.  Created an issue to track with help wanted flag

https://github.com/kedacore/dashboard/issues/14

@Aarthisk in case she can confirm or not if that makes sense

### DIFF
--- a/content/_index/features/dashboard.md
+++ b/content/_index/features/dashboard.md
@@ -1,9 +1,0 @@
-+++
-title = "Dashboard"
-weight = 60
-
-[asset]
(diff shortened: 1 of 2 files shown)


tomkerkhove:
We could still list it as a feature but mark it as `(Coming Soon)` rather than `(opt-in)`?

jeffhollan:
@tomkerkhove Id feel better doing that if we had an assigned owner. Since we don’t it’s harder to commit to

PR #47, opened 2019-11-18
removing func install

jeffhollan:
Just learned while testing out some KEDA stuff that the func tools deploy an older version of KEDA.  Have an issue with the core tools team to fix, but in meantime preparing this in case we don't want to bubble up while it'sg etting sorted out.


tomkerkhove:
Too bad but we can re-add it later on indeed!

PR #50, opened 2019-11-22
add huawei-cloudeye docs

iyacontrol:


content/scalers/huawei-cloudeye.md
```diff
@@ -0,0 +1,120 @@
...
+      metricName: mb_l7_qps
+      targetMetricValue: "100"
+      minMetricValue: "1"
+```
+
+### Authentication Parameters
+
+You can use `TriggerAuthentication` CRD to configure the authenticate by providing a set of IAM credentials.
+
+**Credential based authentication:**
```
tomkerkhove: That's interesting - Do you need all of these to authenticate or are there different combinations at play here?

content/scalers/huawei-cloudeye.md
```diff
@@ -0,0 +1,120 @@
++++
+fragment = "content"
+weight = 100
+title = "Huawei Cloudeye"
+background = "light"
++++
+
+Scale applications based on a Huawei Cloudeye.
+
+**Availability:** v1.0+ | **Maintainer:** Community
```
tomkerkhove: v1.0 has already shipped so this should be 1.1

```suggestion
**Availability:** v1.1+ | **Maintainer:** Community
```

content/scalers/huawei-cloudeye.md
```diff
@@ -0,0 +1,120 @@
...
+  - type: huawei-cloudeye
+    metadata:
+      namespace: SYS.ELB
+      dimensionName: lbaas_instance_id
+      dimensionValue: 5e052238-0346-xxb0-86ea-92d9f33e29d2
+      metricName: mb_l7_qps
+      targetMetricValue: "100"
+      minMetricValue: "1"
+```
+
```
tomkerkhove: Would it be possible to provide some brief description on what field means what? They are kind of straight forward but doesn't hurt.

Not mandatory but nice to have.

iyacontrol: thanks. done it.

content/scalers/huawei-cloudeye.md
iyacontrol: done it.

content/scalers/huawei-cloudeye.md
iyacontrol: Thanks @tomkerkhove . At the moment, it is indeed like this for huaweicloud.

tomkerkhove: That's OK, just to confirm - I'd need to specify all parameters, it's not that we need to provide a, b and c of d, e and f.

Feel free to resolve if it's all or nothing.

content/scalers/huawei-cloudeye.md
tomkerkhove: Nitpick - Can you do it in the text rather than YAMLs similar to other scalers please?

https://keda.sh/scalers/azure-service-bus/

iyacontrol: hi @tomkerkhove , i add parameter list.

tomkerkhove: Thanks man!

tomkerkhove: Awesome, thanks!

PR #58, opened 2019-12-12
added docs for azure blob

IsharaPannila:
Relates to https://github.com/kedacore/keda/issues/154

### DIFF
--- a/content/scalers/azure-storage-blob.md
+++ b/content/scalers/azure-storage-blob.md
@@ -13,13 +13,13 @@ Scale applications based on Azure Blob Storage.
 
 ### Trigger Specification
 
-This specification describes the `azure-blob` trigger for Azure Storage Blob.
+This specification describes the `azure-blob` trigger for Azure Blob Storage.
 
 ```yaml
 triggers:
   - type: azure-blob
     metadata:
-      containerName: functions-blob # Required: Name of Azure Blob Storage container
+      blobContainerName: functions-blob # Required: Name of Azure Blob Storage container
       blobCount: '5' # Optional. Amount of blobs to scale out on. Default: 5 blobs 
       connection: STORAGE_CONNECTIONSTRING_ENV_NAME
       blobPrefix:  # Optional. Prefix for the Blob. Use this to specifiy sub path for the blobs if required. Default : ""
@@ -61,15 +61,15 @@ spec:
   - type: azure-blob
     metadata:
       # Required
-      containerName: functionsblob
+      blobContainerName: functionsblob
       # Required: connection OR authenticationRef that defines connection
       connection: STORAGE_CONNECTIONSTRING_ENV_NAME # Default: AzureWebJobsStorage. Reference to a connection string in deployment
       # or authenticationRef as defined below
       #
       # Optional
       blobCount: "5" # default 5
-      blobPrefix:  # Default : ""
-      blobDelimiter: # Default: "/"
+      blobPrefix: blobsubpath # Default : ""
+      blobDelimiter: "/" # Default: "/"
     authenticationRef:
         name: azure-blob-auth # authenticationRef would need either podIdentity or define a connection parameter
 ```
--- a/content/scalers/azure-storage-blob.md
+++ b/content/scalers/azure-storage-blob.md
@@ -21,7 +21,7 @@ triggers:
     metadata:
       blobContainerName: functions-blob # Required: Name of Azure Blob Storage container
       blobCount: '5' # Optional. Amount of blobs to scale out on. Default: 5 blobs 
-      connection: STORAGE_CONNECTIONSTRING_ENV_NAME
+      connection: STORAGE_CONNECTIONSTRING_ENV_NAME # Optional if TriggerAuthentication defined with pod identity or connection string authentication.
       blobPrefix:  # Optional. Prefix for the Blob. Use this to specifiy sub path for the blobs if required. Default : ""
       blobDelimiter: # Optional. Delimiter for identifying the blob Prefix. Default: "/"
 ```
@@ -62,10 +62,6 @@ spec:
     metadata:
       # Required
       blobContainerName: functionsblob
-      # Required: connection OR authenticationRef that defines connection
-      connection: STORAGE_CONNECTIONSTRING_ENV_NAME # Default: AzureWebJobsStorage. Reference to a connection string in deployment
-      # or authenticationRef as defined below
-      #
       # Optional
       blobCount: "5" # default 5
       blobPrefix: blobsubpath # Default : ""
--- a/content/scalers/azure-storage-blob.md
+++ b/content/scalers/azure-storage-blob.md
@@ -18,7 +18,7 @@ triggers:
   - type: azure-blob
     metadata:
       containerName: functions-blob # Required: Container name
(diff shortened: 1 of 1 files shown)


content/scalers/azure-storage-blob.md
```diff
@@ -0,0 +1,73 @@
++++
+fragment = "content"
+weight = 100
+title = "Azure Storage Blob"
+background = "light"
++++
+
+Scale applications based on Azure Storage Blobs.
+
```
tomkerkhove: We also provide additional information such as when it's available and whom maintains it:
```suggestion

**Availability:** v1.1+ | **Maintainer:** Community

```

Can we list you as owner of this scaler? Or do you want to support it as Microsoft @jeffhollan?

content/scalers/azure-storage-blob.md
```diff
@@ -0,0 +1,73 @@
...
+spec:
+  scaleTargetRef:
+    deploymentName: azureblob-function
+  triggers:
+  - type: azure-blob
+    metadata:
+      # Required
+      containerName: functionsblob
+      # Required: connection OR authenticationRef that defines connection
+      connection: STORAGE_CONNECTIONSTRING_ENV_NAME # Default: AzureWebJobsStorage. Reference to a connection string in deployment
```
tomkerkhove: You're specifying both a connection here, as well as trigger authentication which doesn't really make sense - Would you mind adapting so it only uses the trigger authentication object?

content/scalers/azure-storage-blob.md
```diff
@@ -0,0 +1,73 @@
...
+
+### Trigger Specification
+
+This specification describes the `azure-blob` trigger for Azure Storage Blob.
+
+```yaml
+triggers:
+  - type: azure-blob
+    metadata:
+      containerName: functions-blob # Required: Container name
```
tomkerkhove: I presume this is aimed at the storage container, not runtime container?

Might be better to clarify this:
```suggestion
      containerName: functions-blob # Required: Name of Azure Blob Storage container
```

Or even rename it to `blobContainerName`?

content/scalers/azure-storage-blob.md
```diff
@@ -0,0 +1,73 @@
...
+      blobCount: '5' # Optional. Blob count target for HPA. Default: 5 blobs
```
tomkerkhove: What does this mean? Amount of blobs to scale on or?

Maybe this is more clear.
```suggestion
      blobCount: '5' # Optional. Amount of blobs to scale out on. Default: 5 blobs 
```

content/scalers/azure-storage-blob.md
```diff
@@ -0,0 +1,73 @@
...
+  - type: azure-blob
+    metadata:
+      containerName: functions-blob # Required: Container name
+      blobCount: '5' # Optional. Blob count target for HPA. Default: 5 blobs 
+      connection: STORAGE_CONNECTIONSTRING_ENV_NAME
+      blobPrefix:  # Optional. Prefix for the Blob. Use this to specifiy sub path for the blobs if required. Default : ""
+      blobDelimiter: # Optional. Delimiter for identifying the blob Prefix. Default: "/"
+```
+
+The `connection` value is the name of the environment variable your deployment uses to get the connection string. This is usually resolved from a `Secret V1` or a `ConfigMap V1` collections. `env` and `envFrom` are both supported.
```
tomkerkhove: I would leave this out as we now also support `TriggerAuthentication`

content/scalers/azure-storage-blob.md
```diff
@@ -0,0 +1,73 @@
++++
+fragment = "content"
+weight = 100
+title = "Azure Storage Blob"
```
tomkerkhove: I think they are typically called `Azure Blob Storage` or `Azure Storage - Blobs`
```suggestion
title = "Azure Blob Storage"
```

content/scalers/azure-storage-blob.md
```diff
@@ -0,0 +1,73 @@
...
+background = "light"
++++
+
+Scale applications based on Azure Storage Blobs.
```
tomkerkhove: I think they are typically called `Azure Blob Storage` or `Azure Storage - Blobs`
```suggestion
Scale applications based on Azure Blob Storage.
```

tomkerkhove:
Thanks for the PR @IsharaPannila! I've added a few comments to improve it from a reading/explanatory perspective.

Would you mind having a look?

content/scalers/azure-storage-blob.md
IsharaPannila: Sure. I'll update the code to use only trigger authentication and update this doc as well.

content/scalers/azure-storage-blob.md
IsharaPannila: yeah. better rename it to "blobContainerName" for the clarity. I'll update the code and then update this doc.

content/scalers/azure-storage-blob.md
IsharaPannila: Agreed. I'll first do the changes to work only with TriggerAuthentication and later remove this connection section.

content/scalers/azure-storage-blob.md
IsharaPannila: I'm fine with any decision (list me as owner or Microsoft). Anyway we are going to use this in couple of heavy blob processing projects soon and I'm always available to give the feedback and help to update/maintain this.

tomkerkhove: Well in that case I'd be happy to keep you as a maintainer, but don't want to pressure you!

If you want, feel free to use `**Maintainer:** Community (@IsharaPannila)`

content/scalers/azure-storage-blob.md
tomkerkhove: No worries, you don't need to change the code to only support trigger auth - My remark is more that this example uses a mixed approach where I think it's better to use either the environment variable approach on ScaledObject **OR** TriggerAuthentication approach

content/scalers/azure-storage-blob.md
tomkerkhove: Ah now I get it - No, please keep this around as it should support both TriggerAuthentication & this approach.

Sorry for the confusion!

IsharaPannila: Sure. No worries!

content/scalers/azure-storage-blob.md
IsharaPannila: I've updated the code and rename the param to `blobContainerName`. Also updated the docs. Please review again.

tomkerkhove: Thanks!

content/scalers/azure-storage-blob.md
```diff
@@ -0,0 +1,75 @@
...
+spec:
+  scaleTargetRef:
+    deploymentName: azureblob-function
+  triggers:
+  - type: azure-blob
+    metadata:
+      # Required
+      blobContainerName: functionsblob
+      # Required: connection OR authenticationRef that defines connection
+      connection: STORAGE_CONNECTIONSTRING_ENV_NAME # Default: AzureWebJobsStorage. Reference to a connection string in deployment
```
tomkerkhove: I would remove this or at least comment it out since it's confusing now.

IsharaPannila: done. Please review.

PR #62, opened 2020-01-09
Event Hub Scalar documentation updates and bug fixes

rasavant-ms:
- connection string for Event Hub needs Entity path with event hub name
- Added mention of Java application compatibility
- Added optional blobContainer field in trigger spec for running applications


tomkerkhove:
I didn't know we enforce the entity path, you learn something new every day!

PR #67, opened 2020-01-16
update Blob storage trigger docs to describe the scaling factor as blob count in given container

IsharaPannila:


tomkerkhove:
Thanks for the PR! Looks like we already have a page for this, would you mind updating this one instead?

IsharaPannila:
@tomkerkhove , my bad , just updated the existing page.

IsharaPannila:
closing this as Update azure-storage-blob.md #68 is the correct document update.

PR #68, opened 2020-01-16
Update azure-storage-blob.md

IsharaPannila:


ahmelsayed:
I'm still hesitant to accept this change. This is not the general contract of blobTrigger in azure functions. blobTrigger behavior is to scan the container logs from storage, store a marker where processing stopped, and interpret new activity to blobs on the container as a "trigger event". That does complicate the implementation of the scaler as it would tie it to azure functions interpretation of how blobTrigger should work, as well as the shared marker resource to let both the runtime and scaler know where the processing was left off.

jeffhollan:
Added some comments here to move description to PR https://github.com/kedacore/keda/pull/514

jeffhollan:
@IsharaPannila not sure if you ended up making any changes before merged - are these docs still the right ones with limitations or did you change the `isActive` behavior?

jeffhollan:
Oh actually I see you edited it after fact. Thanks!

Issue #65, opened 2020-01-09
Update companies contributing in "partnership" section


tomkerkhove:
Update companies contributing in `Partnership` section to not only list Microsoft & Red Hat.

PR #69, opened 2020-01-17
Provide list of companies contributing to KEDA

tomkerkhove:
Provide list of companies contributing to KEDA.

I've asked around for customers on Slack as well, once we have a few we can do the same.

Closes https://github.com/kedacore/keda-docs/issues/65


tomkerkhove:
Thanks for checking @jeffhollan !

PR #76, opened 2020-01-30
Added flag for Helm 2 installation

robece:
Added --set customResourceDefinition.create=true for Helm 2 chart installation


tomkerkhove:
Waiting with merging until  https://github.com/kedacore/charts/pull/17 is merged.

tomkerkhove:
I presume we can abandon this one since we'll stick with current approach, right?

tomkerkhove:
OK if we just close this one since nothing has changed?

PR #78, opened 2020-01-31
Updated Kafka Scalar documentation to include new authmode for Event Hubs

rasavant-ms:
Updated Kafka Scalar documentation to include new authmode for Event Hubs

### DIFF
--- a/content/scalers/apache-kafka-topic.md
+++ b/content/scalers/apache-kafka-topic.md
@@ -5,7 +5,7 @@ title = "Apache Kafka Topic"
 background = "light"
 +++
 
-Scale applications based on Apache Kafka Topic.
+Scale applications based on Apache Kafka Topic. This can also be used to scale applications based on Azure Event Hubs w/Kafka enabled.
 
 **Availability:** v1.0+ | **Maintainer:** Microsoft
 
@@ -33,9 +33,11 @@ This specification describes the `kafka` trigger for Apache Kafka Topic.
 
  You can use `TriggerAuthentication` CRD to configure the authenticate by providing authMode, username, password. If your kafka cluster does not have sasl authentication turned on, you will not need to pay attention to it.
 
+ The sasl_ssl_plain authMode can be used when using the Kafka head on Azure Event Hubs to act as trigger.
+
 **Credential based authentication:**
 
-- `authMode` Kafka sasl auth mode. Optional. The default value is none. For now, it must be one of none, sasl_plaintext, sasl_ssl, sasl_scram_sha256, sasl_scram_sha512.
+- `authMode` Kafka sasl auth mode. Optional. The default value is none. For now, it must be one of none, sasl_plaintext, sasl_ssl, sasl_ssl_plain, sasl_scram_sha256, sasl_scram_sha512.
 - `username` Optional. If authmode is not none, this is required.
 - `password` Optional.If authmode is not none, this is required.
 - `ca` Certificate authority file for TLS client authentication. Optional. If authmode is sasl_ssl, this is required.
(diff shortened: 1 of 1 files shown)


content/scalers/apache-kafka-topic.md
```diff
@@ -5,7 +5,7 @@ title = "Apache Kafka Topic"
 background = "light"
 +++
 
-Scale applications based on Apache Kafka Topic.
+Scale applications based on Apache Kafka Topic. This can also be used to scale applications based on Azure Event Hubs w/Kafka enabled.
```
ppatierno: To be honest I don't think that this kind of "specific" use case should be mentioned. Who is using AEH with Kafka head knows that can use this trigger because ... it's about Kafka topic.

content/scalers/apache-kafka-topic.md
```diff
@@ -33,9 +33,11 @@ This specification describes the `kafka` trigger for Apache Kafka Topic.
 
  You can use `TriggerAuthentication` CRD to configure the authenticate by providing authMode, username, password. If your kafka cluster does not have sasl authentication turned on, you will not need to pay attention to it.
 
+ The sasl_ssl_plain authMode can be used when using the Kafka head on Azure Event Hubs to act as trigger.
```
ppatierno: Same as above, who uses AEH knows that it supports SASL_SSL_PLAIN so the trigger has to be configured in that way.

content/scalers/apache-kafka-topic.md
```diff
@@ -132,4 +134,59 @@ spec:
       lagThreshold: "50"
     authenticationRef:
       name: keda-trigger-auth-kafka-credential
+```
+
+Your kafka head on Azure Event Hubs - turn on sasl ssl plain auth
```
ppatierno: It should be a more generic example about SASL_SSL_PLAIN not specific to AEH.
This new added mechanism works with any Kafka "installation" not just AEH. From this doc, it could be misled imho.

content/scalers/apache-kafka-topic.md
```diff
@@ -132,4 +134,59 @@ spec:
...
+
+For more details on how to get the Event Hub Namespace and Event Hub ConnectionString, refer to documentations [here](https://docs.microsoft.com/en-us/azure/event-hubs/event-hubs-quickstart-kafka-enabled-event-hubs).
```
ppatierno: I don't think that a "specific" example like this should be part of the generic Kafka topic trigger documentation. Maybe we need more a section for examples.

content/scalers/apache-kafka-topic.md
tomkerkhove: That's a fair point, what about this instead @ppatierno?
```suggestion
Scale applications based on Apache Kafka Topic or other services that support Kafka protocol.
```
Or:
```suggestion
Scale applications based on Apache Kafka Topic or other services that support Kafka protocol such as Azure Event Hubs.
```

content/scalers/apache-kafka-topic.md
tomkerkhove: I think this is a fair point in the documentation, what is your concern with this? We provide several means to authenticate and here we just point out that this works with Event Hubs, don't see an issue with that.

content/scalers/apache-kafka-topic.md
tomkerkhove: Event Hubs speaks Kafka so it's a fair point to note in the scaler docs here, but I'd suggest to introduce a "Using Kafka with other implementations" section and move it under there, ok for you?

content/scalers/apache-kafka-topic.md
tomkerkhove: Agreed, I'd delete this line here or new section as mentioned below.

content/scalers/apache-kafka-topic.md
rasavant-ms: Thanks for the feedback!
I definitely like the 2nd recommendation made by @tomkerkhove

content/scalers/apache-kafka-topic.md
rasavant-ms: @ppatierno, actually anyone who has played with producers and consumers to AEH w/Kafka endpoint would know that this is the way to authenticate with it. I feel the mention here does more good than harm :)

content/scalers/apache-kafka-topic.md
rasavant-ms: point taken! Thanks for the feedback :)

content/scalers/apache-kafka-topic.md
rasavant-ms: I guess in this case I will go with just providing a sample in the Samples documentation instead. Would that be ok?

content/scalers/apache-kafka-topic.md
ppatierno: I see it the other way around. :-) The point I am making here is that the user who uses EH and is interested in using it via Kafka's head should read the Microsoft official doc to check what kind of authentication it supports. At that point, if he wants to use KEDA then he should already know to set `sasl_ssl_plain` as auth mode so it's not worth mentioning EH here.

content/scalers/apache-kafka-topic.md
ppatierno: Yes, I agree with having a dedicated sample for it.

content/scalers/apache-kafka-topic.md
ppatierno: I am more for the first one, because if at some point another service will come supporting the Kafka protocol, then we should mention it here as well. For this reason, I would be more generic.
Then, of course, having a dedicated example of using Azure EH (as I mentioned in another comment).

content/scalers/apache-kafka-topic.md
rasavant-ms: I've modified the doc to read `The sasl_ssl_plain authMode can also be used when using the Kafka head on Azure Event Hubs to act as trigger.`

ppatierno: Which again should be something that people using Azure EH with Kafka should already know :-)
As the same point I tried to make before: imagine that the project Apache Pulsar will add a Kafka protocol head, at this point we should update the doc saying " ... when using the Kafka head on Azure Event Hubs or Apache Pulsar to act as trigger". It should happen again and again for other new services with same support. I would be more generic not mentioning specific cases but leaving them as dedicated examples.

ppatierno:
@rasavant-ms as side note on this. I think that using Azure Event Hub with Kafka could be something to add as "quickstart" alongside the other based on RabbitMQ, Kafka, ...

content/scalers/apache-kafka-topic.md
rasavant-ms: Have made the requested changes

content/scalers/apache-kafka-topic.md
```diff
@@ -132,4 +132,4 @@ spec:
       lagThreshold: "50"
     authenticationRef:
       name: keda-trigger-auth-kafka-credential
-```
\ No newline at end of file
+```
```
ppatierno: not clear what's the difference here may be git is going crazy :-)

tomkerkhove: Think it's a diff thing for sure!

PR #96, opened 2020-02-19
Remove example website of Syna theme

tomkerkhove:
Remove example website of Syna theme which has security vulnerabilities.


stp-ip:
What were/are the security vulnerabilities, if I may ask as maintainer of Syna? :)

Issue #136, opened 2020-04-08
In the Community section there are listed Users instead of Partners


zroubalik:
In https://keda.sh/community/#partners section, there should be supporting companies of our community, but there are listed Users.

It is correct on the main page https://keda.sh/, see the difference between Users/Community part over there.

PR #137, opened 2020-04-08
Fix listing users & companies in community

tomkerkhove:

Fix listing users & companies in community.

Closes https://github.com/kedacore/keda-docs/issues/136


tomkerkhove:
Can remove the users on community as well, not sure if we should add them there as well, but can't hurt I gues

PR #146, opened 2020-04-21
Fixing Apache Kafka topic naming

ppatierno:
This PR is mostly a cosmetic change about how using the Apache Kafka naming.
The doc talks about Apache Kafka Topic in the scalers menu as if it was the "project" name, which is not true. It's like saying that instead of having Azure Service Bus on the menu, we had Azure Service Bus Queue/Topic.
This PR highlight the scaler as an Apache Kafka based scaler also renaming the corresponding markdown file.


tomkerkhove:
Thanks!

PR #152, opened 2020-04-23
Extend proxy-network troubleshooting for GKE

### LINKED ISSUE #148: Extend How can I use KEDA in a proxy network? troubleshooting docs
The solution described in How can I use KEDA in a proxy network? troubleshooting docs cannot be used in managed Kubernetes clusters like GKE or EKS: https://keda.sh/docs/troubleshooting/

It would be helpful to extend this section with solution for common cloud providers:

- GKE: Add default prom-adapter port 6443 to allowed ports in master node firewall rules. I tested this and it's working.
  - https://github.com/kubernetes-sigs/metrics-server/issues/45#issuecomment-541110986
  - https://cloud.google.com/kubernetes-engine/docs/how-to/private-clusters#add_firewall_rules
- EKS: Update the Kubernetes nodes security groups to allow ingress / incoming https connections from the EKS masters security group and allow 443 egress from the control plane to the nodes. I haven't tested this.
  - https://github.com/kubernetes-sigs/metrics-server/issues/45#issuecomment-421378413
- AKS and other providers: Could be described by someone who has experience with other providers.

hmoravec:
Closes #148
I left out suggested solution for AWS as I haven't tested it.

### DIFF
--- a/content/troubleshooting/proxy-network.md
+++ b/content/troubleshooting/proxy-network.md
@@ -29,7 +29,7 @@ kubectl get apiservice v1beta1.external.metrics.k8s.io -o yaml
 
 If the status is `False`, then there seems to be an issue and proxy network might be the primary reason for it.
 
-### Follow these steps:
+### Solution for self-managed Kubernetes cluster:
 
 Find the cluster IP for the `keda-metrics-apiserver` and `keda-operator-metrics`:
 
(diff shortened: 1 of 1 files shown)


tomkerkhove:
Woops, seems like I've missed the notification on this one, sorry @hmoravec!

PR #156, opened 2020-04-29
Clarify `cooldownPeriod` for Deployments.

simpsonw:
I was a little confused by the fact that KEDA will scale your `Deployment` to `minReplicatCount` immediately after it's created.  I thought that it would wait for the `cooldownPeriod` to elapse before doing this.  I realize the docs specify that `cooldownPeriod` is the period to wait *after* the last trigger and that when you first create a `Deployment`, nothing has triggered, but I think it would be helpful to tell say that explicitly.


simpsonw:
I messed up the DCO stuff and made a new PR https://github.com/kedacore/keda-docs/pull/157

PR #157, opened 2020-04-29
Clarify `cooldownPeriod` for Deployments.

simpsonw:
Signed-off-by: Will Simpson <wsimpson@email.arizona.edu>

I was a little confused by the fact that KEDA will scale your Deployment to minReplicatCount immediately after it's created. I thought that it would wait for the cooldownPeriod to elapse before doing this. I realize the docs specify that cooldownPeriod is the period to wait after the last trigger and that when you first create a Deployment, nothing has triggered, but I think it would be helpful to tell say that explicitly.


tomkerkhove:
Ah great, my small tweak messed up the DCO; let me see if I can fix that.

Also, need more ☕

tomkerkhove:
Thanks @simpsonw !

PR #180, opened 2020-05-18
Adding Cron scaler docs for KEDA

ckuduvalli:
Signed-off-by: Chaitanya Kuduvalli Ramachandra <chaitanya.r@ETC02Y77ZVJG5H.local>

### DIFF
--- a/content/docs/1.5/scalers/cron.md
+++ b/content/docs/1.5/scalers/cron.md
@@ -1,0 +1,50 @@
++++
+title = "Cron"
+layout = "scaler"
+availability = "v1.5+"
+maintainer = "Community"
+description = "Scale applications based on a cron schedule."
+go_file = "cron_scaler"
++++
+
+### Trigger Specification
+
+This specification describes the `cron` trigger that scales based on a Cron Schedule.
+
+```yaml
+triggers:
+- type: cron
+  metadata:
+    # Required
+    timezone: Asia/Kolkata  # The acceptable values would be a value from the IANA Time Zone Database.
+    start: 0 0/30 * * *
+    end: 0 15/30 * * *
+    desiredReplicas: "10"
+```
+
+**Parameter list:**
+
+- `timezone` is one of the acceptable values from the IANA Time Zone Database. Required.
+- `start` is the cron expression indicating the start of the cron schedule. Required.
+- `end` is the cron expression indicating the end of the cron schedule. Required.
+- `desiredReplicas` is the number of replicas to which the resource has to be scaled between the start and end of the cron schedule. Required.
+
+### Example
+
+```yaml
+apiVersion: keda.k8s.io/v1alpha1
+kind: ScaledObject
+metadata:
+  name: cron-scaledobject
+  namespace: default
+spec:
+  scaleTargetRef:
+    deploymentName: my-deployment
+  triggers:
+  - type: cron
+    metadata:
+      timezone: Asia/Kolkata
+      start: 0 0/30 * * *
+      end: 0 15/30 * * *
+      desiredReplicas: "10"
+```
--- a/content/docs/1.5/scalers/cron.md
+++ b/content/docs/1.5/scalers/cron.md
@@ -17,8 +17,8 @@ triggers:
   metadata:
     # Required
     timezone: Asia/Kolkata  # The acceptable values would be a value from the IANA Time Zone Database.
-    start: 0 0/30 * * *
-    end: 0 15/30 * * *
+    start: 30 * * * *       # Every hour on the 30th minute
+    end: 45 * * * *         # Every hour on the 45th minute
     desiredReplicas: "10"
 ```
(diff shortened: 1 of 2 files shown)


content/docs/1.5/scalers/cron.md
```diff
@@ -0,0 +1,50 @@
...
+    # Required
+    timezone: Asia/Kolkata  # The acceptable values would be a value from the IANA Time Zone Database.
+    start: 0 0/30 * * *
+    end: 0 15/30 * * *
+    desiredReplicas: "10"
+```
+
+**Parameter list:**
+
+- `timezone` is one of the acceptable values from the IANA Time Zone Database. Required.
```
zroubalik, line 27: Would be nice to add link to the list of timezones, so KEDA user could find the suitable easily.

content/docs/1.5/scalers/cron.md
```diff
@@ -0,0 +1,50 @@
...
+
+This specification describes the `cron` trigger that scales based on a Cron Schedule.
+
+```yaml
+triggers:
+- type: cron
+  metadata:
+    # Required
+    timezone: Asia/Kolkata  # The acceptable values would be a value from the IANA Time Zone Database.
+    start: 0 0/30 * * *
```
zroubalik, line 20: What kind of Cron sytax are we using here? Are we using standard of do we support some non-standard syntax? From the `/` it seems like that we support some non-standard one. https://en.wikipedia.org/wiki/Cron Would be nice to describe. 

What is this specific example specify? (hours = 0/30) ?

tomkerkhove: I'm wondering if cron is actually the best approach for defining this meaning, why don't we use time stamps for defining start & end of a window (in UTC ofcourse)?

A bit late to come up with this, but was just wondering about the reasoning behind it.

ckuduvalli: This is the library I used. https://github.com/ringtail/go-cron
0 0/30 * * * indicates "Every 30 minutes starting at :00 minute after the hour"

ckuduvalli: The Slashes are non standard characters as specified in the wiki itself. 
Slash (/)
    In vixie-cron, slashes can be combined with ranges to specify step values. For example, */5 in the minutes field indicates every 5 minutes. So, '/' is used mostly to describe frequencies.

zroubalik: @tomkerkhove I think that cron is a good approach to specify this kind of (repetitive) tasks.

zroubalik: @ckuduvalli I am not a cron expert, but as  `/` is used to describe frequencies, I'd think that it is saying something like "every 30 hours". As the format is: `minutes, hours, day of month, mont, day of week` https://en.wikipedia.org/wiki/Cron#Overview

zroubalik: so that's why I am asking since this specific example is a little bit confusing.

zroubalik: Another thing, (I haven't seen the code yet), but you are mentioning, that you want to use, why don't you use the original corn: https://github.com/robfig/cron which is far more popular/stable on github and you are using a fork which seems to be not so popular and it's last commit is from Aug 2019?

ckuduvalli: ringtail cron was an improvement over robfig. But I have changed it back to robfig since its a popular one. Will send out a code review by tomorrow. Testing out some stuff.

content/docs/1.5/scalers/cron.md
ckuduvalli: // LoadLocation returns the Location with the given name.
//
// If the name is "" or "UTC", LoadLocation returns UTC.
// If the name is "Local", LoadLocation returns Local.
//
// Otherwise, the name is taken to be a location name corresponding to a file
// in the IANA Time Zone database, such as "America/New_York".
//
// The time zone database needed by LoadLocation may not be
// present on all systems, especially non-Unix systems.
// LoadLocation looks in the directory or uncompressed zip file
// named by the ZONEINFO environment variable, if any, then looks in
// known installation locations on Unix systems,
// and finally looks in $GOROOT/lib/time/zoneinfo.zip.

ckuduvalli: This is the comment from the code zoneinfo.go
Does it make sense to capture this kind of information in the documentation?

content/docs/1.5/scalers/cron.md
ckuduvalli: @zroubalik please check the format now

content/docs/1.5/scalers/cron.md
ckuduvalli: @zroubalik have added the link to the list of timezones

PR #183, opened 2020-05-21
added Manage persmission requirement

programaka:
Updated Service Bus scaler docs: SAS Policy needs to be of type 'Manage'.

https://github.com/kedacore/keda/issues/806

Signed-off-by: Daria Kiritseva <v-dariak@microsoft.com>


tomkerkhove:
That's correct, should be in all our versions indeed!

programaka:
added the change to v1.4 and v1.5

programaka:
@zroubalik signed, thanks!

PR #189, opened 2020-06-11
Patch 2

audunsol:
Just found a broken link on the web page https://keda.sh/docs/1.4/concepts/scaling-jobs/ that I thought I could try to fix here.


audunsol:
Not very experienced with this. Forgot the signing part. Will try again. Closing this one.

PR #195, opened 2020-06-16
WIP - Ensure scalers are only listed for a given version

tomkerkhove:

### DIFF
--- a/layouts/partials/scalers.html
+++ b/layouts/partials/scalers.html
@@ -3,11 +3,14 @@
 <div class="buttons has-extra-top-margin">
   {{ range $scalers }}
   {{ $isHere := eq .RelPermalink $here }}
-  {{ $version := index (split .Dir "/") 1 }}
+  {{ $version1 := index (split .Dir "\\") 0 }}
+  {{ $version2 := index (split .Dir "\\") 1 }}
+  {{ $version3 := index (split .Dir "\\") 3 }}
+  {{ $version4 := index (split .Dir "\\") 4 }}
   <!-- Use this if you are on Windows:
        {{ $version := index (split .Dir "\\") 1 }} -->
   <a class="button is-primary{{ if not $isHere }} is-outlined{{ end }}" href="{{ .RelPermalink }}">
-    {{ .Title }} (Version: {{ $version }})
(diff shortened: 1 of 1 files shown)


tomkerkhove:
Let's see if this works since I cannot fully run it locally.

tomkerkhove:
Had a working version, cleaned things up and messed it up again 😂

The fun of frontend...

lucperkins:
I think the issue is that you’re using \\ as the file path separator instead of / further down

layouts/partials/scalers.html
```diff
@@ -1,10 +1,15 @@
 {{ $here    := $.Page.RelPermalink }}
 {{ $scalers := where site.RegularPages ".CurrentSection.Title" "Scalers" }}
+{{ $thisVersion    := index (split .File.Path "/") 1 }}
 <div class="buttons has-extra-top-margin">
   {{ range $scalers }}
   {{ $isHere := eq .RelPermalink $here }}
+  {{ $version := index (split .File.Path "\\") 1 }}
```
lucperkins, line 7: Use the `/` separator here as well, instead of `\\`

lucperkins:
@tomkerkhove What's the issue when you run it locally?

zroubalik:
@tomkerkhove have you been able to fix this one?

tomkerkhove:
Nope, unfortunately not...

layouts/partials/scalers.html
tomkerkhove: This works fine on my machine:
`{{ $version := index (split .Dir "\\") 1 }}`

But must be because I'm on Windows.

tomkerkhove:
The current approach works on my box (Windows) by using `\\` but if I move it to `/`. Any pointers for this @lucperkins ?

Local:
image

lucperkins:
I don't think there's a way around this problem. The only way to infer the version of a given document is to read it from the filepath. Since the production site is built on Linux, you'll need to use `/`.

tomkerkhove:
That's what I do indeed, but the build is failing then and I'm not sure why/how to fix it.

tomkerkhove:
Closing in favor of #200

PR #205, opened 2020-07-08
Correcting incorrect property reference for the mySQL scaler example

tbickford:
The mySQL AuthenticationRef should use `name` (https://github.com/kedacore/keda/blob/1b38a7fe1fc7c05535c96955fd043a99e9fb22f1/deploy/crds/keda.k8s.io_scaledobjects_crd.yaml#L4506) instead of `connectionString`


tomkerkhove:
Well spotted, thanks!

PR #208, opened 2020-07-10
Fix users icon layout

ahmelsayed:


tomkerkhove:
I will test it on my pc this weekend!

PR #217, opened 2020-08-05
Prom metrics

tbickford:
Adding section for documentation around the Prometheus Client Exporter


tomkerkhove:
Small suggestion to move it above FAQ:
image

content/docs/2.0/operate.md
```diff
@@ -0,0 +1,14 @@
...
+## Prometheus Exporter Metrics
+
+The KEDA Metrics Adapter exposes Prometheus metrics which can be scraped on port `9022` (this can be changed by setting the `metrics-port` argument for the Metrics Adapter) at `/metrics`.  The metrics collected in the Metrics Adapter are only active when the HPA is active (> 0 replicas).
+
+The following metrics are being gathered:
+
+- `keda_metrics_adapter_scaler_error_totals` - The total number of errors encountered for all scalers
+- `keda_metrics_adapter_scaled_object_error_totals`- The number of errors that have occurred for each scaled object
+- `keda_metrics_adapter_scaler_errors` - The number of erros that have occurred for each scaler
+- `keda_metrics_adapter_scaler_metrics_value`- The current value for each scaler's metric that would be used by the HPA in computing the target average.
```
tomkerkhove, line 14: This is more the metric for a given scaled object then, no?

content/docs/2.0/operate.md
```diff
@@ -0,0 +1,14 @@
...
++++
+
+## Prometheus Exporter Metrics
+
+The KEDA Metrics Adapter exposes Prometheus metrics which can be scraped on port `9022` (this can be changed by setting the `metrics-port` argument for the Metrics Adapter) at `/metrics`.  The metrics collected in the Metrics Adapter are only active when the HPA is active (> 0 replicas).
+
+The following metrics are being gathered:
+
+- `keda_metrics_adapter_scaler_error_totals` - The total number of errors encountered for all scalers
+- `keda_metrics_adapter_scaled_object_error_totals`- The number of errors that have occurred for each scaled object
```
tomkerkhove, line 12: Given ScaledJobs are now seperate, should we report the metric separately?

content/docs/2.0/operate.md
tbickford: If a scaled object that multiple scalers there would be separate metrics by scaler so I thought calling it scaler would be more appropriate. I can rename it. If so, it would also make sense to rename `keda_metrics_adapter_scaler_errors` to `keda_metrics_adapter_scaled_object_metric_errors` or something

content/docs/2.0/operate.md
tbickford: I believe all the scaling metrics for the scaled jobs is completely done inside the Operator and not the Metrics Adapter. Maybe the next step is add a prometheus exporter for the operator as well?

zroubalik: +1 operator-sdk (which is using kubebuilder under the hood) has built in Prometheus Server and provides some basic metrics. It should be possible to extend the exposed metrics https://book.kubebuilder.io/reference/metrics.html#publishing-additional-metrics
But I'd keep it for a separate PR

content/docs/2.0/operate.md
tomkerkhove: My remark was actually more for `keda_metrics_adapter_scaled_object_error_totals` or is that what you are talking about?

content/docs/2.0/operate.md
```diff
@@ -0,0 +1,14 @@
...
+- `keda_metrics_adapter_scaler_errors` - The number of erros that have occurred for each scaler
```
tomkerkhove, line 13:
```suggestion
- `keda_metrics_adapter_scaler_errors` - The number of errors that have occurred for each scaler
```

config.toml
```diff
@@ -162,17 +162,23 @@ name = "Migration Guide"
 parent = "docs"
 weight = 4
 
+[[menu.docs]]
+url = "/docs/operate/"
+name = "Operate"
```
tomkerkhove, line 167: Nit - Can we move this in front of the Migration Guide with weight 4 please?

content/docs/2.0/operate.md
tomkerkhove: Ok, resolving!

PR #218, opened 2020-08-05
Migration guide: Deployments scaling part

zroubalik:
Signed-off-by: Zbynek Roubalik <zroubali@redhat.com>

### DIFF
--- a/content/docs/2.0/migration.md
+++ b/content/docs/2.0/migration.md
@@ -2,20 +2,20 @@
 title = "Migration Guide"
 +++
 
-# Migrating from KEDA v1 to v2
+## Migrating from KEDA v1 to v2
 
 Please note that you **can not** run both KEDA v1 and v2 on the same Kubernetes cluster. You need to [uninstall](../../1.5/deploy) KEDA v1 first, in order to [install](../deploy) and use KEDA v2.
 
 KEDA v2 is using a new namespace for it's Custom Resources: `keda.sh` instead of `keda.k8s.io` and introduces a new Custom Resource for scaling of Jobs. See full details on KEDA Custom Resources [here](../concepts/#custom-resources-crd).
 
-## Scaling of Deployments
+### Scaling of Deployments
 In order to scale `Deployments` with KEDA v2, you need to do only a few modifications to existing v1 `ScaledObjects` definitions, so they comply with v2:
 - change the value of `apiVersion` property from `keda.k8s.io/v1alpha1` to `keda.sh/v1alpha1`
 - rename property `spec.scaleTargetRef.deploymentName` to `spec.scaleTargetRef.name`
 
-Please see the examples below or refer to the full [v2 ScaledObject documentation](../concepts/scaling-deployments)
+Please see the examples below or refer to the full [v2 ScaledObject Specification](../concepts/scaling-deployments/#scaledobject-spec)
 
-### Example of v1 ScaledObject
+ **Example of v1 ScaledObject**
 ```yaml
 apiVersion: keda.k8s.io/v1alpha1
 kind: ScaledObject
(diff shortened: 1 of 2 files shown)


content/docs/2.0/migration.md
```diff
@@ -0,0 +1,86 @@
...
+- rename property `spec.scaleTargetRef.deploymentName` to `spec.scaleTargetRef.name`
+
+Please see the examples below or refer to the full [v2 ScaledObject Specification](../concepts/scaling-deployments/#scaledobject-spec)
+
+ **Example of v1 ScaledObject**
+```yaml
+apiVersion: keda.k8s.io/v1alpha1
+kind: ScaledObject
+metadata:
+  name: {scaled-object-name}
```
tomkerkhove, line 23: Should we add the label here that was mandatory in the past and we are now removing (although not even using it anymore)?

content/docs/2.0/migration.md
```diff
@@ -0,0 +1,86 @@
++++
+title = "Migration Guide"
++++
+
+## Migrating from KEDA v1 to v2
+
+Please note that you **can not** run both KEDA v1 and v2 on the same Kubernetes cluster. You need to [uninstall](../../1.5/deploy) KEDA v1 first, in order to [install](../deploy) and use KEDA v2.
+
+KEDA v2 is using a new namespace for it's Custom Resources: `keda.sh` instead of `keda.k8s.io` and introduces a new Custom Resource for scaling of Jobs. See full details on KEDA Custom Resources [here](../concepts/#custom-resources-crd).
```
tomkerkhove, line 9: Would this be more clear? At first I thought you meant a Kubernetes namespace.
```suggestion
KEDA v2 is using a new API version namespace for it's Custom Resources Definitions (CRD): `keda.sh` instead of `keda.k8s.io` and introduces a new Custom Resource for scaling of Jobs. See full details on KEDA Custom Resources [here](../concepts/#custom-resources-crd).
```

content/docs/2.0/migration.md
```diff
@@ -0,0 +1,86 @@
...
+
+### Scaling of Deployments
+In order to scale `Deployments` with KEDA v2, you need to do only a few modifications to existing v1 `ScaledObjects` definitions, so they comply with v2:
+- change the value of `apiVersion` property from `keda.k8s.io/v1alpha1` to `keda.sh/v1alpha1`
```
tomkerkhove, line 13: Nit
```suggestion
- Change the value of `apiVersion` property from `keda.k8s.io/v1alpha1` to `keda.sh/v1alpha1`
```

content/docs/2.0/migration.md
```diff
@@ -0,0 +1,86 @@
...
+- rename property `spec.scaleTargetRef.deploymentName` to `spec.scaleTargetRef.name`
```
tomkerkhove, line 14: Nit
```suggestion
- Rename property `spec.scaleTargetRef.deploymentName` to `spec.scaleTargetRef.name`
```

content/docs/2.0/migration.md
```diff
@@ -0,0 +1,86 @@
...
+  pollingInterval: 30            
+  cooldownPeriod:  300           
+  minReplicaCount: 0             
+  maxReplicaCount: 100           
+  triggers:
+  # {list of triggers to activate the deployment}
+```
+
+
+### Scaling of Jobs
```
tomkerkhove, line 55: If it's a seperate PR, maybe best to leave out for now?

zroubalik: I wanted to add the Jobs part stub, so we have the basic structure and don't forget about it :) It should be added anyway. But if it is a big problem, I can remove it

content/docs/2.0/migration.md
zroubalik: It is not mandatory for a few last v1 releases, so I am not sure whether we should mention it here?

content/docs/2.0/migration.md
tomkerkhove: It's ok!

content/docs/2.0/migration.md
tomkerkhove: I'd say it can't hurt; we don't know what version folks are coming from?

Maybe add it but mention that this has not been required for a while and finally being removed?

Maybe overthinking here but we're not all always on latest version.

zroubalik: Check the latest commit. Something like this?

tomkerkhove: Perfect, thanks ❤️

PR #239, opened 2020-08-25
Add metrics API scaler docs

turbaszek:
Related to https://github.com/kedacore/keda/issues/929

### DIFF
--- a/content/docs/2.0/scalers/metrics-api.md
+++ b/content/docs/2.0/scalers/metrics-api.md
@@ -3,82 +3,23 @@ title = "Metrics API"
 layout = "scaler"
 availability = "v2.0+"
 maintainer = "Community"
-description = "Scale applications based on a metric value from an API"
+description = "Scale applications based on a metric from an API"
 go_file = "metrics_api_scaler"
 +++
 
-This specification describes the `metrics-api` trigger that scales based on a metric value from an API.
-
-### Trigger prerequisites
-
-To use an external API as a source of metric you need to have a following endpoint:
-
-```yaml
-/api/v1/metric/{metric}:
-  get:
-    summary: Get Metric
-    description: Get value for a given metric
-    operationId: GetMetric
-    parameters: 
-     - name: metric
-       in: path
-       required: true
-       schema:
-          type: string
-       description: Name of the metric for which information is requested
-    responses:
-      '200':
-        description: Metric inforation was provided
-        content:
-          application/json:
-            schema:
-              $ref: '#/components/schemas/ReportedMetric'
-      '404':
-        description: Metric is not supported
-      '401':
-        $ref: '#/components/responses/UnauthorizedError'
-      '500':
-        description: Unable to determine metric information
-    tags:
-      - Metrics
-    security:
-      - basic_auth: []
-      - api_key: []
-```
-which returns:
-```yaml
-ReportedMetric:
-  type: object
-  required:
-    - name
-    - value
-  properties:
-    name:
-      type: string
-    value:
-      type: number
-```
-
-
 ### Trigger Specification
+
+This specification describes the `metrics-api` trigger that scales based on a metric value from an API.
 
 The metrics API requires following configuration values:
 - `apiURL` string repesenting the base URL of the API (i.e `http://app:1317/api/v1/`).
 - `metricName` the metric name to be used to retive value .
-- `targetValue` the target value for HPA. As long as the current metric doesn't match `TargetValue`, 
-  HPA will increase the number of the pods until it reaches the maximum number of pods allowed to scale to.
+- `targetValue` the target value is the target value to scale on. When the metric provided by the 
+    API is equal or higher to this value, KEDA will start scaling out.
 
-
-An example of scaled object using metric API:
+An example of scaled object spec using metric API:
 
 ```yaml
-apiVersion: keda.k8s.io/v1alpha1
-kind: ScaledObject
-metadata:
-  name: metrics-api-scaledobject
-  namespace: default
-  labels:
-    deploymentName: dummy
 spec:
   maxReplicaCount: 4
   scaleTargetRef:
@@ -93,3 +34,134 @@ spec:
 
 **Note**:
 This scaler scales deployment to 0 if and only if the value of the metric is lower or equal to zero.
+
+### Trigger prerequisites
+
+To use an external API it has to comply with:
+
+```yaml
+openapi: 3.0.1
+info:
+  title: KEDA - External Metric Source Petstore
+  description: >-
+    This is the specification to comply with for external KEDA metric sources.
+  termsOfService: 'http://swagger.io/terms/'
+  contact:
+    email: apiteam@swagger.io
+  license:
+    name: Apache 2.0
+    url: 'http://www.apache.org/licenses/LICENSE-2.0.html'
+  version: 1.0.0
+externalDocs:
+  description: Find out more about Swagger
+  url: 'http://swagger.io'
+servers:
+  - url: 'https://samples.keda.sh/'
+tags:
+  - name: Health
+    description: Operations about runtime health
+  - name: Metrics
+    description: Operations about providing metrics to KEDA
+paths:
+  /api/v1/metric/{metric}:
+    get:
+      summary: Get Metric
+      description: Get value for a given metric
+      operationId: GetMetric
+      parameters: 
+       - name: metric
+         in: path
+         required: true
+         schema:
+            type: string
+         description: Name of the metric for which information is requested
+      responses:
+        '200':
+          description: Metric inforation was provided
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ReportedMetric'
+        '404':
+          description: Metric is not supported
+        '401':
+          $ref: '#/components/responses/UnauthorizedError'
+        '500':
+          description: Unable to determine metric information
+      tags:
+        - Metrics
+      security:
+        - basic_auth: []
+        - api_key: []
+  /api/v1/metrics:
+    get:
+      summary: Get All Metrics
+      description: Provides a list of all supported metrics
+      operationId: GetMetrics
+      responses:
+        '200':
+          description: Metric inforation was provided
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/MetricInfo'
+        '401':
+          $ref: '#/components/responses/UnauthorizedError'
+        '500':
+          description: Unable to determine metric information
+      tags:
+        - Metrics
+      security:
+        - basic_auth: []
+        - api_key: []
+  /api/v1/health:
+    get:
+      summary: Get Health
+      description: Provides information about the health of the runtime
+      operationId: GetHealth
+      responses:
+        '200':
+          description: External metric source is healthy
+        '401':
+          $ref: '#/components/responses/UnauthorizedError'
+        '500':
+          description: External metric source is not healthy
+      tags:
+        - Health
+      security:
+        - basic_auth: []
+        - api_key: []
+components:
+  schemas:
+    MetricInfo:
+      type: array
+      items:
+        type: object
+        properties:
+          name:
+            type: string
+    ReportedMetric:
+      type: object
+      required:
+        - name
+        - value
+      properties:
+        name:
+          type: string
+        value:
+          type: number
+  responses:
+    UnauthorizedError:
+      description: Authentication information is missing or invalid
+  securitySchemes:
+    basic_auth:
+      type: http
+      scheme: basic
+    api_key:
+      type: apiKey
+      name: X-API-Key
+      in: header
+security:
+  - basic_auth: []
+  - api_key: []
+```
--- a/content/docs/2.0/scalers/metrics-api.md
+++ b/content/docs/2.0/scalers/metrics-api.md
@@ -12,7 +12,7 @@ go_file = "metrics_api_scaler"
 This specification describes the `metrics-api` trigger that scales based on a metric value from an API.
 
 The metrics API requires following configuration values:
-- `apiURL` string repesenting the base URL of the API (i.e `http://app:1317/api/v1/`).
+- `apiURL` string representing the base URL of the API (i.e `http://app:1317/api/v1/`).
 - `metricName` the metric name to be used to retive value .
 - `targetValue` the target value is the target value to scale on. When the metric provided by the 
     API is equal or higher to this value, KEDA will start scaling out.
@@ -23,12 +23,12 @@ An example of scaled object spec using metric API:
 spec:
   maxReplicaCount: 4
   scaleTargetRef:
-    deploymentName: dummy
+    name: dummy
   triggers:
     - type: metric-api
       metadata:
         targetValue: "1"
-        apiURL: "http://api:3232/api/v1"
+        url: "http://api:3232/api/v1"
         metricName: "workersNumber"
 ```
 
--- a/content/docs/2.0/scalers/metrics-api.md
+++ b/content/docs/2.0/scalers/metrics-api.md
@@ -10,14 +10,8 @@ go_file = "metrics_api_scaler"
 ### Trigger Specification
 
 This specification describes the `metrics-api` trigger that scales based on a metric value provided by an API. 
-This scaler allows users to utilize any existing APIs as a metric provider.  
 
-**Parameter list:**
-- `url`: Full URL of the API operation to call to get the metric value (eg. `http://app:1317/api/v1/stats`).
-- `valueLocation`: [GJSON path notation](https://github.com/tidwall/gjson#path-syntax) to refer to the field in 
-    the payload containing the metric value
-- `targetValue`: Target value to scale on. When the metric provided by the API is equal or higher to this value, 
-    KEDA will start scaling out. When the metric is 0 or less, KEDA will scale down to 0.
+This scaler allows users to utilize **any existing APIs** as a metric provider.  
 
 Here is an example of trigger configuration using metric-api scaler:
 
@@ -30,13 +24,42 @@ triggers:
     valueLocation: "components.worker.tasks"
 ```
 
+**Parameter list:**
+- `url`: Full URL of the API operation to call to get the metric value (eg. `http://app:1317/api/v1/stats`).
+- `valueLocation`: [GJSON path notation](https://github.com/tidwall/gjson#path-syntax) to refer to the field in 
+    the payload containing the metric value
+- `targetValue`: Target value to scale on. When the metric provided by the API is equal or higher to this value, 
+    KEDA will start scaling out. When the metric is 0 or less, KEDA will scale down to 0.
+
 ### Authentication Parameters
 
 Not supported yet.
 
 ### Example
 
-When checking current metric this scaler sends GET request to provided `url` and then uses `valueLocation`
+Here is a full example of scaled object definition using Metric API trigger:
+
+```yaml
+apiVersion: keda.sh/v1alpha1
+kind: ScaledObject
+metadata:
+  name: http-scaledobject
+  namespace: keda
+  labels:
+    deploymentName: dummy
+spec:
+  maxReplicaCount: 12
+  scaleTargetRef:
+    name: dummy
+  triggers:
+    - type: metrics-api
+      metadata:
+        targetValue: "7"
+        url: "http://api:3232/components/stats"
+        valueLocation: 'components.worker.tasks'
+```
+
+When checking current metric Metrics API scaler sends GET request to provided `url` and then uses `valueLocation`
 to access the value in response's payload. 
 
 The above example expects that the API endpoint will return response similar to this one:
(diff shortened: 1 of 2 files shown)


content/docs/2.0/scalers/metrics-api.md
```diff
@@ -0,0 +1,95 @@
...
+    deploymentName: dummy
+spec:
+  maxReplicaCount: 4
+  scaleTargetRef:
+    deploymentName: dummy
+  triggers:
+    - type: metric-api
+      metadata:
+        targetValue: "1"
+        apiURL: "http://api:3232/api/v1"
```
tomkerkhove, line 90: I would just go for `url` here since it's the metrics API scaler
```suggestion
        url: "http://api:3232/api/v1"
```

content/docs/2.0/scalers/metrics-api.md
```diff
@@ -0,0 +1,95 @@
...
+
+An example of scaled object using metric API:
+
+```yaml
+apiVersion: keda.k8s.io/v1alpha1
+kind: ScaledObject
+metadata:
+  name: metrics-api-scaledobject
+  namespace: default
+  labels:
```
tomkerkhove, line 80: No need to include this

content/docs/2.0/scalers/metrics-api.md
```diff
@@ -0,0 +1,95 @@
...
+go_file = "metrics_api_scaler"
++++
+
+This specification describes the `metrics-api` trigger that scales based on a metric value from an API.
+
+### Trigger prerequisites
+
+To use an external API as a source of metric you need to have a following endpoint:
+
+```yaml
```
tomkerkhove, line 16: Let's add the whole OpenAPI spec here so that people can use it to validate their endpoint

content/docs/2.0/scalers/metrics-api.md
```diff
@@ -0,0 +1,95 @@
++++
```
tomkerkhove, line 1: Would be good to follow same structure from other scalers such as https://github.com/kedacore/keda-docs/blob/master/content/docs/2.0/scalers/redis-streams.md.

Let's talk about the Trigger Specification first, add a section on the API they need and what needs to comply with and end with an example of how to configure it.

content/docs/2.0/scalers/metrics-api.md
```diff
@@ -0,0 +1,95 @@
...
+title = "Metrics API"
+layout = "scaler"
+availability = "v2.0+"
+maintainer = "Community"
+description = "Scale applications based on a metric value from an API"
```
tomkerkhove, line 6: This feels more natural to me but that's personal pref, and a big nit😂
```suggestion
description = "Scale applications based on a metric from an API"
```

But not a must though!

content/docs/2.0/scalers/metrics-api.md
```diff
@@ -0,0 +1,95 @@
...
+      type: number
+```
+
+
+### Trigger Specification
+
+The metrics API requires following configuration values:
+- `apiURL` string repesenting the base URL of the API (i.e `http://app:1317/api/v1/`).
+- `metricName` the metric name to be used to retive value .
+- `targetValue` the target value for HPA. As long as the current metric doesn't match `TargetValue`,
```
tomkerkhove, line 68: We cannot assume everybody knows what an HPA is (that's one of the selling points of KEDA, no need to know how Kubernetes scales), so how about:

```suggestion
- `targetValue` is the target value to scale on. When the metric provided by the API is equal or higher to this value, KEDA will start scaling out.
```

content/docs/2.0/scalers/metrics-api.md
tomkerkhove: Another question - Should they configure `http://api:3232` or `http://api:3232/api/v1`? Given we've defined the operations and the endpoint might evolve over time I'd argue that it should just be `http://api:3232` and we'll handle the rest, no?

tomkerkhove (changes requested):
Looks good overall but would like to follow the structure of other scalers and added some remarks.

content/docs/2.0/scalers/metrics-api.md
tomkerkhove: What do you think @zroubalik @ahmelsayed ?

tomkerkhove:
And don't hesitate to drop a note about this new scaler in https://github.com/kedacore/keda-docs/blob/master/content/blog/keda-2.0-beta.md.draft#L26 given it allows for higher extensibility as well.

Happy to do for you if it's too much of a hassle.

content/docs/2.0/scalers/metrics-api.md
turbaszek: @tomkerkhove that means that we have to incorporate the API version somewhere else, right?

tomkerkhove: Given they have to comply with our OpenAPI spec it's already covered.

tomkerkhove: https://github.com/kedacore/keda-docs/pull/239/files#diff-2079ef1d5bc7dbe2c2470f43d8749c87R17

tomkerkhove: If that endpoint doesn't exist, then the API is not supported 🤷‍♂️.

If we introduce a `v2

content/docs/2.0/scalers/metrics-api.md
```diff
@@ -0,0 +1,95 @@
...
+- `apiURL` string repesenting the base URL of the API (i.e `http://app:1317/api/v1/`).
+- `metricName` the metric name to be used to retive value .
+- `targetValue` the target value for HPA. As long as the current metric doesn't match `TargetValue`, 
+  HPA will increase the number of the pods until it reaches the maximum number of pods allowed to scale to.
+
+
+An example of scaled object using metric API:
+
+```yaml
+apiVersion: keda.k8s.io/v1alpha1
```
zroubalik:
```suggestion
apiVersion: keda.sh/v1alpha1
```

content/docs/2.0/scalers/metrics-api.md
```diff
@@ -0,0 +1,95 @@
...
+kind: ScaledObject
+metadata:
+  name: metrics-api-scaledobject
+  namespace: default
+  labels:
+    deploymentName: dummy
+spec:
+  maxReplicaCount: 4
+  scaleTargetRef:
+    deploymentName: dummy
```
zroubalik, line 26:
```suggestion
    name: dummy
```

content/docs/2.0/scalers/metrics-api.md
zroubalik: I like the current approach, I am not sure if there's a benefit in splitting those fields (the config gets more complex)

turbaszek: But if I update my API version in a backward-compatible manner then the endpoint will exist. I think it's better to have full base link as this allows more extensibility. For example, some may have `http://app:1234/experimental/api/v1/` due to ongoing migration process or whatever

tomkerkhove: I'm not saying it has to be separated but just that we care about the base URL.

All metric APIs need to comply with the OpenAPI spec; if we add endpoints there, KEDA will determine what the best one is to call. If other endpoints exist next to it, we'll simply ignore them.

If we configure the full prefix in it, what's the point of having an OpenAPI spec anyway :)

content/docs/2.0/scalers/metrics-api.md
```diff
@@ -0,0 +1,167 @@
...
+The metrics API requires following configuration values:
+- `apiURL` string representing the base URL of the API (i.e `http://app:1317/api/v1/`).
+- `metricName` the metric name to be used to retive value .
+- `targetValue` the target value is the target value to scale on. When the metric provided by the 
+    API is equal or higher to this value, KEDA will start scaling out.
+
+An example of scaled object spec using metric API:
+
+```yaml
+spec:
```
zroubalik, line 23: I think you can remove the whole `spec:` section and start from `triggers`

content/docs/2.0/scalers/metrics-api.md
turbaszek: >I'm not saying it has to be separated but just that we care about the base URL.

What you mean by "base URL"?

> If we configure the full prefix in it, what's the point of having an OpenAPI spec anyway :)

To be honest, that's the advantage. If you use OpenAPI spec that's good. If you don't... well you can still use the scaler. And in my opinion, that's an advantage. Moreover, we only use `api/v1//metric/{name}` in scaler

tomkerkhove: With base URL I was referring to `http://app:1234`, the rest is handled by the scaler which knows `api/v1/metric/{name}` must be there so it just adds it automatically.

tomkerkhove: For me every user has to use that endpoint, if you want to use ´http://app:1234/experimental/api/v1/` then that's a violation imo

turbaszek: Btw. here's @zroubalik comment https://github.com/kedacore/keda-docs/pull/239#discussion_r477143636

If we use only `api/v1/metric/{name}` then how users can change the API version they want to use? And the version can change.

> if you want to use ´http://app:1234/experimental/api/v1/` then that's a violation imo

I would call this violation an "extensibility". Let assume that base url is `http://app:1234`. Should we validate it to do not allow `http://app:1234/staging/user/test/and/more/` as input? 😉

content/docs/2.0/scalers/metrics-api.md
```diff
@@ -0,0 +1,166 @@
...
+description = "Scale applications based on a metric from an API"
+go_file = "metrics_api_scaler"
++++
+
+### Trigger Specification
+
+This specification describes the `metrics-api` trigger that scales based on a metric value from an API.
+
+The metrics API requires following configuration values:
+- `url` string representing the base URL of the API (i.e `http://app:1317/api/v1/`).
```
zroubalik, line 15:
```suggestion
- `url` string representing the base URL of the API (eg. `http://app:1317/api/v1/`).
```

lonnix:
I looked over the docs and things look good. The only part I had to re-read was the bit about scaling down to 0 but it made sense after I read it a 2nd time.

content/docs/2.0/scalers/metrics-api.md
```diff
@@ -0,0 +1,53 @@
++++
+title = "Metrics API"
+layout = "scaler"
+availability = "v2.0+"
+maintainer = "Community"
+description = "Scale applications based on a metric from an API"
```
tomkerkhove, line 6:
```suggestion
description = "Scale applications based on a metric provided by an API"
```

content/docs/2.0/scalers/metrics-api.md
```diff
@@ -0,0 +1,53 @@
...
+go_file = "metrics_api_scaler"
++++
+
+### Trigger Specification
+
+This specification describes the `metrics-api` trigger that scales based on a metric value from an API. This scaler
```
tomkerkhove, line 12:
```suggestion
This specification describes the `metrics-api` trigger that scales based on a metric value provided by an API. This scaler
```

content/docs/2.0/scalers/metrics-api.md
```diff
@@ -0,0 +1,53 @@
...
+allows users to utilize any existing APIs as a source of metric.
```
tomkerkhove, line 13:
```suggestion
allows users to utilize any existing APIs as a metric source.  
```

Alternatively:
```suggestion
allows users to utilize any existing APIs as a metric provider.  
```

content/docs/2.0/scalers/metrics-api.md
```diff
@@ -0,0 +1,53 @@
...
+triggers:
+- type: metric-api
+  metadata:
+    targetValue: "8"
+    url: "http://api:3232/api/v1/stats"
+    valueLocation: "components.worker.tasks"
+```
+
+When checking current metric this scaler sends GET request to provided `url` and then uses `valueLocation`
+to access the value in response's payload. The above example expects that the API endpoint will return response
```
tomkerkhove, line 34: To improve rendered docs:
```suggestion
to access the value in response's payload.

The above example expects that the API endpoint will return response
```

content/docs/2.0/scalers/metrics-api.md
```diff
@@ -0,0 +1,53 @@
...
+      ...
+    },
+    ...
+  },
+  ...
+}
+```
+Assuming such response, metric-api trigger will figure out that current metric value is 12.
+
+**Note**:
```
tomkerkhove, line 50: Might be good to align with our other notes:
```suggestion
> **Note**: This scaler scales deployment to 0 if and only if the value of the metric is lower or equal to zero.
```
Similar to https://github.com/kedacore/keda-docs/blob/master/content/docs/2.0/concepts/scaling-deployments.md#L193

Alternatively, use `notice` as we do for Azure Event Hubs but not a must.
https://github.com/kedacore/keda-docs/blob/master/content/docs/2.0/scalers/azure-event-hub.md

content/docs/2.0/scalers/metrics-api.md
```diff
@@ -0,0 +1,53 @@
...
+description = "Scale applications based on a metric from an API"
+go_file = "metrics_api_scaler"
++++
+
+### Trigger Specification
+
+This specification describes the `metrics-api` trigger that scales based on a metric value from an API. This scaler
+allows users to utilize any existing APIs as a source of metric.  
+
+The metrics-api trigger requires following configuration values:
```
tomkerkhove, line 15:
```suggestion
**Parameter list:**
```

content/docs/2.0/scalers/metrics-api.md
```diff
@@ -0,0 +1,53 @@
...
+- `url` string representing the full URL that will be used to get the metric (eg. `http://app:1317/api/v1/stats`).
```
tomkerkhove, line 16:
```suggestion
- `url`: Full URL of the API operation to call to get the metric value (eg. `http://app:1317/api/v1/stats`).
```

content/docs/2.0/scalers/metrics-api.md
```diff
@@ -0,0 +1,53 @@
...
+- `valueLocation` string using [GJSON path notation](https://github.com/tidwall/gjson#path-syntax) to point to
+    required value that will be used as current metric value 
+- `targetValue` the target value is the target value to scale on. When the metric provided by the
```
tomkerkhove, line 19:
```suggestion
- `targetValue`: Target value to scale on. When the metric provided by the 
```

content/docs/2.0/scalers/metrics-api.md
```diff
@@ -0,0 +1,53 @@
...
+
+This specification describes the `metrics-api` trigger that scales based on a metric value from an API. This scaler
+allows users to utilize any existing APIs as a source of metric.  
+
+The metrics-api trigger requires following configuration values:
+- `url` string representing the full URL that will be used to get the metric (eg. `http://app:1317/api/v1/stats`).
+- `valueLocation` string using [GJSON path notation](https://github.com/tidwall/gjson#path-syntax) to point to
+    required value that will be used as current metric value 
+- `targetValue` the target value is the target value to scale on. When the metric provided by the 
+    API is equal or higher to this value, KEDA will start scaling out.
```
tomkerkhove, line 20: Might be good to document this here already
```suggestion
    API is equal or higher to this value, KEDA will start scaling out. When the metric is 0 or less, KEDA will scale down to 0.
```

content/docs/2.0/scalers/metrics-api.md
```diff
@@ -0,0 +1,53 @@
...
++++
+
+### Trigger Specification
+
+This specification describes the `metrics-api` trigger that scales based on a metric value from an API. This scaler
+allows users to utilize any existing APIs as a source of metric.  
+
+The metrics-api trigger requires following configuration values:
+- `url` string representing the full URL that will be used to get the metric (eg. `http://app:1317/api/v1/stats`).
+- `valueLocation` string using [GJSON path notation](https://github.com/tidwall/gjson#path-syntax) to point to
```
tomkerkhove, line 17: How about this?
```suggestion
- `valueLocation`: [GJSON path notation](https://github.com/tidwall/gjson#path-syntax) to refer to the field in the payload containing the metric value
```

content/docs/2.0/scalers/metrics-api.md
```diff
@@ -0,0 +1,53 @@
...
+    "worker": {
+      "tasks": 12,
+      ...
+    },
+    ...
+  },
+  ...
+}
+```
+Assuming such response, metric-api trigger will figure out that current metric value is 12.
```
tomkerkhove, line 48:
```suggestion
Assuming such response, Metrics API trigger will figure out that current metric value is 12.
```
or:
```suggestion
Assuming such response, `metric-api` trigger will figure out that current metric value is 12.
```

tomkerkhove (changes requested):
I've added a bunch of suggestions to make it more readable/less tech detailed but it would be nice if you could align with the structure of all our scalers ([example](https://keda.sh/docs/2.0/scalers/rabbitmq-queue/)):
- Trigger Specification with parameter list after sample
- Authentication Parameters (none for now, but following https://keda.sh/docs/1.5/scalers/prometheus/#authentication-parameters)
- Example

How it works with the API could be a (sub)section before Authentication Parameters to talk about that or as part of the sample.

turbaszek:
Thanks @tomkerkhove for your suggestions! I applied all of them

content/docs/2.0/scalers/metrics-api.md
```diff
@@ -0,0 +1,59 @@
...
+**Parameter list:**
+- `url`: Full URL of the API operation to call to get the metric value (eg. `http://app:1317/api/v1/stats`).
+- `valueLocation`: [GJSON path notation](https://github.com/tidwall/gjson#path-syntax) to refer to the field in 
+    the payload containing the metric value
+- `targetValue`: Target value to scale on. When the metric provided by the API is equal or higher to this value, 
+    KEDA will start scaling out. When the metric is 0 or less, KEDA will scale down to 0.
+
+Here is an example of trigger configuration using metric-api scaler:
+
+```yaml
```
tomkerkhove, line 24: Nit - This is typically listed before the parameters list; it's easier to resonate.

content/docs/2.0/scalers/metrics-api.md
```diff
@@ -0,0 +1,59 @@
...
+      ...
+    },
+    ...
+  },
+  ...
+}
+```
+Assuming such response, Metrics API trigger will figure out that current metric value is 12.
+
+> NOTE: This scaler scales deployment to 0 if and only if the value of the metric is lower or equal to zero.
```
tomkerkhove, line 57: Actually I think we can remove it since it's in the parameters list

content/docs/2.0/scalers/metrics-api.md
```diff
@@ -0,0 +1,59 @@
...
+    targetValue: "8"
+    url: "http://api:3232/api/v1/stats"
+    valueLocation: "components.worker.tasks"
+```
+
+### Authentication Parameters
+
+Not supported yet.
+
+### Example
```
tomkerkhove, line 37: Here we typically add a full YAML sample on how to use it.

content/docs/2.0/scalers/metrics-api.md
```diff
@@ -0,0 +1,59 @@
...
+availability = "v2.0+"
+maintainer = "Community"
+description = "Scale applications based on a metric provided by an API"
+go_file = "metrics_api_scaler"
++++
+
+### Trigger Specification
+
+This specification describes the `metrics-api` trigger that scales based on a metric value provided by an API. 
+This scaler allows users to utilize any existing APIs as a metric provider.
```
tomkerkhove, line 13: Just a suggestion to improve rendered docs:
```suggestion

This scaler allows users to utilize any existing APIs as a metric provider.  
```

tomkerkhove (commented):
Almost there!

content/docs/2.0/scalers/metrics-api.md
turbaszek: Added, used an example that I was using for testing purposes

tomkerkhove (approved):
LGTM - Thanks @turbaszek!

content/docs/2.0/scalers/metrics-api.md
```diff
@@ -0,0 +1,80 @@
...
+maintainer = "Community"
+description = "Scale applications based on a metric provided by an API"
+go_file = "metrics_api_scaler"
++++
+
+### Trigger Specification
+
+This specification describes the `metrics-api` trigger that scales based on a metric value provided by an API. 
+
+This scaler allows users to utilize **any existing APIs** as a metric provider.
```
turbaszek, line 14: Bold to emphasize than any API mean in fact any API, so one can scale their deployment using current day length from https://ipgeolocation.io/astronomy-api.html 😃

PR #241, opened 2020-09-02
add FromEnv suffix to metadata properties

ahmelsayed:

Relates to https://github.com/kedacore/keda/issues/1044

### DIFF
--- a/content/docs/2.0/scalers/aws-cloudwatch.md
+++ b/content/docs/2.0/scalers/aws-cloudwatch.md
@@ -26,9 +26,9 @@ triggers:
     # Required: region
     awsRegion: "eu-west-1"
     # Optional: AWS Access Key ID, can use TriggerAuthentication as well
-    awsAccessKeyID: AWS_ACCESS_KEY_ID # default AWS_ACCESS_KEY_ID
+    awsAccessKeyIDFromEnv: AWS_ACCESS_KEY_ID # default AWS_ACCESS_KEY_ID
     # Optional: AWS Secret Access Key, can use TriggerAuthentication as well
-    awsSecretAccessKey: AWS_SECRET_ACCESS_KEY # default AWS_SECRET_ACCESS_KEY
+    awsSecretAccessKeyFromEnv: AWS_SECRET_ACCESS_KEY # default AWS_SECRET_ACCESS_KEY
     identityOwner: pod | operator # Optional. Default: pod
 ```
 
@@ -41,7 +41,7 @@ triggers:
 
 ### Authentication Parameters
 
-> These parameters are relevant only when `identityOwner` is set to `pod`. 
+> These parameters are relevant only when `identityOwner` is set to `pod`.
 
 You can use `TriggerAuthentication` CRD to configure the authenticate by providing either a role ARN or a set of IAM credentials.
 
@@ -70,7 +70,7 @@ metadata:
 data:
   AWS_ACCESS_KEY_ID: <user-id>
   AWS_SECRET_ACCESS_KEY: <key>
---- 
+---
 apiVersion: keda.sh/v1alpha1
 kind: TriggerAuthentication
 metadata:
--- a/content/docs/2.0/scalers/azure-storage-blob.md
+++ b/content/docs/2.0/scalers/azure-storage-blob.md
@@ -17,13 +17,13 @@ triggers:
 - type: azure-blob
   metadata:
     blobContainerName: functions-blob # Required: Name of Azure Blob Storage container
-    blobCount: '5' # Optional. Amount of blobs to scale out on. Default: 5 blobs 
-    connection: STORAGE_CONNECTIONSTRING_ENV_NAME # Optional if TriggerAuthentication defined with pod identity or connection string authentication.
+    blobCount: '5' # Optional. Amount of blobs to scale out on. Default: 5 blobs
+    connectionFromEnv: STORAGE_CONNECTIONSTRING_ENV_NAME # Optional if TriggerAuthentication defined with pod identity or connection string authentication.
     blobPrefix:  # Optional. Prefix for the Blob. Use this to specifiy sub path for the blobs if required. Default : ""
     blobDelimiter: # Optional. Delimiter for identifying the blob Prefix. Default: "/"
 ```
 
-The `connection` value is the name of the environment variable your deployment uses to get the connection string. This is usually resolved from a `Secret V1` or a `ConfigMap V1` collections. `env` and `envFrom` are both supported.
+The `connectionFromEnv` value is the name of the environment variable your deployment uses to get the connection string. This is usually resolved from a `Secret V1` or a `ConfigMap V1` collections. `env` and `envFrom` are both supported.
 
 ### Authentication Parameters
 
--- a/content/docs/2.0/scalers/azure-storage-queue.md
+++ b/content/docs/2.0/scalers/azure-storage-queue.md
@@ -17,10 +17,10 @@ triggers:
   metadata:
     queueName: functionsqueue
     queueLength: '5' # Optional. Queue length target for HPA. Default: 5 messages
-    connection: STORAGE_CONNECTIONSTRING_ENV_NAME
+    connectionFromEnv: STORAGE_CONNECTIONSTRING_ENV_NAME
 ```
 
-The `connection` value is the name of the environment variable your deployment uses to get the connection string. This is usually resolved from a `Secret V1` or a `ConfigMap V1` collections. `env` and `envFrom` are both supported.
+The `connectionFromEnv` value is the name of the environment variable your deployment uses to get the connection string. This is usually resolved from a `Secret V1` or a `ConfigMap V1` collections. `env` and `envFrom` are both supported.
 
 ### Authentication Parameters
 
@@ -55,7 +55,7 @@ spec:
       # Required
       queueName: functionsqueue
       # Required: connection OR authenticationRef that defines connection
-      connection: STORAGE_CONNECTIONSTRING_ENV_NAME # Default: AzureWebJobsStorage. Reference to a connection string in deployment
+      connectionFromEnv: STORAGE_CONNECTIONSTRING_ENV_NAME # Default: AzureWebJobsStorage. Reference to a connection string in deployment
       # or authenticationRef as defined below
       #
       # Optional
--- a/content/docs/2.0/scalers/mysql.md
+++ b/content/docs/2.0/scalers/mysql.md
@@ -33,11 +33,11 @@ You can authenticate by using connection string or password authentication.
 
 **Connection String Authentication:**
 
-- `connectionFromEnv` - Connection string for MySQL database
+- `connectionString` - Connection string for MySQL database
 
 **Password Authentication:**
 
-- `connectionFromEnv` - Password for configured user to login to MySQL database
+- `password` - Password for configured user to login to MySQL database
 variables.
 
 ### Example
--- a/content/docs/2.0/scalers/mysql.md
+++ b/content/docs/2.0/scalers/mysql.md
@@ -16,15 +16,15 @@ The trigger always requires the following information:
 - `query` - a MySQL query that should return single numeric value
 - `queryValue` - a threshold that is used as `targetAverageValue` in HPA.
 
-To provide information about how to connect to MySQL you can provide 
-- `connectionString` MySQL connection string that should point to environment variable with valid value
+To provide information about how to connect to MySQL you can provide
+- `connectionStringFromEnv` MySQL connection string that should point to environment variable with valid value
 
 Or provide more detailed information:
 
 - `username` used to access MySQL database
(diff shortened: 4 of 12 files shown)


content/docs/2.0/scalers/aws-cloudwatch.md
```diff
@@ -26,9 +26,9 @@ triggers:
     # Required: region
     awsRegion: "eu-west-1"
     # Optional: AWS Access Key ID, can use TriggerAuthentication as well
-    awsAccessKeyID: AWS_ACCESS_KEY_ID # default AWS_ACCESS_KEY_ID
+    awsAccessKeyIDFromEnv: AWS_ACCESS_KEY_ID # default AWS_ACCESS_KEY_ID
```
tomkerkhove, line 29: Can't we specify literal values of the ID anymore? Think that's still valuable.

tomkerkhove: Looks like the code still supports this so maybe just good to list both options every time?

ahmelsayed: Yes, all the properties with `FromEnv` suffix can be overridden by one that doesn't have `FromEnv`. I was thinking I can add a `> ` note on all the scalers with something like 

> Metadata properties with `*FromEnv` suffix are resolved from the environment of the pods in scaleTargetRef. It's recommended to source these settings from a Secret on the `scaleTargetRef` or through a TriggerAuthentication CRD. Dropping `FromEnv` will use the literal value defined in the ScaledObject

ahmelsayed: I'm a bit conflicted though about this not discouraging users enough from storing secrets in the ScaledObject CRD. What do you think about having a `*PlainText` suffix for ones with literal values in the CRD itself? I understand the ease of use when you're trying to get started by defining everything in one CRD, but just want a clear indication that there is a better way of doing it.

tomkerkhove: There are a few things:
1. I would simply add every property to our trigger spec & parameter list in the docs, this gives a clear overview of the configuration options
2. A notice on how it works is definitely a good idea, certainly if one will be favored over another
3. I do get your point about stimulating them to use secrets for secrets, I don't think we should have plain text options for all the options. For example, literal value for `awsAccessKeyID` is fine and we can just document it. However, `awsSecretAccessKey` should only be configurable via `awsSecretAccessKeyFromEnv` or through TriggerAuthentication (and thus documented like this)

That way; safe things can be configured as literal values, env var or TriggerAuth while secrets can only be configured through env var or Trigger Auth.

I personally think that's the sweet spot.

tomkerkhove: Adding @zroubalik as he might have some input as well

zroubalik: Agree with @tomkerkhove

zroubalik (commented):
Would be really great to put the table from the PR https://github.com/kedacore/keda/pull/1072#issue-478227944 to Migration Guide as well.

content/docs/2.0/scalers/aws-cloudwatch.md
ahmelsayed: [This](https://github.com/kedacore/keda/pull/1131) makes the change needed to resolve secrets only for properties using `FromEnv`. Updating the docs now

ahmelsayed:
@tomkerkhove I just updated the docs. Can you please take another look and let me know what you think

content/docs/2.0/scalers/azure-storage-blob.md
```diff
@@ -17,13 +17,13 @@ triggers:
...
-    blobCount: '5' # Optional. Amount of blobs to scale out on. Default: 5 blobs 
-    connection: STORAGE_CONNECTIONSTRING_ENV_NAME # Optional if TriggerAuthentication defined with pod identity or connection string authentication.
+    blobCount: '5' # Optional. Amount of blobs to scale out on. Default: 5 blobs
+    connectionFromEnv: STORAGE_CONNECTIONSTRING_ENV_NAME # Optional if TriggerAuthentication defined with pod identity or connection string authentication.
     blobPrefix:  # Optional. Prefix for the Blob. Use this to specifiy sub path for the blobs if required. Default : ""
     blobDelimiter: # Optional. Delimiter for identifying the blob Prefix. Default: "/"
 ```
 
-The `connection` value is the name of the environment variable your deployment uses to get the connection string. This is usually resolved from a `Secret V1` or a `ConfigMap V1` collections. `env` and `envFrom` are both supported.
+The `connectionFromEnv` value is the name of the environment variable your deployment uses to get the connection string. This is usually resolved from a `Secret V1` or a `ConfigMap V1` collections. `env` and `envFrom` are both supported.
```
tomkerkhove, line 26: General remark for all scalers - Does this mean our auth parameter is renamed to this as well? I don't think so, right?

For example here, do I use `connection` or `connectionFromEnv` [here](https://github.com/kedacore/keda-docs/pull/241/files#diff-7cf10601f41d343322f1a34a3d4ef023R34)?

content/docs/2.0/scalers/mysql.md
```diff
@@ -33,11 +33,11 @@ You can authenticate by using connection string or password authentication.
 
 **Connection String Authentication:**
 
-- `connection` - Connection string for MySQL database
+- `connectionFromEnv` - Connection string for MySQL database
```
tomkerkhove, line 36: This feels a bit odd to me that we are renaming the auth parameters given FromEnv conflicts with pod identity, secrets,etc

tomkerkhove: Sorry that I've missed this in the PR

content/docs/2.0/scalers/redis-lists.md
```diff
@@ -15,19 +15,19 @@ This specification describes the `redis` trigger that scales based on the length
 triggers:
 - type: redis
   metadata:
-    address: REDIS_HOST # Required host:port format
-    password: REDIS_PASSWORD
+    addressFromEnv: REDIS_HOST # Required host:port format
```
tomkerkhove, line 18:
```suggestion
    address: localhost:6379 # Required host:port format
    addressFromEnv: REDIS_HOST # Required host:port format
```

content/docs/2.0/scalers/redis-streams.md
```diff
@@ -33,29 +33,32 @@ triggers:
 
 **Parameter list:**
```
tomkerkhove, line 34: Can we update above as well please?

content/docs/2.0/scalers/redis-lists.md
```diff
@@ -53,7 +53,7 @@ You can authenticate by using a password.
 
 Here is an example of how to deploy a scaled object with the `redis` scale trigger which uses `TriggerAuthentication`.
 
-You can also provide the `password` on the `ScaledObject` directly.
+You can also provide the `passwordFromEnv` on the `ScaledObject` directly.
 
 ```yaml
```
tomkerkhove, line 58: Sample below is not updated: https://github.com/kedacore/keda-docs/pull/241/files#diff-6bfbb18b82b17a7405561fd1316e3f5bR90

content/docs/2.0/scalers/redis-streams.md
```diff
@@ -89,8 +92,8 @@ spec:
   triggers:
     - type: redis-streams
       metadata:
-        address: REDIS_HOST
-        password: REDIS_PASSWORD # name of the environment variable in the Deployment
+        addressFromEnv: REDIS_HOST
+        passwordFromEnv: REDIS_PASSWORD # name of the environment variable in the Deployment
         stream: my-stream
         consumerGroup: consumer-group-1
         pendingEntriesCount: "10"
```
tomkerkhove, line 99: Sample below is not updated: https://github.com/kedacore/keda-docs/pull/241/files#diff-b6608de35c29cd153d0fce47f4357c26R140

content/docs/2.0/scalers/rabbitmq-queue.md
```diff
@@ -26,20 +26,21 @@ triggers:
 
 **Parameter list:**
```
tomkerkhove, line 27: Can we update above please?

content/docs/2.0/scalers/mysql.md
```diff
@@ -33,11 +33,11 @@ You can authenticate by using connection string or password authentication.
...
 
 **Password Authentication:**
 
-- `connection` - Password for configured user to login to MySQL database
+- `connectionFromEnv` - Password for configured user to login to MySQL database
 variables.
```
tomkerkhove, line 42: Can we update below please?

content/docs/2.0/scalers/mysql.md
```diff
@@ -33,11 +33,11 @@ You can authenticate by using connection string or password authentication.
 
 **Connection String Authentication:**
 
-- `connection` - Connection string for MySQL database
+- `connectionFromEnv` - Connection string for MySQL database
 
 **Password Authentication:**
 
-- `connection` - Password for configured user to login to MySQL database
+- `connectionFromEnv` - Password for configured user to login to MySQL database
```
tomkerkhove, line 40: is this correct? Because above we use connectionstring (which feels more natural as well)

content/docs/2.0/scalers/azure-storage-queue.md
```diff
@@ -17,10 +17,10 @@ triggers:
...
     queueName: functionsqueue
     queueLength: '5' # Optional. Queue length target for HPA. Default: 5 messages
-    connection: STORAGE_CONNECTIONSTRING_ENV_NAME
+    connectionFromEnv: STORAGE_CONNECTIONSTRING_ENV_NAME
 ```
 
-The `connection` value is the name of the environment variable your deployment uses to get the connection string. This is usually resolved from a `Secret V1` or a `ConfigMap V1` collections. `env` and `envFrom` are both supported.
+The `connectionFromEnv` value is the name of the environment variable your deployment uses to get the connection string. This is usually resolved from a `Secret V1` or a `ConfigMap V1` collections. `env` and `envFrom` are both supported.
 
 ### Authentication Parameters
```
tomkerkhove, line 25: The connection parameter here is not in line with the sample, which one is correct?

tomkerkhove (changes requested):
Added a few notes to update examples/specs.

Added a general remark with regards to trigger authentication parameters as they are not aligned. Some use `connection` while others use `connectionFromEnv`.

I'd prefer the first approach given `FromEnv` in that context is odd but I think this is a find+replace issue. (example: https://github.com/kedacore/keda/blob/v2/pkg/scalers/postgresql_scaler.go#L78)

content/docs/2.0/scalers/azure-storage-blob.md
ahmelsayed: yes, the auth parameters are not renamed. Those are defined under `### Authentication Parameters` in each scaler doc. 

So if you are specifying the setting on the metadata, it takes `FromEnv` for values that are resolved from the scale target env. If you are specifying them through a `TriggerAuthentication`, then there is no `FromEnv` because it wouldn't make sense there as the `TriggerAuthentication` already defines where to resolve the value from, right?

The confusing thing about the linked example is that it shows either using `connectionFromEnv` or podIdentity. to use TriggerAuthentication in that example it would be 


```yaml
apiVersion: keda.sh/v1alpha1
kind: TriggerAuthentication
metadata:
  name: keda-trigger-auth-for-blob
spec:
  secretTargetRef:
  - parameter: connection
    name: my-secret-object
    key: AZURE_STORAGE_CONNECTION
```

content/docs/2.0/scalers/mysql.md
ahmelsayed: you are right, this shouldn't change.

content/docs/2.0/scalers/azure-storage-queue.md
ahmelsayed: mentioned this in https://github.com/kedacore/keda-docs/pull/241/files#r486098364

content/docs/2.0/scalers/mysql.md
ahmelsayed: the example below is fine as is because it's using a `TriggerAuthentication` which hasn't changed.

tomkerkhove:
> So if you are specifying the setting on the metadata, it takes FromEnv for values that are resolved from the scale target env. If you are specifying them through a TriggerAuthentication, then there is no FromEnv because it wouldn't make sense there as the TriggerAuthentication already defines where to resolve the value from, right?

This definitely makes sense and was what I was hoping for, thanks for confirming! Just commented it because some were updated but looks like you are already on it. Thanks!

ahmelsayed:
went a bit on auto pilot with vim.

zroubalik (approved):
LGTM!

tomkerkhove (approved):

PR #245, opened 2020-09-07
1056 Updated property name to match the update

kashifsoofi:
Documentation update for https://github.com/kedacore/keda/pull/1088


kashifsoofi:
I guess this can be declined as there is no change to api any more

zroubalik:
> I guess this can be declined as there is no change to api any more

Agree, thanks!

PR #253, opened 2020-09-10
Consolidate apiHost & host for RabbitMQ scaler

turbaszek:
Signed-off-by: Tomek Urbaszek <tomasz.urbaszek@polidea.com>

### DIFF
--- a/content/docs/1.5/scalers/rabbitmq-queue.md
+++ b/content/docs/1.5/scalers/rabbitmq-queue.md
@@ -16,30 +16,30 @@ triggers:
 - type: rabbitmq
   metadata:
     host: RabbitMqHost # Optional. If not specified, it must be done by using TriggerAuthentication.
+    protocol: amqp # Optional. Specifies protocol to use. Either amqp or http 
     queueLength: '20' # Optional. Queue length target for HPA. Default: 20 messages
     queueName: testqueue
-    includeUnacked: 'true' # Optional, use unacked + ready messages count
-    apiHost: RabbitApiHost # Optional. Represents the HTTP management API endpoint. If not specified, it must be done by using TriggerAuthentication.
   authenticationRef:
     name: keda-trigger-auth-rabbitmq-conn
 ```
 
 **Parameter list:**
 
-- `host`: Value is the name of the environment variable your deployment uses to get the connection string. This is usually resolved from a `Secret V1` or a `ConfigMap V1` collections. `env` and `envFrom` are both supported.  The resolved host should follow a format like `amqp://guest:password@localhost:5672/vhost`
+- `host`: Value is the name of the environment variable your deployment uses to get the connection string. 
+    This is usually resolved from a `Secret V1` or a `ConfigMap V1` collections. `env` and `envFrom` are both 
+    supported. The resolved host should either follow a format like `amqp://guest:password@localhost:5672/vhost` or 
+    `https://guest:password@localhost:443/vhostname`.
 - `queueName`: Name of the queue to read message from. Required.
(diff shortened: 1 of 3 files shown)


content/docs/1.5/scalers/rabbitmq-queue.md
```diff
@@ -16,30 +16,30 @@ triggers:
 - type: rabbitmq
   metadata:
     host: RabbitMqHost # Optional. If not specified, it must be done by using TriggerAuthentication.
```
tomkerkhove, line 18: Hold your horses here, we are not changing v1.5 😄

tomkerkhove: Can you change this in v2 please?

content/docs/1.5/scalers/rabbitmq-queue.md
```diff
@@ -16,30 +16,30 @@ triggers:
...
+    protocol: amqp # Optional. Specifies protocol to use. Either amqp or http
```
tomkerkhove, line 19: Is it really optional? I think we should enforce it

content/docs/1.5/scalers/rabbitmq-queue.md
```diff
@@ -16,30 +16,30 @@ triggers:
...
-    includeUnacked: 'true' # Optional, use unacked + ready messages count
-    apiHost: RabbitApiHost # Optional. Represents the HTTP management API endpoint. If not specified, it must be done by using TriggerAuthentication.
   authenticationRef:
     name: keda-trigger-auth-rabbitmq-conn
 ```
 
 **Parameter list:**
 
-- `host`: Value is the name of the environment variable your deployment uses to get the connection string. This is usually resolved from a `Secret V1` or a `ConfigMap V1` collections. `env` and `envFrom` are both supported.  The resolved host should follow a format like `amqp://guest:password@localhost:5672/vhost`
+- `host`: Value is the name of the environment variable your deployment uses to get the connection string.
```
tomkerkhove, line 28: Better check v2 docs given you can now use `host` or `hostFromEnv`

content/docs/1.5/scalers/rabbitmq-queue.md
```diff
@@ -16,30 +16,30 @@ triggers:
...
+- `protocol`: Protocol to be used for communication. Either `http` or `amqp`. It should correspond with the `host` value. 
+ 
 
 ### Authentication Parameters
 
 TriggerAuthentication CRD is used to connect and authenticate to RabbitMQ:
 
-- `host`: AMQP URI connection string, like `amqp://guest:password@localhost:5672/vhost`.
-- `apiHost`: HTTP API endpoint, like `https://guest:password@localhost:443/vhostname`. 
+- `host`: AMQP URI connection string if `protocol=amqp`, like `amqp://guest:password@localhost:5672/vhost` or
```
tomkerkhove, line 41:
```suggestion
- `host`: Uri to talk to the host which depends on the protocol that is specified.
  - For AMQP, the URI should look similar to `amqp://guest:password@localhost:5672/vhost`
  - For HTTP, the URI should look similar to `https://guest:password@localhost:443/vhostname`
```

content/docs/1.5/scalers/rabbitmq-queue.md
tomkerkhove: I see AMQP is default, it's ok then but should be documented

tomkerkhove (changes requested):
A few suggestions and we'd need to move it to the 2.0 docs

tomkerkhove (approved):

PR #270, opened 2020-09-28
adds detailed docs for offset reset policy and invalid offset management

grassiale:
Adds a detailed documentation for offset reset policy and management of invalid offset when new the consumer has no offset committed.
As suggested in: https://github.com/kedacore/keda/discussions/1191#discussioncomment-85066

Signed-off-by: grassiale <alessandro.grassi01@gmail.com>


content/docs/2.0/scalers/apache-kafka.md
```diff
@@ -48,6 +48,14 @@ TLS:
 - `cert`: Certificate for client authentication. Optional.If `tls` is enabled, this is required.
 - `key`: Key for client authentication. Optional. If `tls` is enabled, this is required.
 
+### New Consumers and Offset Reset Policy
+ 
+ When a new kafka consumer is created, it must determine its consumer group initial poisition, i.e. the offset it will start to read from. The position is decided in kafka consumers via a parameter `auto.offset.reset` and the possible values to set are `latest` (kafka default), and `earliest`. This parameter in Keda should be set accordingly. In this initial status, no offset has been committed to kafka for the consumer group and any request for offset metadata will return an `INVALID_OFFSET`; so Keda has to manage the consumer pod's autoscaling in relation to the offset reset policy that has been specified in the parameters:
```
tomkerkhove, line 53:
```suggestion
 When a new Kafka consumer is created, it must determine its consumer group initial position, i.e. the offset it will start to read from. The position is decided in Kafka consumers via a parameter `auto.offset.reset` and the possible values to set are `latest` (Kafka default), and `earliest`. This parameter in Keda should be set accordingly. In this initial status, no offset has been committed to Kafka for the consumer group and any request for offset metadata will return an `INVALID_OFFSET`; so Keda has to manage the consumer pod's autoscaling in relation to the offset reset policy that has been specified in the parameters:
```

content/docs/2.0/scalers/apache-kafka.md
```diff
@@ -48,6 +48,14 @@ TLS:
...
+ - If the policy is set to `earliest` (a new consumer wants to replay everything in the topic from its beginning) and no offset is committed, the scaler will return a lag value equal to the last offset in the topic (in the case of a new topic, 0), so it will scale the eployment to 0 replicas. If a new messages is produced to the topic, Keda will return the new value of the offset (1), and will scale the deployments to consume the message.
```
tomkerkhove, line 54:
```suggestion
 - If the policy is set to `earliest` (a new consumer wants to replay everything in the topic from its beginning) and no offset is committed, the scaler will return a lag value equal to the last offset in the topic (in the case of a new topic, 0), so it will scale the deployment to 0 replicas. If a new message is produced to the topic, Keda will return the new value of the offset (1), and will scale the deployments to consume the message.
```

content/docs/2.0/scalers/apache-kafka.md
```diff
@@ -48,6 +48,14 @@ TLS:
...
+ - If the policy is set to `latest` (so the new consumer will only consume new messages) and no offset is committed, the scaler will return a negative lag value, and will also tell the hpa to remain `active`, hence the deployment should have the minimum number of replicas running. This is to allow the consumer to read any new message in the topic, and commit its offset.
```
tomkerkhove, line 55:
```suggestion
 - If the policy is set to `latest` (so the new consumer will only consume new messages) and no offset is committed, the scaler will return a negative lag value, and will also tell the HPA to remain `active`, hence the deployment should have the minimum number of replicas running. This is to allow the consumer to read any new message on the topic, and commit its offset.
```

tomkerkhove (approved):
LGTM, some minor suggestions

grassiale:
shoot, I broke the DCO by committing via github web ui, will fix this.

zroubalik:
@grassiale thanks a lot!

content/docs/2.0/scalers/apache-kafka.md
```diff
@@ -48,6 +48,14 @@ TLS:
...
 
+### New Consumers and Offset Reset Policy
+ 
+ When a new Kafka consumer is created, it must determine its consumer group initial position, i.e. the offset it will start to read from. The position is decided in Kafka consumers via a parameter `auto.offset.reset` and the possible values to set are `latest` (Kafka default), and `earliest`. This parameter in Keda should be set accordingly. In this initial status, no offset has been committed to Kafka for the consumer group and any request for offset metadata will return an `INVALID_OFFSET`; so Keda has to manage the consumer pod's autoscaling in relation to the offset reset policy that has been specified in the parameters:
+ - If the policy is set to `earliest` (a new consumer wants to replay everything in the topic from its beginning) and no offset is committed, the scaler will return a lag value equal to the last offset in the topic (in the case of a new topic, 0), so it will scale the deployment to 0 replicas. If a new message is produced to the topic, Keda will return the new value of the offset (1), and will scale the deployments to consume the message.
```
davidboweninrupt, line 54: Hi @grassiale,

I may have misunderstood, but I the way I read this suggests that when using `earliest` it will always scale to 0 replicas. That would imply there will be no pods run to move along from `earliest`.

I was wondering if perhaps the "in the case" part should be a new sentence which would then set the context to explain why there are no replicas and also the last sentence would be referring to a previously empty queue.

Something like:

>  - If the policy is set to `earliest` (a new consumer wants to replay everything in the topic from its beginning) and no offset is committed, the scaler will return a lag value equal to the last offset in the topic. In the case of a new topic the last offset will be 0, so it will scale the deployment to 0 replicas. If a new message is produced to the topic, Keda will return the new value of the offset (1), and will scale the deployments to consume the message.

zroubalik: @davidboweninrupt this is a nice improvement, could you please open a PR with this?

davidboweninrupt: Thank you for the kind words. I've created [Make behavior clearer for empty queue #770](https://github.com/kedacore/keda-docs/pull/770).

PR #298, opened 2020-11-02
Fix the issue of installing KEDA 2.0 beta chart

zzxwill:
Running `helm install keda kedacore/keda –namespace keda –version 2.0.0-beta1.2` will hit the issue as below.
"Error: expected at most two arguments, unexpected arguments: –-namespace, keda, –-version, 2.0.0-beta1.2"


tomkerkhove (approved):
Thanks!

PR #303, opened 2020-11-05
Update scaling-deployments.md

whaakman:
Added quotes as a string is required instead of an integer


tomkerkhove (approved):

whaakman:
Ignore this. I made a booboo I was storing these values under metadata (spec.triggers.metadata.maxReplicaCount) of course this accepts a string. Works fine if I use the proper configuration(ScaledObject.spec.). 

Apoligies!

PR #321, opened 2020-12-06
Improve RabbitMQ queue autoscaler doc

mavimo:
- Fix reference to Redis server instad of RabbitMQ
 - make more clear that VirtualHost is a parameter


tomkerkhove (approved):
Thanks for fixing the docs!

PR #331, opened 2020-12-12
add readOnlyRootFilesystem to FAQ

lambohamp:
Added instructions on how to enable readOnlyRootFilesystem to FAQ as per [#103](https://github.com/kedacore/charts/issues/103)


tomkerkhove (approved):
LGTM, what do you think @zroubalik ?

zroubalik (commented):
Looking good! Though we should add this to 2.1 as well.

PR #332, opened 2020-12-15
Fix amqp rabbitmq-queue documentation

Cirrith:
The protocol for amqp was misspelled causing Keda to error on copied config.


tomkerkhove (approved):
LGTM - Thanks for the fix!

zroubalik:
@tomkerkhove we should fix that in 2.1 as well

tomkerkhove:
Sorry about that - https://github.com/kedacore/keda-docs/pull/334

PR #333, opened 2020-12-16
Modify the value of name in ScaledObject metadata.

Shubham82:
Signed-off-by: Shubham82 <shubham.kuchhal@india.nec.com>

The term used for value of name in scaledobject metadata is {deployment-name} which is incorrect. instead it should be {scaled-object-name}. i have changed in 1.4, 1.5, 2.0, 2.1 documents.


tomkerkhove (approved):
Well-spotted, thanks!

PR #340, opened 2020-12-28
Documentation for ClusterTriggerAuthentication

coderanger:
Docs PR for https://github.com/kedacore/keda/pull/1452

Relates to https://github.com/kedacore/keda/issues/1469

### DIFF
--- a/content/docs/2.1/concepts/authentication.md
+++ b/content/docs/2.1/concepts/authentication.md
@@ -8,7 +8,8 @@ Often a scaler will require authentication or secrets and config to check for ev
 KEDA provides a few secure patterns to manage authentication flows:
 
 * Configure authentication per `ScaledObject`
-* Re-use credentials or delegate authentication with `TriggerAuthentication`
+* Re-use per-namespace credentials or delegate authentication with `TriggerAuthentication`
+* Re-use global credentials with `ClusterTriggerAuthentication`
 
 ## Defining secrets and config maps on ScaledObject
 
@@ -142,6 +143,17 @@ Each `TriggerAuthentication` is defined in one namespace and can only be used by
 ```
 
 By default, Secrets loaded from a `secretTargetRef` must be in the same namespace as KEDA is deployed in (usually `keda`). This can be overridden by setting a `$KEDA_CLUSTER_OBJECT_NAMESPACE` environment variable for the `keda-operator` container.
+
+Defining a `ClusterTriggerAuthentication` works almost identically to a `TriggerAuthentication`, except there is no `metadata.namespace` value:
+
+```yaml
+apiVersion: keda.sh/v1alpha1
+kind: ClusterTriggerAuthentication
+metadata:
+  name: {cluster-trigger-auth-name}
+spec:
+  # As before ...
+```
 
 ## Authentication parameters
 
--- a/content/docs/2.1/concepts/authentication.md
+++ b/content/docs/2.1/concepts/authentication.md
@@ -131,6 +131,22 @@ Every parameter you define in `TriggerAuthentication` definition does not need t
       name: {trigger-authencation-name} # this may define other params not defined in metadata
 ```
 
+## Authentication scopes: Namespace vs. Cluster
+
+Each `TriggerAuthentication` is defined in one namespace and can only be used by a `ScaledObject` in that same namespace. For cases where you want to share a single set of credentials between scalers in many namespaces, you can instead create a `ClusterTriggerAuthentication`. As a global object, this can be used from any namespace. To set a trigger to use a `ClusterTriggerAuthentication`, add a `kind` field to the authentication reference:
+
+```yaml
+    authenticationRef:
+      name: {trigger-authencation-name}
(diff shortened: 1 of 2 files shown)


content/docs/2.1/concepts/authentication.md
```diff
@@ -223,3 +223,17 @@ You can tell KEDA to use Kiam via `podIdentity.provider`.
...
+
+When referencing a `ClusterTriggerAuthentication` in a trigger, specify the Kind:
+
+```yaml
+    authenticationRef:
+      name: {trigger-authencation-name}
+      kind: ClusterTriggerAuthentication
+```
+
+The namespace to fetch secrets from can be overridden by setting a `$KEDA_CLUSTER_OBJECT_NAMESPACE` environment variable for the `keda-operator` container.
```
tomkerkhove, line 239: Wouldn't it make sense to make this configurable in the ClusterTriggerAuthentication itself through something like this?

```yaml
secretTargetRef:                          # Optional.
  - parameter: connectionString           # Required - Defined by the scale trigger
    name: my-keda-secret-entity           # Required.
    key: azure-storage-connectionstring   # Required.
    namespace: <namespace-name-here>      # Optional.
```

tomkerkhove: It would be good to add a section at the top that describes the different scopes before diving into it I think.

That way, it's not "hidden" at the bottom of the docs.

tomkerkhove (changes requested):
LGTM, but I'd improve it a bit more.

content/docs/2.1/concepts/authentication.md
coderanger: Making this configurable per object makes me nervous about security implications. That could be used to read any secret in any namespace, possibly allowing for unexpected data access. Having it be limited to just one global namespace makes it easier for an admin to think through the risk factors.

Will rework the section order :)

tomkerkhove: Makes sense.

Thanks!

coderanger:
The wording in the new `Authentication parameters` heading is a bit awkward but gets the job done (and keeps a semi-reasonable TOC hierarchy).

content/docs/2.1/concepts/authentication.md
```diff
@@ -131,6 +131,22 @@ Every parameter you define in `TriggerAuthentication` definition does not need t
       name: {trigger-authencation-name} # this may define other params not defined in metadata
 ```
 
+## Authentication scopes: Namespace vs. Cluster
+
+Each `TriggerAuthentication` is defined in one namespace and can only be used by a `ScaledObject` in that same namespace. For cases where you want to share a single set of credentials between scalers in many namespaces, you can instead create a `ClusterTriggerAuthentication`. As a global object, this can be used from any namespace. To set a trigger to use a `ClusterTriggerAuthentication`, add a `kind` field to the authentication reference:
```
zroubalik, line 136: Would be nice to add some details on `ClusterTriggerAuthentication` contents, ie. that it is basically the same object as TA, either describe this or adda trimmed example of CTA

zroubalik (commented):
Looking good, thanks! Minor nits:

Could you please update this part of the docs as well?
https://github.com/kedacore/keda-docs/blob/master/content/docs/2.1/concepts/_index.md#custom-resources-crd

The beginning of content/docs/2.1/concepts/authentication.md (section around `* Re-use credentials or delegate authentication with TriggerAuthentication`)

coderanger:
Not sure what's up with the failed build. Transient issue?

zroubalik:
@coderanger yeah, we can ignore those. Thanks for the update, it is looking good!

zroubalik (approved):
LGTM, waiting for @tomkerkhove for a final sweep

content/docs/2.1/concepts/authentication.md
```diff
@@ -131,6 +132,33 @@ Every parameter you define in `TriggerAuthentication` definition does not need t
...
+
+```yaml
+    authenticationRef:
+      name: {trigger-authencation-name}
+      kind: ClusterTriggerAuthentication
+```
+
+By default, Secrets loaded from a `secretTargetRef` must be in the same namespace as KEDA is deployed in (usually `keda`). This can be overridden by setting a `$KEDA_CLUSTER_OBJECT_NAMESPACE` environment variable for the `keda-operator` container.
```
tomkerkhove, line 145: Does it really need the `$` in the environment variable?
```suggestion
By default, Secrets loaded from a `secretTargetRef` must be in the same namespace as KEDA is deployed in (usually `keda`). This can be overridden by setting a `KEDA_CLUSTER_OBJECT_NAMESPACE` environment variable for the `keda-operator` container.
```

tomkerkhove (approved):
LGTM, just added one question

content/docs/2.1/concepts/authentication.md
coderanger: I usually put it as a secondary reminder than it's an env var. That can backfire if people take it literally though.

PR #344, opened 2021-01-04
Provide documentation on KEDA resources

tomkerkhove:

Inspired by https://v1-rc2.docs.dapr.io/operations/hosting/kubernetes/kubernetes-production/

Relates to https://github.com/kedacore/keda/issues/1450


zroubalik (approved):
Just FYI, those values aren't based on some hard data, they could be imho tweaked a little bit based on some proper research (potentially smaller Requests and Limits).

But for now, it is looking good

PR #346, opened 2021-01-08
Clarify queueLength parameter of queue-based scalers

basduineveld:
Per discussion https://github.com/kedacore/keda/discussions/1395. The description of the queueLength parameter of queue-based scalers is short and quite technical. The practical implication that it should be set to the number of queue messages one pod processes was not clear to me until I got a working configuration using trial and error. This PR is intended to clarify that.

### DIFF
--- a/content/docs/2.0/scalers/artemis.md
+++ b/content/docs/2.0/scalers/artemis.md
@@ -31,7 +31,7 @@ triggers:
 - `queueName` - Name of the queue to check for the number of messages available.
 - `brokerName` - Name of the broker as defined in Artemis.
 - `brokerAddress` - Address of the broker.
-- `queueLength` - Average target value to trigger scaling actions. (default: 10)
+- `queueLength` - Target value for queue length passed to Horizontal Pod Autoscaler (HPA). Example: if one pod can handle 10 messages, set the queue length target to 10. If the actual number of messages in the queue is 30, HPA scales to 3 pods. (default: 10)
 - `restApiTemplate` - Template to build REST API url to get queue size.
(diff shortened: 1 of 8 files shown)


tomkerkhove:
Thanks for the improvement, definately helpful!

However, I need to pull in @TsuyoshiUshio because for ScaledJob this is not the case I think?

basduineveld:
@tomkerkhove I'm actually using this for ScaledJobs myself. The same logic applies to job instances.

TsuyoshiUshio:
Hi @basduineveld 
Thank you for the contribution. When I started it, I also confused. :) I believe this is a great contribution!
At least, the explanation is correct for the ScaledJob. Only the correction might be, For the Scaled Job, it is not using HPA. ScaledJob has original logic to watch/scale the job. 
@tomkerkhove  queueLength is the same meaning on ScaledObject based deployment scaling? If so, sounds great!

basduineveld:
@TsuyoshiUshio Thank you for the clarification. If I replace "HPA" by a more generic "the scaler", would that make it correct for both ScaledObject and ScaledJob?

I think the page https://keda.sh/docs/2.0/concepts/scaling-jobs/ needs to be updated as well. The correct calculation of the number of jobs created is shown in the table, but the equations below the table don't include all 4 relevant parameters. Can you help me correct these equations? I think the default strategy should be `min(actualQueueLength / targetQueueLength - runningJobCount, maxReplicaCount)`, but I'm not sure about the custom and accurate strategies.

tomkerkhove:
> @TsuyoshiUshio Thank you for the clarification. If I replace "HPA" by a more generic "the scaler", would that make it correct for both ScaledObject and ScaledJob?
Jup, sounds good to me!

> I think the page [keda.sh/docs/2.0/concepts/scaling-jobs](https://keda.sh/docs/2.0/concepts/scaling-jobs/) needs to be updated as well. The correct calculation of the number of jobs created is shown in the table, but the equations below the table don't include all 4 relevant parameters. Can you help me correct these equations? I think the default strategy should be `min(actualQueueLength / targetQueueLength - runningJobCount, maxReplicaCount)`, but I'm not sure about the custom and accurate strategies.

I'll leave this up to @TsuyoshiUshio

zroubalik (approved):

PR #350, opened 2021-01-19
Added double quotes to value.

Shubham82:
Added the double quotes to the value as it is of string type not integer. if not mentioned then it would give this on creation of scaledobject:
`The ScaledObject "cpu-scaledobject" is invalid: spec.triggers.metadata.value: Invalid value: "integer": spec.triggers.metadata.value in body must be of type string: "integer"`

Signed-off-by: Shubham Kuchhal <shubham.kuchhal@india.nec.com>

### DIFF
--- a/content/docs/2.0/scalers/cpu.md
+++ b/content/docs/2.0/scalers/cpu.md
@@ -21,7 +21,7 @@ triggers:
   metadata:
     # Required
     type: Value/ Utilization/ AverageValue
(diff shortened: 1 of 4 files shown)


Shubham82:
Hi @zroubalik @tomkerkhove 
Could you Please take a look.

Thanks

tomkerkhove (approved):
Thanks for the fix!

PR #381, opened 2021-02-22
Support for HA with leader election in our operator

mtparet:


tomkerkhove:
This is not supported, but adding @zroubalik to confirm

mtparet:
I see the args and boolean variable to enable leader election https://github.com/kedacore/keda/blob/main/main.go#L64, that's why I thought it was supported.

mtparet:
I tested to run 2 replicas of Keda (with leader election activated).
Only one of the 2 took the lead and was active. When I killed it, the second took the lead and become active.

zroubalik:
Yeah, you can deploy multiple KEDA Operators, but it's just Failover support. As @mtparet mentioned, only one of the pods can actually serve the load. So when we mention 2 replicas, as suggested in this PR, users might get the impression that the load is balanced across 2 Operator replicas, which is not true.

So you can run more replicas of KEDA Operator, but not sure if it does have a real benefit, because the same scenario (2 replicas) with killing one Operator pod and serving the load by another Pod. could actually be handled by k8s scheduler (in a setup with just 1 replica). When k8s scheduler finds out that the pod is down, it will spin another one to serve the requests, but I concur that the delay could be a few seconds longer.

If we want to document support of 2 replicas, we should mention the actual behaviour as well (in some note or description).

tomkerkhove:
So here's what I'd suggest:
- We leave the Helm chart as is, because it would be misleading indeed
- We add more context in the docs

How does that sound?

zroubalik:
I am fine with adding this option to Helm Chart, in case someone would like to use Leader election mechanism. But it should come with solid docs.

content/docs/2.1/operate/cluster.md
```diff
@@ -12,7 +12,7 @@ Here is an overview of all KEDA deployments and the supported replicas:
 
 | Deployment     | Support Replicas        | Reasoning                     |
 |----------------|-------------------------|-------------------------------|
-| Operator       | 1                       |                               |
+| Operator       | 2                       |                               |
```
tomkerkhove, line 15:
```suggestion
| Operator       | 1                       | While you can run more replicas of our operator, only one operator instance will be the leader and serving traffic. You can run multiple replicas, but they will not improve the performance of KEDA. This behavior is how operators work. |
```

tomkerkhove (commented):
What do you think about this @mtparet & @zroubalik?

I would port the same comment to our Helm chart to make it more explicit.

PS: Can you make this change in our 2.0 & 2.2 docs as well please?

content/docs/2.1/operate/cluster.md
zroubalik: What about, or something like this more polished:
```
While you can run more replicas of our operator, only one operator instance will be the leader and serving traffic. You can run multiple replicas, but they will not improve the performance of KEDA, it could only reduce a downtime during a failover.
```

tomkerkhove: Even better!

tomkerkhove:
@mtparet What do you think of the proposed changes? Would you mind updating our recent versions as well please?

mtparet:
Doc updated for all new versions.

mtparet:
Sorry for not responding during a long time, changes proposed are good, thank you!

JorTurFer:
I think that the PR only should change v2.6 because the v2.5 is already released (and previous versions will not be modified)
WDYT @kedacore/keda-helm-contributors ? Have we got plans of a chart release before next release?

tomkerkhove:
This relates to https://github.com/kedacore/charts/pull/125.

We can document in all versions, but we should emphasize that this is only possible as of v2.6 if you are using Helm (because we support other installation flavors as well).

content/docs/2.1/operate/cluster.md
```diff
@@ -36,12 +36,12 @@ Here is an overview of all KEDA deployments and the supported replicas:
 
 | Deployment     | Support Replicas        | Reasoning                     |
 |----------------|-------------------------|-------------------------------|
-| Operator       | 1                       |                               |
+| Operator       | 2                | While you can run more replicas of our operator, only one operator instance will be the leader and serving traffic. You can run multiple replicas, but they will not improve the performance of KEDA, it could only reduce a downtime during a failover.|
```
tomkerkhove, line 39:
```suggestion
| Operator       | 2                | While you can run more replicas of our operator, only one operator instance will be the leader and serving traffic. You can run multiple replicas, but they will not improve the performance of KEDA, it could only reduce downtime during a failover.<br /><br />This is only supported as of KEDA v2.6 if you are using our Helm chart.|
```

content/docs/2.2/operate/cluster.md
```diff
@@ -39,11 +39,11 @@ Here is an overview of all KEDA deployments and the supported replicas:
 | Deployment     | Support Replicas | Reasoning                                                                                                        |
 | -------------- | ---------------- | ---------------------------------------------------------------------------------------------------------------- |
 | Metrics Server | 1                | Limitation in [k8s custom metrics server](https://github.com/kubernetes-sigs/custom-metrics-apiserver/issues/70) |
-| Operator       | 1                |                                                                                                                  |
+| Operator       | 2                | While you can run more replicas of our operator, only one operator instance will be the leader and serving traffic. You can run multiple replicas, but they will not improve the performance of KEDA, it could only reduce a downtime during a failover.|
```
tomkerkhove, line 42:
```suggestion
| Operator       | 2                | While you can run more replicas of our operator, only one operator instance will be the leader and serving traffic. You can run multiple replicas, but they will not improve the performance of KEDA, it could only reduce downtime during a failover.<br /><br />This is only supported as of KEDA v2.6 if you are using our Helm chart.|
```

content/docs/2.3/operate/cluster.md
```diff
@@ -39,7 +39,7 @@ Here is an overview of all KEDA deployments and the supported replicas:
 | Deployment     | Support Replicas | Reasoning                                                                                                        |
 | -------------- | ---------------- | ---------------------------------------------------------------------------------------------------------------- |
 | Metrics Server | 1                | Limitation in [k8s custom metrics server](https://github.com/kubernetes-sigs/custom-metrics-apiserver/issues/70) |
-| Operator       | 1                |                                                                                                                  |
+| Operator       | 2                | While you can run more replicas of our operator, only one operator instance will be the leader and serving traffic. You can run multiple replicas, but they will not improve the performance of KEDA, it could only reduce a downtime during a failover.|
```
tomkerkhove, line 42:
```suggestion
| Operator       | 2                | While you can run more replicas of our operator, only one operator instance will be the leader and serving traffic. You can run multiple replicas, but they will not improve the performance of KEDA, it could only reduce downtime during a failover.<br /><br />This is only supported as of KEDA v2.6 if you are using our Helm chart.|
```

content/docs/2.4/operate/cluster.md
```diff
@@ -39,7 +39,7 @@ Here is an overview of all KEDA deployments and the supported replicas:
 | Deployment     | Support Replicas | Reasoning                                                                                                        |
 | -------------- | ---------------- | ---------------------------------------------------------------------------------------------------------------- |
 | Metrics Server | 1                | Limitation in [k8s custom metrics server](https://github.com/kubernetes-sigs/custom-metrics-apiserver/issues/70) |
-| Operator       | 1                |                                                                                                                  |
+| Operator       | 2                | While you can run more replicas of our operator, only one operator instance will be the leader and serving traffic. You can run multiple replicas, but they will not improve the performance of KEDA, it could only reduce a downtime during a failover.|
```
tomkerkhove, line 42:
```suggestion
| Operator       | 2                | While you can run more replicas of our operator, only one operator instance will be the leader and serving traffic. You can run multiple replicas, but they will not improve the performance of KEDA, it could only reduce downtime during a failover.<br /><br />This is only supported as of KEDA v2.6 if you are using our Helm chart.|
```

content/docs/2.5/operate/cluster.md
```diff
@@ -39,7 +39,7 @@ Here is an overview of all KEDA deployments and the supported replicas:
 | Deployment     | Support Replicas | Reasoning                                                                                                        |
 | -------------- | ---------------- | ---------------------------------------------------------------------------------------------------------------- |
 | Metrics Server | 1                | Limitation in [k8s custom metrics server](https://github.com/kubernetes-sigs/custom-metrics-apiserver/issues/70) |
-| Operator       | 1                |                                                                                                                  |
+| Operator       | 2                | While you can run more replicas of our operator, only one operator instance will be the leader and serving traffic. You can run multiple replicas, but they will not improve the performance of KEDA, it could only reduce a downtime during a failover.|
```
tomkerkhove, line 42:
```suggestion
| Operator       | 2                | While you can run more replicas of our operator, only one operator instance will be the leader and serving traffic. You can run multiple replicas, but they will not improve the performance of KEDA, it could only reduce downtime during a failover.<br /><br />This is only supported as of KEDA v2.6 if you are using our Helm chart.|
```

tomkerkhove (changes requested):
As per the discussion, let's be explicit

content/docs/2.1/operate/cluster.md
arschles, line 39: here's some possible alternative wording. I won't be offended if you want to ignore this 😄 

```suggestion
| Operator       | 2                | While you can run more replicas of our operator, only one operator instance will be active. The rest will be standing by, which may reduce downtime during a failure. Multiple replicas will not improve the performance of KEDA. |
```

tomkerkhove (approved):
LGTM, Merging to temp branch to fix DCO

PR #386, opened 2021-03-03
Update OpenStack Swift scaler contract

pdominguite:
Signed-off-by: Pedro Felipe Dominguite (<p.dominguite@sidi.org.br>)

### Changes

- The `swiftURL` is now an optional parameter since it can be retrieved from the catalog.
- The optional parameter `regionName` was added.
- Since the scaler only supports Keystone v3, the parameter `/v3` shouldn't be passed to the URL path as before.

Relates to: [#1652](https://github.com/kedacore/keda/pull/1652)


zroubalik (approved):
LGTM

PR #388, opened 2021-03-03
document new `publishRate` trigger

rwkarg:
Corresponding code change: https://github.com/kedacore/keda/pull/1653

### DIFF
--- a/content/docs/2.2/scalers/rabbitmq-queue.md
+++ b/content/docs/2.2/scalers/rabbitmq-queue.md
@@ -44,7 +44,7 @@ Some parameters could be provided using environmental variables, instead of sett
 
 > 💡 **Note:** `host`/`hostFromEnv` has an optional vhost name after the host slash which will be used to scope API request.
 
-> 💡 **Note:** `mode` MessageRate requires protocol `http`.
+> 💡 **Note:** `mode: MessageRate` requires protocol `http`.
 
 > ⚠ **Important:** if you have unacknowledged messages and want to have these counted for the scaling to happen, make sure to utilize the `http` REST API interface which allows for these to be counted.
 
--- a/content/docs/2.2/scalers/rabbitmq-queue.md
+++ b/content/docs/2.2/scalers/rabbitmq-queue.md
@@ -17,8 +17,8 @@ triggers:
   metadata:
     host: amqp://localhost:5672/vhost # Optional. If not specified, it must be done by using TriggerAuthentication.
     protocol: auto # Optional. Specifies protocol to use, either amqp or http, or auto to autodetect based on the `host` value. Default value is auto.
-    queueLength: '20' # Optional. Queue length target for HPA. Default: 20 messages if `publishRate` is not specified, or 0 if `publishRate` is specified
-    publishRate: '100' # Optional. Publish/sec. target on the queue for HPA. Requires http host/protocol
+    mode: QueueLength # QueueLength or MessageRate
+    value: 100 # message backlog or publish/sec. target per instance
     queueName: testqueue
     vhostName: / # Optional. If not specified, use the vhost in the `host` connection string.
     # Alternatively, you can use existing environment variables to read configuration from:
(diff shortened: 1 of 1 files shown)


tomkerkhove:
Just a thought: Do we use this approach, or do we introduce a `mode` which defines how it should work?

Thoughts on this @zroubalik?

rwkarg:
I had talked with @coderanger [on Slack](https://kubernetes.slack.com/archives/CKZJ36A5D/p1614644063093000?thread_ts=1614637755.088700&cid=CKZJ36A5D) about this.

`queueLength` needs to be kept to prevent a breaking change from existing ScaledObject definitions. That leaves the following options:

1. Add a `mode`-like field to make `queueLength` really mean publish rate (that seems confusing)
2. Add a generic `triggerValue` field and a `triggerType` field (with queue length and publish rate possible values). This would give two ways to set a queue length trigger (`queueLength: X` or `triggerValue: X, triggerType: QUEUE_LENGTH`) which is not ideal but prevents breaking changes. Publish rate would have only one option for config (`triggerValue: X, triggerType: PUBLISH_RATE`)
3. Separate `queueLength` and `publishRate` fields that are mutually exclusive.

If this were a completely new scaler then only having the `triggerValue` and `triggerType` fields would probably be the way to go, but we've already got existing uses with the `queueLength` field configured.

I'm happy to align with another direction if that would be more consistent with how other scalers are set up. The one thing I don't want to do is introduce a change that would break existing users.

tomkerkhove:
I was just thinking of this so that it's clear what you want to use and simplify the validation of the config without relying on assumptions. I don't like assumptions as it is not always straightforward if you are new to it or not aware of how it works without reading the code/docs carefully.

```yaml
  metadata:
    mode: QueueDepth
    queueLength: '20' # Optional. Queue length target for HPA. Default: 20 messages
    # Standard config
```

```yaml
  metadata:
    mode: MessageRate
    publishRate: '100' # Optional. Publish/sec. target on the queue for HPA. Requires http host/protocol
    # Standard config
```

That way, you could also use these configs:

```yaml
  metadata:
    mode: QueueDepth
    # Standard config
```

```yaml
  metadata:
    mode: MessageRate
    # Standard config
```

This allows people to rely on the standard value, but still, define the trigger behavior.

When using both fields, this can be tricky if you are new (and maybe didn't read the docs correctly); but if you use a flag then it is explicit. The only caveat would be, though, that if `mode` is not provided that it defaults to queue depth.

That said, this was just my idea; I'm happy to go with the current if everyone prefers that.

> `mode` is just an idea, can be `triggerMode` as well

coderanger:
Using both fields is impossible, it's disjoint mode. And I'm not sure any default value for `publishRate` makes sense so I think this is overly verbose.

rwkarg:
What if we kept `queueLength` but marked it obsolete until the next major version when it would be removed. Write out a log/event (that nobody will probably read...) to update to using the `mode` version. 

This would leave three usage patterns:

`queueLength` (obsolete but supported until the next major version)
```
 metadata:
    queueLength: "10"
```

mode: QueueDepth (preferred over `queueLength`)
```
 metadata:
    mode: "QueueDepth"
    value: "10" # this could still default to 20 when omitted, if desired
```

mode: MessageRate
```
 metadata:
    mode: "MessageRate"
    value: "100" # required with `mode: "MessageRate"`
```

rwkarg:
Or we could just remove `queueLength` from the documentation and only surface the `mode:` variants, but still accept the `queueLength` config forever as an undocumented back compat. config?

tomkerkhove:
I like the deprecation suggestion!

coderanger:
> Or we could just remove `queueLength` from the documentation and only surface the `mode:` variants, but still accept the `queueLength` config forever as an undocumented back compat. config?

-1 on undocumented config. Folks may have a running Keda and be trying to understand it, and having undocumented legacy flags makes that much harder. But a clear "legacy, deprecated flags" section is :+1:

rwkarg:
@coderanger I've updated to include the deprecated `queueLength` in the docs but call it out as deprecated and how to migrate. It is still fully supported with the same behavior, but it is mutually exclusive to the `mode`/`value` config.

content/docs/2.2/scalers/rabbitmq-queue.md
```diff
@@ -27,23 +28,28 @@ triggers:
...
-
+- `queueLength`: DEPRECATED! Use `mode: QueueLength` and `value: ##` instead. Target value for queue length passed to the scaler. Example: if one pod can handle 10 messages, set the queue length target to 10. If the actual number of messages in the queue is 30, the scaler scales to 3 pods. Default is 20 unless `publishRate` is specified, in which case `queueLength` is disabled for this trigger.
 
 Some parameters could be provided using environmental variables, instead of setting them directly in metadata. Here is a list of parameters you can use to retrieve values from environment variables:
 
 - `hostFromEnv`: The host and port of the Redis server, similar to `host`, but reads it from an environment variable on the scale target.
 
 > 💡 **Note:** `host`/`hostFromEnv` has an optional vhost name after the host slash which will be used to scope API request.
 
+> 💡 **Note:** `mode` MessageRate requires protocol `http`.
```
tomkerkhove, line 47:
```suggestion
> 💡 **Note:** `mode: MessageRate` requires protocol `http`.
```

tomkerkhove (commented):
LGTM, WDYT @zroubalik @ahmelsayed ?

tomkerkhove (approved):

tomkerkhove:
Thanks for this nice change btw, @rwkarg!

zroubalik (approved):
LGTM, just the small nit:  https://github.com/kedacore/keda-docs/pull/388#discussion_r589193767

tomkerkhove:
Thanks a ton @rwkarg!

PR #390, opened 2021-03-04
docs: Provide overview of required ports to be accessible

tomkerkhove:
Provide overview of required ports to be accessible for KEDA to work and restructured the ordering to show the most important guidance first, followed by more advanced topics.

Relates to https://github.com/kedacore/keda-docs/discussions/389
Closes https://github.com/kedacore/keda-docs/issues/384


tomkerkhove:
It's very minimal but let me know what you think @rwkarg @nidiculageorge

rwkarg:
Should clarify what the source -> destination is:

For GKE specifically, the firewall is from the Cluster Control Plane -> port 6443 on the Pod IP range.

For other non-GKE clusters it is from the Cluster Control Plane -> port 443 on the Service IP range (not sure how that's specified on every other managed provider or self-service cluster, or which ones require specific firewall rules to be added)

tomkerkhove:
I've included it, feel free to have a look

zroubalik (approved):

PR #392, opened 2021-03-04
Updating liiklus document: adding optional to lagthreshold

Ritikaa96:
Presently in liiklus document's Example the metadata shows lagThreshold in required field:
```
 metadata:
      # Required
      address: localhost:6565
      group: my-group       # Make sure that this consumer group name is the same one as the one that is consuming topics
      topic: test-topic
      lagThreshold: "50"
```

Changing it to this:
```
metadata:
      # Required
      address: localhost:6565
      group: my-group       # Make sure that this consumer group name is the same one as the one that is consuming topics
      topic: test-topic
      # Optional
      lagThreshold: "50"    # Default value is set to 10
```

Signed-off-by: Ritikaa96 <ritika@india.nec.com>


tomkerkhove (approved):
Thanks!

PR #410, opened 2021-03-30
quote value in rabbitmq trigger example

PaulFarver:
The current documentation suggests that the value field in the rabbitmq trigger is an integer, which can lead to confusion.


zroubalik:
@PaulFarver thanks! could you please update the other versions of this doc as well? (in `content/docs/*`)

PaulFarver:
@zroubalik thanks for the quick reply. I've updated the 2.3 version as well. :)

zroubalik (approved):
LGTM, thanks!

PR #423, opened 2021-04-22
add-graphite-scaled-docs

shuoliu1:
add-graphite-scaled-docs

https://github.com/kedacore/keda/pull/1749


shuoliu1:
sorry not sig

PR #429, opened 2021-04-30
add checkpointStrategy parameter to event-hub

christle:
Signed-off-by: Christian Leinweber <christian.leinweber@maibornwolff.de>


content/docs/2.3/scalers/azure-event-hub.md
```diff
@@ -30,6 +30,11 @@ triggers:
 - `consumerGroup` - Consumer group of event hub consumer. (default: `$default`)
 - `unprocessedEventThreshold` - : Average target value to trigger scaling actions. (default: 64)
 - `blobContainer` - Container name to store checkpoint. This is needed when a using an Event Hub application written in dotnet or java, and not an Azure function
+- `checkpointStrategy` - configure the checkpoint behaviour of different eventhub sdk's. (default: `""`)
+    - `""`: the default behaviour works for C#, older Python and Java eventhub implementations if a `blobContainer` is configured.
+    - `AzureWebJob`: suitable for Azure Functions. This is the default setting, if `blobcontainer` is empty.
```
tomkerkhove, line 35: Now that I read this again, what if we just call this `AzureFunctions`?

content/docs/2.3/scalers/azure-event-hub.md
```diff
@@ -30,6 +30,11 @@ triggers:
 - `consumerGroup` - Consumer group of event hub consumer. (default: `$default`)
 - `unprocessedEventThreshold` - : Average target value to trigger scaling actions. (default: 64)
 - `blobContainer` - Container name to store checkpoint. This is needed when a using an Event Hub application written in dotnet or java, and not an Azure function
+- `checkpointStrategy` - configure the checkpoint behaviour of different eventhub sdk's. (default: `""`)
+    - `""`: the default behaviour works for C#, older Python and Java eventhub implementations if a `blobContainer` is configured.
```
tomkerkhove, line 34: We might want to rename this to `Default` or `NotSpecified`?

tomkerkhove (commented):
What do you think @zroubalik ?

content/docs/2.3/scalers/azure-event-hub.md
christle: I've thought about it too. I think the corresponding lib is AzureWebJob. But in this context it is probably only used by Azure Functions. I am going to rename this

content/docs/2.3/scalers/azure-event-hub.md
christle: you mean simply instead of an empty string, "NotSpecified" as parameter description, or also an explicit naming of the parameter in code? I wouldn't be so happy with "default" as the parameter name in the code. I think more and more libs will implement the metadata approach. Currently this is the default setting, due to backward compatibility

tomkerkhove: I think I get it now, wit this line you mean that it will use this when it's not specified, is that correct?

So there's two things:
- `blobContainer` that is used for checkpointing and works for C#, older Python and Java eventhub implementations
- If nothing is specified, then `blobContainer` is used

Is that correct? If so, I would document `blobContainer` and add `(default: blobContainer)` on line 33 where we explain what `checkpointStrategy` is. Makes sense?

christle: The blobContainer description is currently not 100 percent correct. A `blobContainer` name is for all checkpointStrategies required except for azure Functions. The checkpointing for Azure Function is the only implementation which sets the containername automatically and cannot be overridden. So the old scaler logic was simply:

if blobContainer == "" then
 --> azureFunction
 else
  --> everything else (which works for c# and older frameworks)
end

So for backward compatibility, when no strategy is specified, then the "old checkpointing" (else path above) is taken.
For the same reason, if no blobContainer name is set, the azure Function checkpointing take place. Alternatively is it now possible to set the "azureFunction" strategy explicit. 

The new strategies requires a value for `checkpointStrategy`, no backward compatibility is necessary.

I can try to write a more clear description in the evening. The other possibility is, we require with keda 2.3 an explicit value in checkpointStrategy, and add a breaking Change comment. Then we need also an explicit name for the "old checkpointing" behavior. Maybe C#, (but this is also true for older implementations of other languages).

FYI: Yesterday someone mentioned  on the code PR that the new blobMetadata strategy works also for the JS implementation. We should add it in the list above, too.

tomkerkhove: Let's use the approach with non-breaking changes but maybe provide some more information in the docs then indeed.

Thanks and sorry for dragging on but it's a pretty "big" change.

christle:
I've add more details to the parameter list.

additionally i found out there are also 2 implementations for C#. I have tested both, and the behavior is like Python and Java. One Implementation based on the EventProcessorHost and a new one, which is called EventProcessorClient. The EventProcessorHost works with the default path(no checkpointStrategy specified), the EventProcessorClient with the `blobMetadata` strategy (which is great, i think). Microsoft called the EventProcessorHost implementation "legacy" (see: https://docs.microsoft.com/de-de/azure/event-hubs/event-hubs-dotnet-standard-get-started-send-legacy).

content/docs/2.3/scalers/azure-event-hub.md
```diff
@@ -27,10 +27,19 @@ triggers:
...
-- `blobContainer` - Container name to store checkpoint. This is needed when a using an Event Hub application written in dotnet or java, and not an Azure function
+- `blobContainer` - Container name to store checkpoint. This is needed for every `checkpointStrategy` except of `AzureFunction`. With Azure Functions the `blobContainer` is autogenerated and cannot be overridden.
+- `checkpointStrategy` - configure the checkpoint behaviour of different Event Hub SDK's. (default: `""`)
+    - `""`: if no checkpointStrategy is specified, the Event Hub scaler is in backward compatibility mode and able to scale older implementations of C#, Python or Java Event Hub SDK's. (see "Legacy checkpointing"). if this behaviour should be used, `blobContainer` is also required.
+    - `azureFunction`: suitable for Azure Functions. This is automatically the default setting, if `blobcontainer` is empty.
+    - `blobMetadata`: For all implementations which stores checkpoint informations on blob metadata like current C#, Python, Java and JavaScript Event Hub SDK's.
+    - `goSdk` For all implementations which use the Golang SDK checkpointing like Dapr.io
```
tomkerkhove, line 37: I've reshuffled it a bit, what do you think?
```suggestion
- `checkpointStrategy` - configure the checkpoint behaviour of different Event Hub SDKs. (default: `""`)
    - `azureFunction`: Suitable for Azure Functions & Azure WebJobs SDK. This is the default setting, when `blobcontainer` is not specified.
    - `blobMetadata`: For all implementations that store checkpoint informations on blob metadata such as current C#, Python, Java and JavaScript Event Hub SDKs.
    - `goSdk` For all implementations using the [Golang SDK](https://github.com/Azure/azure-event-hubs-go)'s checkpointing, for example Dapr
    - When no checkpoint strategy is specified, the Event Hub scaler will use backwards compatibility and able to scale older implementations of C#, Python or Java Event Hub SDKs. (see "Legacy checkpointing"). If this behaviour should be used, `blobContainer` is also required.
```

content/docs/2.3/scalers/azure-event-hub.md
```diff
@@ -27,10 +27,19 @@ triggers:
 
 - `connectionFromEnv` - Name of the environment variable your deployment uses to get the connection string appended with `EntityPath=<event_hub_name>`
 - `storageConnectionFromEnv` - Name of the environment variable that provides connection string for Azure Storage Account to store checkpoint. As of now the Event Hub scaler only reads from Azure Blob Storage.
-- `consumerGroup` - Consumer group of event hub consumer. (default: `$default`)
+- `consumerGroup` - Consumer group of Event Hub consumer. (default: `$default`)
```
tomkerkhove, line 30:
```suggestion
- `consumerGroup` - Consumer group of Azure Event Hub consumer. (default: `$default`)
```

content/docs/2.3/scalers/azure-event-hub.md
```diff
@@ -27,10 +27,19 @@ triggers:
...
+    - `azureFunction`: suitable for Azure Functions. This is automatically the default setting, if `blobcontainer` is empty.
+    - `blobMetadata`: For all implementations which stores checkpoint informations on blob metadata like current C#, Python, Java and JavaScript Event Hub SDK's.
+    - `goSdk` For all implementations which use the Golang SDK checkpointing like Dapr.io
 
+**Legacy checkpointing**  
+C# applications, which use the `Microsoft.Azure.EventHubs` package, Java applications which use the `azure-eventhubs-eph` package  or Python applications which use the `azure-eventhub` below `v5` are supported if no `checkpointStrategy`is specified. All these legacy implementations based on the `EventProcessorHost` Client, which stores checkpoint information as blobcontainer content
+
+C# Applications which use the current `Azure.Messaging.EventHubs` package, Java applications which use the current `azure-messaging-eventhubs` package or Python applications since `azure-eventhub v5`, have to set the `checkpointStrategy`to `blobMetadata`. All these implementations based on the `EventProcessorClient` which stores checkpoint information as blobcontainer metadata.
```
tomkerkhove, line 42:
```suggestion
C# Applications which use the current `Azure.Messaging.EventHubs` package, Java applications which use the current `azure-messaging-eventhubs` package or Python applications since `azure-eventhub v5`, have to set the `checkpointStrategy`to `blobMetadata`. All these implementations based on the `EventProcessorClient` which stores checkpoint information as blobcontainer metadata.

```

content/docs/2.3/scalers/azure-event-hub.md
```diff
@@ -27,10 +27,19 @@ triggers:
```
tomkerkhove, line 27: Can you remove the notice in the metadata please?

> notice = "As of now, the Event Hub scaler only supports reading checkpoints from Blob Storage, as well as scaling only Event Hub applications written in C#, Java, Python or created with Azure Functions."

content/docs/2.3/scalers/azure-event-hub.md
tomkerkhove, line 42: Let's use a quote to align with other scalers
```suggestion
> 💡 **Legacy Checkpointing:** C# applications, which use the `Microsoft.Azure.EventHubs` package, Java applications which use the `azure-eventhubs-eph` package or Python applications which use the `azure-eventhub` below `v5` are supported if no `checkpointStrategy` is specified. These legacy implementations are based on the `EventProcessorHost` client, which stores checkpoint information as blob container content
> 
> C# Applications that use the current `Azure.Messaging.EventHubs` package, Java applications which use the current `azure-messaging-eventhubs` package or Python applications since `azure-eventhub v5`, have to set the `checkpointStrategy`to `blobMetadata`. All these implementations based on the `EventProcessorClient` which stores checkpoint information as blob metadata.

```

tomkerkhove (approved):
LGTM, just a few tweaks!

content/docs/2.3/scalers/azure-event-hub.md
christle: done

content/docs/1.4/scalers/azure-event-hub.md
```diff
@@ -4,7 +4,6 @@ layout = "scaler"
 availability = "v1.0+"
 maintainer = "Community"
 description = "Scale applications based on Azure Event Hubs."
-notice = "As of now, the Event Hub scaler only supports reading checkpoints from Blob Storage, as well as scaling only Event Hub applications written in C#, Java, Python or created with Azure Functions."
```
tomkerkhove, line 7: Can you remove this in v2.3 instead please?

christle: oh, damn sry. done

tomkerkhove: No problem, thanks!

tomkerkhove:
Thank you!

PR #430, opened 2021-05-01
Update Azure Service Bus documentation for queueLength -> messageCount spec changes

ankrause:
Replace "queueLength" with "messageCount" per a recent commit to the Service Bus scaler logic.

The PR in question is [#1138](https://github.com/kedacore/keda/pull/1138) in KEDA.


ankrause:
I'll close this as I see that it's correct in v2.2, but if you google search for the scaler then the indexed page is 1.4. The "Click here for latest" button will take you to the 2.2 version of the page, but manually selecting a version will take you to the base docs page (rather than respecting relative path) and the dropdown shows (latest) next to whichever version you have selected. Just overall a bit odd and hinges on clicking the yellow button for an optimal experience.

PR #441, opened 2021-05-07
Add HMCTS to keda users

timja:


tomkerkhove (approved):
Thank you for using KEDA!

PR #469, opened 2021-06-18
Provide FAQ topics

tomkerkhove:
Provide more FAQ entries:
- What is the value on the HPA that I'm seeing?
- Update HTTP FAQ to fix dead line and point to our experimental add-on scaler
- Introduce best practices
  - Using multiple triggers
  - Mixing KEDA with HPA is not a good bet

### Checklist

- [x] A PR is opened to update the documentation on [our docs repo](https://github.com/kedacore/keda-docs)

### DIFF
--- a/data/faq20.toml
+++ b/data/faq20.toml
@@ -23,7 +23,7 @@ q = "Can I scale my HTTP container or function with KEDA and Kubernetes?"
 a = """
 KEDA will scale a container using metrics from a scaler, but unfortunately there is no scaler today for HTTP workloads.
 
-We recommend using the [Prometheus scaler](/scalers/prometheus/) to create scale rule based on metrics around HTTP events for now. Read [Anirudh Garg's blog post](https://dev.to/anirudhgarg_99/scale-up-and-down-a-http-triggered-function-app-in-kubernetes-using-keda-4m42) to learn more.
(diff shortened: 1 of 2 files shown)


data/faq20.toml
```diff
@@ -32,6 +35,39 @@ q = "Is short polling intervals a problem?"
...
+[[qna]]
+q = "Using multiple triggers for the same scale target"
+a = """
+KEDA allows you to use multiple triggers as part of the same `ScaledObject` or `ScaledJob`.
+
+By doing this, your autoscaling becomes better:
+- All your autoscaling rules are in one place
+- You will not have multiple `ScaledObject`'s or `ScaledJob`'s interferring with each other
+
+KEDA will start scaling as soon as when one of the triggers meets the criteria.
```
zroubalik, line 47: I'd add a note that the maximum number is used for the target replica count calculation

zroubalik (commented):
Looking good, minor nit on multiple triggers

data/faq20.toml
tomkerkhove: You mean the maximum number of what? The configured value for that trigger or?

zroubalik: I meant max number acrross multiple triggers. Let' say you have mutliple triggers and each reports different metrics eg: trigger1: 5, trigger2: 10, trigger3: 0. Then HPA will  do the calculation on each of these metrics and will select the largest number for desired replica count.

tomkerkhove: What about this?
```suggestion
KEDA will start scaling as soon as when one of the triggers meets the criteria. Horizontal Pod Autoscaler (HPA) will calculate metrics for every scaler and use the highest desired replica count to scale the workload to.
```

zroubalik: Perfect :)

zroubalik (approved):

PR #474, opened 2021-06-28
Add documentation for fallback behaviour

mishamo:
Signed-off-by: misha <mishamo@gmail.com>

### Checklist

- [x] A PR is opened to update the documentation on [our docs repo](https://github.com/kedacore/keda-docs)

Fixes https://github.com/kedacore/keda/issues/1872

### DIFF
--- a/content/docs/2.4/concepts/scaling-deployments.md
+++ b/content/docs/2.4/concepts/scaling-deployments.md
@@ -136,13 +136,14 @@ KEDA will keep track of the number of consecutive times each scaler has failed t
 ```
 target metric value * fallback replicas
 ```
-Due to the HPA metric being of type `AverageValue` (see below), this will have the effect of the HPA scaling the deployment to the expected number of replicas.
-
-**N.B.** Fallback is only compatible with scalers that present their target as an `AverageValue` metric. For other scalers (CPU/Memory etc), KEDA will assume fallback is disabled.
-
-**N.B.** Fallback is only compatible with `ScaledObjects`. It is **not** compatible with `ScaledJobs`.
+Due to the HPA metric being of type `AverageValue` (see below), this will have the effect of the HPA scaling the deployment to the defined number of fallback replicas.
 
 **Example:** When my instance of prometheus is unavailable 3 consecutive times, KEDA will change the HPA metric such that the deployment will scale to 6 replicas.
+
+There are a few limitations to using a fallback:
+ - It is **not** supported by the CPU & memory scalers and will assume that fallback is disabled in those cases. This is because it only supports scalers that present their target as an `AverageValue` metric.
+ - It is only supported by `ScaledObjects` **not** `ScaledJobs`.
+
 
 ---
 
--- a/content/docs/2.4/concepts/scaling-deployments.md
+++ b/content/docs/2.4/concepts/scaling-deployments.md
@@ -48,6 +48,9 @@ spec:
   cooldownPeriod:  300                               # Optional. Default: 300 seconds
   minReplicaCount: 0                                 # Optional. Default: 0
   maxReplicaCount: 100                               # Optional. Default: 100
+  fallback:                                          # Optional. Section to specify fallback options
+    failureThreshold: 3                              # Mandatory if fallback section is included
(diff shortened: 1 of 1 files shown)


content/docs/2.4/concepts/scaling-deployments.md
```diff
@@ -121,6 +124,28 @@ This setting is passed to the HPA definition that KEDA will create for a given r
...
+    replicas: 6                                      # Mandatory if fallback section is included
+```
+
+The `fallback` section is optional. It defines a number of replicas to fallback to if a scaler is in an error state.
+
+KEDA will keep track of the number of consecutive times each scaler has failed to get metrics from its source. Once that value passes the `failureThreshold`, instead of not propagating a metric to the HPA (the default error behaviour), the scaler will, instead, return a normalised metric using the formula:
+```
+target metric value * fallback replicas
+```
+Due to the HPA metric being of type `AverageValue` (see below), this will have the effect of the HPA scaling the deployment to the expected number of replicas.
```
tomkerkhove, line 139: I think this is what you mean, right? Maybe better to make it a bit more explicit because you could also read it as the general replica number.
```suggestion
Due to the HPA metric being of type `AverageValue` (see below), this will have the effect of the HPA scaling the deployment to the defined number of fallback replicas.
```

content/docs/2.4/concepts/scaling-deployments.md
```diff
@@ -121,6 +124,28 @@ This setting is passed to the HPA definition that KEDA will create for a given r
...
+
+**N.B.** Fallback is only compatible with scalers that present their target as an `AverageValue` metric. For other scalers (CPU/Memory etc), KEDA will assume fallback is disabled.
```
tomkerkhove, line 141: We should list all the not supported scalers instead, end-users don't know which ones use `AverageValue` metric since that's managed under the hood.

content/docs/2.4/concepts/scaling-deployments.md
```diff
@@ -121,6 +124,28 @@ This setting is passed to the HPA definition that KEDA will create for a given r
...
+
+**N.B.** Fallback is only compatible with `ScaledObjects`. It is **not** compatible with `ScaledJobs`.
+
+**Example:** When my instance of prometheus is unavailable 3 consecutive times, KEDA will change the HPA metric such that the deployment will scale to 6 replicas.
```
tomkerkhove, line 145:
```suggestion
**Example:** When my instance of prometheus is unavailable 3 consecutive times, KEDA will change the HPA metric such that the deployment will scale to 6 replicas.

There are a few limitations for using a fallback:

- It is not supported for the CPU & memory scaler and will assume fallback is disabled.
- It is only supported for `ScaledObjects`, not with `ScaledJobs`


```

content/docs/2.4/concepts/scaling-deployments.md
tomkerkhove: I have added a suggestion below, let me know what you think.

tomkerkhove (changes requested):
LGTM, just a few improvements

content/docs/2.4/concepts/scaling-deployments.md
mishamo: My worry with this change is twofold:
 - The etc covers custom use cases for the external scalers (which may or may not use `AverageValue`)
 - Explicitly adding unsupported (or indeed supported) scalers here, would be prone to becoming inaccurate over time; I imagine people PRing new scalers would forget to add to this list.

Thoughts?

tomkerkhove: It would also be more explicit to end-users, though...

What do you think @zroubalik ?

zroubalik: +1 for listing the scalers

zroubalik: List becoming inaccurate over time is a valid concern, but I'd still take the risk

zroubalik (approved):

tomkerkhove (approved):

PR #476, opened 2021-06-29
Add Meltwater to Keda users

recollir:
Add Meltwater to Keda users. Thanks for all the work.


tomkerkhove (approved):
Thank you! 💖

tomkerkhove:
@recollir We are looking for KEDA end-users who want to be interviewed by CNCF for our CNCF Graduation proposal.

PR #489, opened 2021-07-15
Update RabbitMQ Scaler documentation to include the option of use regex

JorTurFer:
This PR is related with https://github.com/kedacore/keda/pull/1957

### Checklist

Relates to https://github.com/kedacore/keda/issues/743
Relates to https://github.com/kedacore/keda/pull/1957

### DIFF
--- a/content/docs/2.4/scalers/rabbitmq-queue.md
+++ b/content/docs/2.4/scalers/rabbitmq-queue.md
@@ -37,7 +37,7 @@ triggers:
 - `protocol`: Protocol to be used for communication. Either `auto`, `http`, or `amqp`. It should correspond with the `host` value. Optional, will autodetect based on the `host` URL if possible.
 - `vhostName`: Vhost to use for the connection, overrides any vhost set in the connection string from `host`/`hostFromEnv`.
 - `queueLength`: DEPRECATED! Use `mode: QueueLength` and `value: ##` instead. Target value for queue length passed to the scaler. Example: if one pod can handle 10 messages, set the queue length target to 10. If the actual number of messages in the queue is 30, the scaler scales to 3 pods. Default is 20 unless `publishRate` is specified, in which case `queueLength` is disabled for this trigger.
-- `useRegex`: In case of `http` protocol and `QueueLength` mode, this parameter allows to use regex (in `queueName` parameter) to select queue instead of full name. Optional.
+- `useRegex`: In case of `http` protocol and `QueueLength` mode, this parameter allows to use regex (in `queueName` parameter) to select queue instead of full name, the valid values are: `"true"` and `"false"`.  Optional.
 - `operation`: Operation that will be applied to compute the number of messages in case of `useRegex` enabled. Either `sum` (default),`max`, or `avg`. Optional.
 
 Some parameters could be provided using environmental variables, instead of setting them directly in metadata. Here is a list of parameters you can use to retrieve values from environment variables:
--- a/content/docs/2.4/scalers/rabbitmq-queue.md
+++ b/content/docs/2.4/scalers/rabbitmq-queue.md
@@ -23,7 +23,7 @@ triggers:
     vhostName: / # Optional. If not specified, use the vhost in the `host` connection string.
     # Alternatively, you can use existing environment variables to read configuration from:
     # See details in "Parameter list" section
-    hostFromEnv: RABBITMQ_HOST # Optional. You can use this instead of `host` parameter
+    hostFromEnv: RABBITMQ_HOST # Optional. You can use this instead of `host` parameter    
 ```
 
 **Parameter list:**
@@ -37,6 +37,8 @@ triggers:
 - `protocol`: Protocol to be used for communication. Either `auto`, `http`, or `amqp`. It should correspond with the `host` value. Optional, will autodetect based on the `host` URL if possible.
 - `vhostName`: Vhost to use for the connection, overrides any vhost set in the connection string from `host`/`hostFromEnv`.
 - `queueLength`: DEPRECATED! Use `mode: QueueLength` and `value: ##` instead. Target value for queue length passed to the scaler. Example: if one pod can handle 10 messages, set the queue length target to 10. If the actual number of messages in the queue is 30, the scaler scales to 3 pods. Default is 20 unless `publishRate` is specified, in which case `queueLength` is disabled for this trigger.
+- `useRegex`: In case of `http` protocol and `QueueLength` mode, this parameter allows to use regex to select queue instead full name. Optional.
+- `operation`: Operation that will be applied to compute the number of messages in case of `useRegex` enabled. Either `sum`,`max`, or `avg`. The default value is `sum`.
 
 Some parameters could be provided using environmental variables, instead of setting them directly in metadata. Here is a list of parameters you can use to retrieve values from environment variables:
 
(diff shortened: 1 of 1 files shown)


content/docs/2.4/scalers/rabbitmq-queue.md
```diff
@@ -37,6 +37,8 @@ triggers:
 - `protocol`: Protocol to be used for communication. Either `auto`, `http`, or `amqp`. It should correspond with the `host` value. Optional, will autodetect based on the `host` URL if possible.
 - `vhostName`: Vhost to use for the connection, overrides any vhost set in the connection string from `host`/`hostFromEnv`.
 - `queueLength`: DEPRECATED! Use `mode: QueueLength` and `value: ##` instead. Target value for queue length passed to the scaler. Example: if one pod can handle 10 messages, set the queue length target to 10. If the actual number of messages in the queue is 30, the scaler scales to 3 pods. Default is 20 unless `publishRate` is specified, in which case `queueLength` is disabled for this trigger.
+- `useRegex`: In case of `http` protocol and `QueueLength` mode, this parameter allows to use regex to select queue instead full name. Optional.
+- `operation`: Operation that will be applied to compute the number of messages in case of `useRegex` enabled. Either `sum`,`max`, or `avg`. The default value is `sum`.
```
tomkerkhove, line 41:
```suggestion
- `operation`: Operation that will be applied to compute the number of messages in case of `useRegex` enabled. Either `sum` (default),`max`, or `avg`.
```

tomkerkhove (approved):

content/docs/2.4/scalers/rabbitmq-queue.md
```diff
@@ -37,6 +37,8 @@ triggers:
 - `protocol`: Protocol to be used for communication. Either `auto`, `http`, or `amqp`. It should correspond with the `host` value. Optional, will autodetect based on the `host` URL if possible.
 - `vhostName`: Vhost to use for the connection, overrides any vhost set in the connection string from `host`/`hostFromEnv`.
 - `queueLength`: DEPRECATED! Use `mode: QueueLength` and `value: ##` instead. Target value for queue length passed to the scaler. Example: if one pod can handle 10 messages, set the queue length target to 10. If the actual number of messages in the queue is 30, the scaler scales to 3 pods. Default is 20 unless `publishRate` is specified, in which case `queueLength` is disabled for this trigger.
+- `useRegex`: In case of `http` protocol and `QueueLength` mode, this parameter allows to use regex to select queue instead full name. Optional.
+- `operation`: Operation that will be applied to compute the number of messages in case of `useRegex` enabled. Either `sum` (default),`max`, or `avg`.
```
zroubalik, line 41: Minor nit, `operation` is `Optional` as well, might be worth mentioning.

content/docs/2.4/scalers/rabbitmq-queue.md
```diff
@@ -188,3 +192,45 @@ spec:
...
+  name: rabbitmq-scaledobject
+  namespace: default
+spec:
+  scaleTargetRef:
+    name: rabbitmq-deployment
+  triggers:
+  - type: rabbitmq
+    metadata:
+      protocol: http
+      queueName: testqueue
```
zroubalik, line 229: Might be better to use some more complex regex in here as the example queueName, so it is more clear, where users should write the regex

content/docs/2.4/scalers/rabbitmq-queue.md
```diff
@@ -37,6 +37,8 @@ triggers:
 - `protocol`: Protocol to be used for communication. Either `auto`, `http`, or `amqp`. It should correspond with the `host` value. Optional, will autodetect based on the `host` URL if possible.
 - `vhostName`: Vhost to use for the connection, overrides any vhost set in the connection string from `host`/`hostFromEnv`.
 - `queueLength`: DEPRECATED! Use `mode: QueueLength` and `value: ##` instead. Target value for queue length passed to the scaler. Example: if one pod can handle 10 messages, set the queue length target to 10. If the actual number of messages in the queue is 30, the scaler scales to 3 pods. Default is 20 unless `publishRate` is specified, in which case `queueLength` is disabled for this trigger.
+- `useRegex`: In case of `http` protocol and `QueueLength` mode, this parameter allows to use regex to select queue instead full name. Optional.
```
zroubalik, line 40:
```suggestion
- `useRegex`: In case of `http` protocol and `QueueLength` mode, this parameter allows to use regex (in `queueName` field) to select queue instead full name. Optional.
```

content/docs/2.4/scalers/rabbitmq-queue.md
JorTurFer: you are right! done :)

content/docs/2.4/scalers/rabbitmq-queue.md
JorTurFer: Thanks for the suggestion! Done :)

content/docs/2.4/scalers/rabbitmq-queue.md
JorTurFer: You are right, I have set `^.*incoming$` to clarify that it's a regex

JorTurFer:
I have just applied your suggestions @zroubalik.
Thanks for them! :)

content/docs/2.4/scalers/rabbitmq-queue.md
```diff
@@ -37,6 +37,8 @@ triggers:
 - `protocol`: Protocol to be used for communication. Either `auto`, `http`, or `amqp`. It should correspond with the `host` value. Optional, will autodetect based on the `host` URL if possible.
 - `vhostName`: Vhost to use for the connection, overrides any vhost set in the connection string from `host`/`hostFromEnv`.
 - `queueLength`: DEPRECATED! Use `mode: QueueLength` and `value: ##` instead. Target value for queue length passed to the scaler. Example: if one pod can handle 10 messages, set the queue length target to 10. If the actual number of messages in the queue is 30, the scaler scales to 3 pods. Default is 20 unless `publishRate` is specified, in which case `queueLength` is disabled for this trigger.
+- `useRegex`: In case of `http` protocol and `QueueLength` mode, this parameter allows to use regex (in `queueName` parameter) to select queue instead of full name. Optional.
```
zroubalik, line 40: One last minor thing, we might want to mark the allowed values for `useRegex` parameter, as it might not be clear immediately.

zroubalik (approved):

content/docs/2.4/scalers/rabbitmq-queue.md
zroubalik: I suppose those values are true/false. Should the value be in quotes? Following note suggest that they might be mandatory? https://github.com/kedacore/keda-docs/pull/489/files#diff-4e706b2e67d6b7ef7fb2f4746b593e86326142120d790c6420bb2f785f269dffR51

zroubalik: @JorTurFer ^^

JorTurFer: Maybe I'm wrong, but I understand that all the values in metadata have to be strings to be automatically parsed and accessible  inside the scaler, right? 
You are right, the values are true/false but in order to be automatically parsed, it should be in quotes. It's the same case as `value`, It is a number, but it is converted to string in yaml

zroubalik: Yeah, that's true, my comment was more like a hypothetical question from an user's point of view.

JorTurFer: I can add something like this:
- `useRegex`: In case of `http` protocol and `QueueLength` mode, this parameter allows to use regex (in `queueName` parameter) to select queue instead of full name, the valid values are: `"true"` and `"false"`.  Optional.

WDTY?

zroubalik: Yeah, that's looking good!

JorTurFer: Done ^^

zroubalik (approved):

zroubalik (approved):

PR #496, opened 2021-08-01
pendingPodConditions documentation

yaronya:
Signed-off-by: Yaron Yarimi <yaron.yarimi@env0.com>

Adding the ability to calculate the pending job count per pod conditions rather then it's phase. See https://github.com/kedacore/keda/pull/1970

### Checklist

### DIFF
--- a/content/docs/2.4/concepts/scaling-jobs.md
+++ b/content/docs/2.4/concepts/scaling-jobs.md
@@ -44,6 +44,10 @@ spec:
     strategy: "custom"                        # Optional. Default: default. Which Scaling Strategy to use. 
     customScalingQueueLengthDeduction: 1      # Optional. A parameter to optimize custom ScalingStrategy.
     customScalingRunningJobPercentage: "0.5"  # Optional. A parameter to optimize custom ScalingStrategy.
+    pendingPodConditions:                     # Optional. A parameter to calculate pending job count per the specified pod conditions
+      - "Ready"
+      - "PodScheduled"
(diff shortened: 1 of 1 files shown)


content/docs/2.4/concepts/scaling-jobs.md
```diff
@@ -136,7 +140,9 @@ Select a Scaling Strategy. Possible values are `default`, `custom`, or `accurate
 >		runningJobs++
 >}
 >```
->`PendingJobCount` provides an indication of the amount of jobs that are in pending state, for example a that has not finished yet **and** the underlying pod is either not running or has not completed yet.
+>`PendingJobCount` provides an indication of the amount of jobs that are in pending state. Pending jobs can be calculated in two ways:
+> - Default behavior - Job that has not finished yet **and** the underlying pod is either not running or has not completed yet (Default Behavior)
```
tomkerkhove, line 144:
```suggestion
> - Default behavior - Job that have not finished yet **and** the underlying pod is either not running or has not been completed yet
```

tomkerkhove (approved):
Thank you for your contribution - Very clear and straightforward!

content/docs/2.4/concepts/scaling-jobs.md
yaronya: 👍 
I've added your suggestion + fixed the code block (that has changed in my linked PR): d76801d

zroubalik (approved):

PR #501, opened 2021-08-05
Added documentation for new kubernetes workload scaler

JorTurFer:

Added documentation for new kubernetes workload scaler

### Checklist

Relates to https://github.com/kedacore/keda/issues/1735
Relates to https://github.com/kedacore/keda/pull/2010

### DIFF
--- a/content/docs/2.4/scalers/kubernetes-workload.md
+++ b/content/docs/2.4/scalers/kubernetes-workload.md
@@ -24,6 +24,10 @@ triggers:
 - `namespace` - Namespace where the `podSelector` must be applied. If it's empty, all namespaces are eligibles. This parameter is optional and default value is empty.
 - `value` - Target relation between the scaled workload and the amount of pods which matches the selector. It will be calculated following this formula: `relation = (pods which match selector) / (scaled workload pods)`
 
+### Authentication Parameters
+
+The own KEDA's identity is used to list the pods, so no configuration is needed here.
+
 ### Example
 
 ```yaml
--- a/content/docs/2.4/scalers/kubernetes-workload.md
+++ b/content/docs/2.4/scalers/kubernetes-workload.md
@@ -22,6 +22,8 @@ triggers:
 - `podSelector` - [Label selector](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#label-selectors) that will be used to get the pod count. It supports multiple selectors split by coma (`,`). It also supports [set-based requirements](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#set-based-requirement) and a mix of them.
 - `value` - Target relation between the scaled workload and the amount of pods which matches the selector. It will be calculated following this formula: `relation = (pods which match selector) / (scaled workload pods)`
 
+> 💡 **Note:** The search scope is limited to the namespace where the `ScaledObject` is deployed.
+
 ### Authentication Parameters
 
 The own KEDA's identity is used to list the pods, so no extra configuration is needed here.
--- a/content/docs/2.4/scalers/kubernetes-workload.md
+++ b/content/docs/2.4/scalers/kubernetes-workload.md
@@ -14,14 +14,12 @@ triggers:
 - type: kubernetes-workload
   metadata:
     podSelector: 'app=backend'
-    namespace: 'default'
     value: '1'
 ```
 
 **Parameter list:**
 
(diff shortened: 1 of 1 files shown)


content/docs/2.4/scalers/kubernetes-workload.md
```diff
@@ -0,0 +1,43 @@
...
+    podSelector: 'app=backend'
+    namespace: 'default'
+    value: '1'
+```
+
+**Parameter list:**
+
+- `podSelector` - [Label selector](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#label-selectors) that will be used to get the pod count. It supports multiple selectors split by coma (`,`). It also supports [set-based requirements](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#set-based-requirement) and a mix of them.
+- `namespace` - Namespace where the `podSelector` must be applied. If it's empty, all namespaces are eligibles. This parameter is optional and default value is empty.
+- `value` - Target relation between the scaled workload and the amount of pods which matches the selector. It will be calculated following this formula: `relation = (pods witch match selector) / (scaled workload pods)`
```
zroubalik, line 25: `witch` -> `which` 😄

JorTurFer: done! 
Nice vision! :)
TY

content/docs/2.4/scalers/kubernetes-workload.md
```diff
@@ -0,0 +1,43 @@
...
+  metadata:
+    podSelector: 'app=backend'
+    namespace: 'default'
+    value: '1'
+```
+
+**Parameter list:**
+
+- `podSelector` - [Label selector](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#label-selectors) that will be used to get the pod count. It supports multiple selectors split by coma (`,`). It also supports [set-based requirements](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#set-based-requirement) and a mix of them.
+- `namespace` - Namespace where the `podSelector` must be applied. If it's empty, all namespaces are eligibles. This parameter is optional and default value is empty.
```
tomkerkhove, line 24: Shouldn't we be "secure by default" and only scan the same namespace when not specified? Just thinking out loud.

If we do, then we need to think how they would match cross namespace, but why would you do that since you typically target a given workload that you own, no?

content/docs/2.4/scalers/kubernetes-workload.md
```diff
@@ -0,0 +1,43 @@
...
+maintainer = "Community"
+description = "Scale applications based on the amount of pods which matches the given selectors."
+go_file = "kubernetes_workload_scaler"
++++
+
+### Trigger Specification
+
+```yaml
+triggers:
+- type: kubernetes-workload
```
tomkerkhove, line 14: What is the authentication story here? It uses the KEDA identity I presume? Maybe good to document that as well?

tomkerkhove (commented):
Added some remarks, what do you think @zroubalik ?

content/docs/2.4/scalers/kubernetes-workload.md
JorTurFer: You're right, I will write it explicitly

content/docs/2.4/scalers/kubernetes-workload.md
JorTurFer: I'm not super sure about it, maybe you are right and we should scan only the own namespace as default but idk
From the security pov, maybe it could be better to use a specific service account which should be provided in order to control the permissions with granularity, but it could be more complex than now

JorTurFer: if you agree with it, I believe that I can make the change today :)

content/docs/2.4/scalers/kubernetes-workload.md
JorTurFer: done!

content/docs/2.4/scalers/kubernetes-workload.md
zroubalik: That's a good point Tom. I think that we should limit this only to the current namespace (the one that the ScaledObject is being deployed to).

JorTurFer: In all cases or as default?

zroubalik: All cases. We can enable this in the future, if there's demand.

JorTurFer: okey. I will update it :)

JorTurFer:
As we discussed @tomkerkhove @zroubalik, I have just removed the possibility to select the namespace. The search scope is the namespace where the scaledobject is placed

tomkerkhove (approved):

zroubalik (approved):

PR #512, opened 2021-08-16
docs: Introduce recursive listing for Azure Blob scaler

ahmelsayed:

### Checklist

Fixes #

### DIFF
--- a/content/docs/2.5/scalers/azure-storage-blob.md
+++ b/content/docs/2.5/scalers/azure-storage-blob.md
@@ -31,7 +31,7 @@ triggers:
 - `blobCount` - Average target value to trigger scaling actions. (default: 5)
 - `connectionFromEnv` - Name of the environment variable your deployment uses to get the connection string.
 - `blobPrefix` - Prefix for the Blob. Use this to specify sub path for the blobs if required. (default: `""`)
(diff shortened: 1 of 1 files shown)


content/docs/2.5/scalers/azure-storage-blob.md
```diff
@@ -31,7 +31,7 @@ triggers:
 - `blobCount` - Average target value to trigger scaling actions. (default: 5)
 - `connectionFromEnv` - Name of the environment variable your deployment uses to get the connection string.
 - `blobPrefix` - Prefix for the Blob. Use this to specify sub path for the blobs if required. (default: `""`)
-- `blobDelimiter` - Delimiter for identifying the blob prefix. (default: `/`)
+- `blobDelimiter` - Delimiter for identifying the blob prefix. Set to `""` to list all blobs with `blobPrefix` recursively (default: `/`)
```
tomkerkhove, line 34: Wouldn't it be better to have an option such as `listBlobsRecursively: true` (default: false)?

Because maybe I want to list recursively, but only as of path `foo/bar/*`

JorTurFer:
hi @ahmelsayed ,
Please, remember to move this to version 2.6 🙏 
Thanks!

v-shenoy:
@kedacore/keda-contributors This should be closed now after #738.

PR #523, opened 2021-09-01
Add doc about custom http timeout in RabbitMQ Scaler

JorTurFer:

This PR adds the docs related with "Add support to specify custom http timeout only for RabbitMQ Scaler"

### Checklist

Fixes https://github.com/kedacore/keda/issues/2053

### DIFF
--- a/content/docs/2.5/scalers/rabbitmq-queue.md
+++ b/content/docs/2.5/scalers/rabbitmq-queue.md
@@ -39,7 +39,8 @@ triggers:
 - `queueLength`: DEPRECATED! Use `mode: QueueLength` and `value: ##` instead. Target value for queue length passed to the scaler. Example: if one pod can handle 10 messages, set the queue length target to 10. If the actual number of messages in the queue is 30, the scaler scales to 3 pods. Default is 20 unless `publishRate` is specified, in which case `queueLength` is disabled for this trigger.
 - `useRegex`: In case of `http` protocol, this parameter allows to use regex (in `queueName` parameter) to select queue instead of full name, the valid values are: `"true"` and `"false"`.  Optional.
 - `operation`: Operation that will be applied to compute the number of messages in case of `useRegex` enabled. Either `sum` (default),`max`, or `avg`. Optional.
-- `metricName` - an optional name to assign to the metric. If not set KEDA will generate a name based on the queue name. If using more than one trigger it is required that all metricNames be unique.
+- `metricName`: an optional name to assign to the metric. If not set KEDA will generate a name based on the queue name. If using more than one trigger it is required that all metricNames be unique.
+- `timeout`: Timeout **for the specific trigger** in case of `http` protocol. This value will override the value defined in `KEDA_HTTP_DEFAULT_TIMEOUT`. Optional.
 
 Some parameters could be provided using environmental variables, instead of setting them directly in metadata. Here is a list of parameters you can use to retrieve values from environment variables:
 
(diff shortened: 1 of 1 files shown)


content/docs/2.5/scalers/rabbitmq-queue.md
```diff
@@ -39,7 +39,8 @@ triggers:
...
 - `useRegex`: In case of `http` protocol, this parameter allows to use regex (in `queueName` parameter) to select queue instead of full name, the valid values are: `"true"` and `"false"`.  Optional.
 - `operation`: Operation that will be applied to compute the number of messages in case of `useRegex` enabled. Either `sum` (default),`max`, or `avg`. Optional.
-- `metricName` - an optional name to assign to the metric. If not set KEDA will generate a name based on the queue name. If using more than one trigger it is required that all metricNames be unique.
+- `metricName`: an optional name to assign to the metric. If not set KEDA will generate a name based on the queue name. If using more than one trigger it is required that all metricNames be unique.
+- `timeout`: Timeout **for the specific trigger** in case of `http` protocol. This value will override the value defined in `KEDA_HTTP_DEFAULT_TIMEOUT`. Optional.
```
tomkerkhove, line 43: What do you mean with "for the specific trigger"?

JorTurFer: That this timeout only for this trigger, and it will not affect to other triggers. You can have 2 different triggers that uses RabbitMQ with totally different timeouts. (Maybe is my translation a "false friend"?)

zroubalik: Maybe use `for this specific trigger`?

tomkerkhove: That, or change the docs for the "global" setting to emphasize that it's global because the trigger docs typically focus on the trigger itself.

If you know what I mean 😅

JorTurFer: I got it, I can do both :)
I have already updated to change from for the specific trigger to for this specific trigger, but I can change also this "This value will override the value defined in KEDA_HTTP_DEFAULT_TIMEOUT" with "This value will override the global value defined in KEDA_HTTP_DEFAULT_TIMEOUT"
Is it better in your opinion?

tomkerkhove: Where is KEDA_HTTP_DEFAULT_TIMEOUT documented other than the Helm chart? I'd make it explicit there that this is global

JorTurFer: It's already explained [in docs](https://keda.sh/docs/2.1/operate/cluster/#http-timeouts)
> Some scalers issue HTTP requests to external servers (i.e. cloud services). Each applicable scaler uses its own dedicated HTTP client with its own connection pool, and by default each client is set to time out any HTTP request after 3 seconds.
 ...

tomkerkhove: Good enough, thanks!

JorTurFer: okey, then I leave it as it is now :)

content/docs/2.5/scalers/rabbitmq-queue.md
```diff
@@ -39,7 +39,8 @@ triggers:
...
 - `useRegex`: In case of `http` protocol, this parameter allows to use regex (in `queueName` parameter) to select queue instead of full name, the valid values are: `"true"` and `"false"`.  Optional.
 - `operation`: Operation that will be applied to compute the number of messages in case of `useRegex` enabled. Either `sum` (default),`max`, or `avg`. Optional.
-- `metricName` - an optional name to assign to the metric. If not set KEDA will generate a name based on the queue name. If using more than one trigger it is required that all metricNames be unique.
+- `metricName`: an optional name to assign to the metric. If not set KEDA will generate a name based on the queue name. If using more than one trigger it is required that all metricNames be unique.
+- `timeout`: Timeout **for this specific trigger** in case of `http` protocol. This value will override the value defined in `KEDA_HTTP_DEFAULT_TIMEOUT`. Optional.
```
arschles, line 43: @JorTurFer does the first sentence mean to say that the timeout only applies for this specific trigger, only if it's the `http` protocol? If so, here's a suggestion:

```suggestion
- `timeout`: Timeout **for this specific trigger**. This value will override the value defined in `KEDA_HTTP_DEFAULT_TIMEOUT`. Optional. Only apples to hosts that use the `http` protocol
```

JorTurFer: Nice suggestion! Thanks for do it :)

zroubalik (approved):

content/docs/2.5/scalers/rabbitmq-queue.md
arschles: No problem :)

PR #538, opened 2021-09-22
Reduce size of Ampify logo

ofekrom:

### Checklist

Fixes #


tomkerkhove (approved):

tomkerkhove:
@ofekrom - I do get why you wanted to resize the logo, I was wondering if there is one with a different ratio that would fit in better?


Or are we allowed to just use this?
image

PR #547, opened 2021-10-08
Added Metamaze as KEDA user

AntPeixe:
Signed-off-by: Antonio Peixe <a.peixe@metamaze.eu>


tomkerkhove:
@AntPeixe We are looking for KEDA end-users who want to be interviewed by CNCF for our CNCF Graduation proposal.

tomkerkhove:
Thank you!

PR #553, opened 2021-10-14
chore: Change `feature` as allowed label for stale issues, instead of `feature-request`

tomkerkhove:
Change `feature` as allowed label for stale issues, instead of `feature-request` as our intention was to mark issues as features if we want to keep them around.

### Checklist

Relates to https://github.com/kedacore/governance/issues/33


zroubalik (approved):
Makes sense!

PR #555, opened 2021-10-14
Add KENDAXA as KEDA user

hmoravec:
Signed-off-by: Jan Moravec <j.moravec@kendaxa.com>


tomkerkhove (approved):

tomkerkhove:
@hmoravec We are looking for KEDA end-users who want to be interviewed by CNCF for our CNCF Graduation proposal.

PR #605, opened 2021-12-05
kubernetes-workload scaler: ignore terminated pods

solsson:
Documents the upcoming PR in keda that suggests terminated pods should be excluded from the kubernetes-workload metric.

### Checklist


JorTurFer (approved):
LGTM!

PR #609, opened 2021-12-09
Use `latest`  instead specific version in canonical urls

JorTurFer:
Signed-off-by: Jorge Turrado <jorge.turrado@docplanner.com>

Now all pages declare their self as canonical in `head` section. This makes that google doesn't list correctly the really canonical page. With this PR we replace the canonical to enforce `latest` (only) under `docs` sub path

image
image

### Checklist


tomkerkhove (approved):

tomkerkhove:
Thanks!

PR #618, opened 2022-01-05
Update info on AAD Pod Identity Parameter

chrisreddington:
Update info on AAD Pod Identity Parameter

I have been working on a deployment of Dapr/Keda to expand my knowledge in this area. I stumbled upon this doc, and planned to use it to enable Pod Identity for KEDA. The aadPodIdentity setting did not work for me. After reviewing the GitHub repo at kedacore/charts, I found that the correct parameter is actually podIdentity.activeDirectory.identity. Updating the docs to reflect that, and to also point to the helm chart for future reference.

### Checklist

### DIFF
--- a/content/docs/1.4/concepts/authentication.md
+++ b/content/docs/1.4/concepts/authentication.md
@@ -165,7 +165,7 @@ podIdentity:
   provider: azure # Optional. Default: false
 ```
 
-Azure AD Pod Identity will give access to containers with a defined label for `aadpodidbinding`.  You can set this label on the KEDA operator deployment.  This can be done for you during deployment with Helm with `--set aadPodIdentity={your-label-name}`.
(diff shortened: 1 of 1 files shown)


tomkerkhove:
Thanks for updating! However, this is for v1.4. Is that the version you had issues with? Because at that point it was still called like the original configuration.

chrisreddington:
> Thanks for updating! However, this is for v1.4. Is that the version you had issues with? Because at that point it was still called like the original configuration.

Doh, you're exactly right. I'm using 2.5.1, and can see that it's updated in the [latest version of the docs](https://keda.sh/docs/2.5/concepts/authentication/#azure-pod-identity). 

I navigated to that page through the search, not quite realising that I had clicked on a prior version. If there is an option to display the version of a page in search results, that'd be super (to help eliminate confusion like this). Thanks for your quick help @tomkerkhove - appreciate it.

tomkerkhove:
No problem and thanks for opening a PR anyway! Feel free to open an issue for the search  thingy, good suggestion!

PR #626, opened 2022-01-12
doc: Extend info about hpa metric value in faq

JorTurFer:

This PR extends the information about possible differences between the value that the user could see in the metric source and the target value in the HPA.

The change looks like this:
image

### Checklist

Related https://github.com/kedacore/keda/discussions/2451

### DIFF
--- a/data/faq20.toml
+++ b/data/faq20.toml
@@ -66,7 +66,7 @@ The target metric value is used by the Horizontal Pod Autoscaler (HPA) to make s
 
 The current target value on the Horizontal Pod Autoscaler (HPA) often does not match with the metrics on the system you are scaling on. This is because of how the Horizontal Pod Autoscaler's (HPA) [scaling algorithm](https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/#algorithm-details) works.
 
-If you are checking your metric directly in the HPA, take in consideration that KEDA currently only supports `average` metrics, so the value that you will see in the HPA is the average of the metric between the total amount of pods.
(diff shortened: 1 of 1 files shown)


data/faq20.toml
```diff
@@ -65,6 +65,8 @@ a = """
 The target metric value is used by the Horizontal Pod Autoscaler (HPA) to make scaling decisions.
 
 The current target value on the Horizontal Pod Autoscaler (HPA) often does not match with the metrics on the system you are scaling on. This is because of how the Horizontal Pod Autoscaler's (HPA) [scaling algorithm](https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/#algorithm-details) works.
+
+If you are checking your metric directly in the HPA, take in consideration that KEDA currently only supports average metrics, so the value that you will see in the HPA is the average of the metric between the total amount of pods.
```
tomkerkhove, line 69: What about this?
```suggestion
KEDA currently only supports average metrics. This means that the HPA will use the average 
value of the metric between the total amount of pods.
```

tomkerkhove: So this is saying that it will use `average(metric) / # of pods`; is that correct?

JorTurFer: yes

JorTurFer: if you mean `targetValue = metric/ # of pods` yes

JorTurFer: Basically if the upstream metrics is 10 and you have 5 instances, the value that you will see in the HPA is 2 (10/5)

JorTurFer: is there any reason for the new line between average and value or it's just a typo?

tomkerkhove (approved):

PR #627, opened 2022-01-12
docs: Add information about how to manual query if there are more than 1 SO

JorTurFer:

### Checklist

Fixes https://github.com/kedacore/keda/issues/2456

### DIFF
--- a/content/docs/2.4/operate/metrics-server.md
+++ b/content/docs/2.4/operate/metrics-server.md
@@ -76,21 +76,21 @@ You can recover the metric names from a ScaledObject using `kubectl`:
  kubectl get scaledobject SCALEDOBJECT_NAME -n NAMESPACE -o jsonpath={.status.externalMetricNames}
 ```
 
-## How to get metric if there are more than 1 ScaledObject with same metric name
+## How to get metric when multiple ScaledObjects have the same metric name
 
-KEDA will try to select the proper `ScaledObject` for your metric and it should be only one. In case of having more than 1 `ScaledObject` in the same namespace with the same metric name, an error like this will be thrown:
+KEDA will try to select the proper `ScaledObject` for your metric and there should only be one. In case of having multiple `ScaledObject`s in the same namespace with the same metric name, an error like this will be thrown:
 
 ```
 Error from server: exactly one ScaledObject should match label
 ```
 
-In this case, you should add in the query string the labelSelector to match the proper `ScaledObject` (in url format). The needed selector is `scaledobject.keda.sh/name: {ScaledObjectName}`. You also could get it doing:
+In this case, you should add in the query string the `labelSelector` to match the proper `ScaledObject` (in url format). The needed selector is `scaledobject.keda.sh/name: {ScaledObjectName}`. You can achieve this as following:
 
 ```bash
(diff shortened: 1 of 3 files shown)


content/docs/2.4/operate/metrics-server.md
```diff
@@ -74,4 +74,24 @@ During its work, KEDA updates each ScaledObject with some relevant information w
 You can recover the metric names from a ScaledObject using `kubectl`:
 ```bash
  kubectl get scaledobject SCALEDOBJECT_NAME -n NAMESPACE -o jsonpath={.status.externalMetricNames}
+```
+
+## How to get metric if there are more than 1 ScaledObject with same metric name
```
tomkerkhove, line 79:
```suggestion
## How to get metric when multiple ScaledObjects have the same metric name
```

content/docs/2.4/operate/metrics-server.md
```diff
@@ -74,4 +74,24 @@ During its work, KEDA updates each ScaledObject with some relevant information w
...
+
+KEDA will try to select the proper `ScaledObject` for your metric and it should be only one. In case of having more than 1 `ScaledObject` in the same namespace with the same metric name, an error like this will be thrown:
```
tomkerkhove, line 81:
```suggestion
KEDA will try to select the proper `ScaledObject` for your metric and there should only be one. In case of having multiple `ScaledObject`s in the same namespace with the same metric name, an error like this will be thrown:
```

content/docs/2.4/operate/metrics-server.md
```diff
@@ -74,4 +74,24 @@ During its work, KEDA updates each ScaledObject with some relevant information w
...
+
+```
+Error from server: exactly one ScaledObject should match label
+```
+
+In this case, you should add in the query string the labelSelector to match the proper `ScaledObject` (in url format). The needed selector is `scaledobject.keda.sh/name: {ScaledObjectName}`. You also could get it doing:
```
tomkerkhove, line 87:
```suggestion
In this case, you should add in the query string the `labelSelector` to match the proper `ScaledObject` (in url format). The needed selector is `scaledobject.keda.sh/name: {ScaledObjectName}`. You can achieve this as following:
```

content/docs/2.4/operate/metrics-server.md
```diff
@@ -74,4 +74,24 @@ During its work, KEDA updates each ScaledObject with some relevant information w
...
+
+```bash
+ kubectl get scaledobject SCALEDOBJECT_NAME -n NAMESPACE -o jsonpath={.metadata.labels}
+```
+
+Having the selector, you only need to add it in the query string:
```
tomkerkhove, line 93:
```suggestion
Once you have the selector, you have to add it to your query string as following:
```

tomkerkhove (approved):
LGTM, but added some suggestions for all versions

content/docs/2.4/operate/metrics-server.md
JorTurFer: > LGTM, but added some suggestions for all versions

Nice, I will update all of them, thanks for the text improvements :)

PR #628, opened 2022-01-17
Allow simple SQS queue name

nicolaferraro:
Signed-off-by: nicolaferraro <ni.ferraro@gmail.com>

### Checklist

Relates to https://github.com/kedacore/keda/pull/2483

### DIFF
--- a/content/docs/2.6/scalers/aws-sqs.md
+++ b/content/docs/2.6/scalers/aws-sqs.md
@@ -25,7 +25,7 @@ triggers:
 
 **Parameter list:**
 
-- `queueURL` - Full URL for the SQS Queue.
+- `queueURL` - Full URL for the SQS Queue. The simple name of the queue can be used in case there's no ambiguity.
 - `queueLength` - Target value for queue length passed to the scaler. Example: if one pod can handle 10 messages, set the queue length target to 10. If the actual messages in the SQS Queue is 30, the scaler scales to 3 pods. (default: 5)
 
 > For the purposes of scaling, "actual messages" is equal to `ApproximateNumberOfMessages` + `ApproximateNumberOfMessagesNotVislble`, since `NotVisible` in SQS terms means the message is still in-flight/processing.


content/docs/2.6/scalers/aws-sqs.md
```diff
@@ -25,7 +25,7 @@ triggers:
 
 **Parameter list:**
 
-- `queueURL` - Full URL for the SQS Queue.
+- `queueURL` - Full URL for the SQS Queue. The simple name of the queue can be used in case there's no ambiguity.
```
tomkerkhove, line 28: Hm, shouldn't we go with this instead?
```suggestion
- `queueURL` - Full URL for the SQS Queue. (Optional, when `queueName` is specified)
- `queueName` - Name of the queue.  (Optional, when `queueURL` is specified)
```

nicolaferraro: I'm not sure. At API level you can use a simple name in the queueURL field, so I don't know if a new field is needed.

tomkerkhove: From a KEDA user-perspective they don't know that and providing `orders-queue` in `queueURL` would be confusing, no?

JorTurFer: There is not any extra parameter here @tomkerkhove. The behavior of current `queueURL` parameter has changed because it's not needed "the full name". Am I right @nicolaferraro ?

JorTurFer: Maybe we could transform `queueURL` into `queueName` and explain that it could be full qualified name or simple name. WDYT?

tomkerkhove: That would be a breaking change though. I'd just provide seperate fields and internally you can just use the same.

From a user-perspective, there's no added value to try to cram them both in one field - No?

zroubalik: I think it is OK, to keep it as it is. There's no need to actually create a new property imho, though the naming is not ideal, I know :/ But it is better than actually introducing a breaking change.

zroubalik (approved):

content/docs/2.6/scalers/aws-sqs.md
tomkerkhove: It wouldn't really be breaking if we leave `queueURL` what it is and add `queueName` next to it; but let's go with the current approach as I'm the only one preferring this.

PR #637, opened 2022-01-20
metadata.name must equal spec.scaleTargetRef.deploymentName

megakid:

### Checklist

Fixes #

### DIFF
--- a/content/docs/1.4/scalers/rabbitmq-queue.md
+++ b/content/docs/1.4/scalers/rabbitmq-queue.md
@@ -41,11 +41,11 @@ AMQP protocol:
 apiVersion: keda.k8s.io/v1alpha1
 kind: ScaledObject
 metadata:
-  name: rabbitmq-scaledobject
+  name: rabbitmq-deployment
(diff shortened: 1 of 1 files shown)


tomkerkhove:
Hm this is actually not true, what gives you the impression that that is the case?

megakid:
We had issues getting this to work without these matching fields matching.

Supported by this completely unaffiliated blog post: https://www.gokhan-gokalp.com/en/kubernetes-based-event-driven-autoscaling-with-keda-rabbitmq-and-net-core/ 
`NOTE: Name values in the “metadata” and “scaleTargetRef” must have the same name as deployment.`

tomkerkhove:
Ha but the devil is in the details:
> NOTE: Name values in the “metadata” and “scaleTargetRef” must have the same name **as deployment**.

By default, `scaleTargetRef` is used for scaling deployments (but can be another resource type) and thus the name specified in `scaleTargetRef` must be identical to the name of your deployment, not scaled object.

PR #646, opened 2022-02-02
Fix an error in docs related with RabbitMQ Scaler

JorTurFer:

As we have been discussing [here](https://github.com/kedacore/keda/issues/2572), the docs are wrong because they say:
`> 💡 **Note:** `useRegex: "true"` requires protocol `http` and ignores unacknowledged messages`
But we are using the key `messages` which is already the total count of the messages, including unacknowledged

### Checklist

Fixes https://github.com/kedacore/keda/issues/2572

### DIFF
--- a/content/docs/2.4/scalers/rabbitmq-queue.md
+++ b/content/docs/2.4/scalers/rabbitmq-queue.md
@@ -47,7 +47,7 @@ Some parameters could be provided using environmental variables, instead of sett
 
 > 💡 **Note:** `mode: MessageRate` requires protocol `http`.
 
-> 💡 **Note:** `useRegex: "true"` requires protocol `http` and ignores unacknowledged messages.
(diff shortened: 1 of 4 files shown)


coderanger:
I think it's still correct to say that useRegex requires HTTP mode? No such similar feature exists in AMQP.

JorTurFer:
> I think it's still correct to say that useRegex requires HTTP mode? No such similar feature exists in AMQP.

Wow, you are right, I deleted all the line, not only the last part. Nice catch

coderanger (approved):

tomkerkhove (approved):

tomkerkhove:
Thanks for spotting @coderanger

JorTurFer:
Thanks @coderanger for your knowledge and your help

Issue #594, opened 2021-11-23
Use Google Analytics to better understand end-users


tomkerkhove:
Use Google Analytics to better understand end-users and where to prioritize improvements.

A notice on this should be added according to GDPR.

nate-double-u:
As per @chalin on the [CNCF Service Desk ticket 1008](https://cncfservicedesk.atlassian.net/browse/CNCFSD-1008?focusedCommentId=19231), there is a new GA tag: `G-YSWMV22VTQ`

tomkerkhove:
@JorTurFer Do you know how to configure it?

chalin:
- [ ] Add notice for GDPR compliance
- [ ] Let @tomkerkhove know how maintainers can access analytics

chalin:
Added a note above, regarding Knative's site.

chalin:
@tomkerkhove - FYI, I'm still working on this (for KEDA docs and other projects).

tomkerkhove:
Thanks for the update.

PR #653, opened 2022-02-10
Add analytics

chalin:
Contributes to #594 by adding analytics support. I'm investigating how best to bring in the cookie-consent scripts, styles and Hugo partials. Hopefully you'll agree that we can do this in two steps (with a small window of time between the two).

/cc @nate-double-u

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -4,6 +4,8 @@ title = "KEDA"
 enableGitInfo = true
 disableKinds = ["taxonomy", "taxonomyTerm"]
 
+googleAnalytics =  "G-YSWMV22VTQ"
+
 # See other available styles here: https://xyproto.github.io/splash/docs/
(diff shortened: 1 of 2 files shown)


chalin:
> Can you fix the DCO

Done.

tomkerkhove:
There we go!

tomkerkhove:
> I'm investigating how best to bring in the cookie-consent scripts, styles and Hugo partials. Hopefully you'll agree that we can do this in two steps (with a small window of time between the two).

No problem at all.

Would you mind sharing how maintainers can get access to the Google Analytics information please?

chalin:
> Would you mind sharing how maintainers can get access to the Google Analytics information please?

Yes. I'm looking into options and will let you know soon.

PR #654, opened 2022-02-10
docs: Add troubleshooting section explaining the new behavior on upstream errors

JorTurFer:

Related https://github.com/kedacore/keda/pull/2604


content/troubleshooting/scaledobject-not-scale.md
```diff
@@ -0,0 +1,10 @@
++++
+title = "Why is the `ScaledObject` paused?"
```
tomkerkhove, line 2: I would be careful when using "paused"
```suggestion
title = "Why is my `ScaledObject` not scaling?"
```

content/troubleshooting/scaledobject-not-scale.md
```diff
@@ -0,0 +1,10 @@
...
+weight = 1
++++
+
+We discovered a bug in how KEDA manages the upstream errors. By design we'd like to keep the things like they are in case of failure getting the metrics from the upstream and the `fallback` section is not defined.
+
+Before v2.6.1 the behavior in that case was scaling to zero the workload (which is risky) and [we have fixed](https://github.com/kedacore/keda/pull/2604) it on v2.6.1 so after it, the new behavior is keeping the things like they are till having more information. This behavior could seem that the `ScaledObject` is paused (nothing will be done).
```
tomkerkhove, line 8: Should we even bother covering the behavior before v2.6.1? Doesn't it make it a bit more confusing?

Alternative: Cover it, but only in a bulb after the new behavior was explained similar to this:

```markdown
> ⚠️ Are you using KEDA v2.6.0 or below? Then your workload will be scaled to 0. Because this can be tricky, we have changed this behavior to be more safe and not impact production workloads.
```

content/troubleshooting/scaledobject-not-scale.md
```diff
@@ -0,0 +1,13 @@
++++
+title = "Why is my `ScaledObject` paused?"
+weight = 1
++++
+
+We discovered a bug in how KEDA manages the upstream errors. By design we'd like to keep the things like they are in case of failure getting the metrics from the upstream and the `fallback` section is not defined.
+
+[We have fixed](https://github.com/kedacore/keda/pull/2604) it on v2.6.1 so after it, the new behavior is keeping the things like they are till having more information. This behavior could seem that the `ScaledObject` is paused (nothing will be done). 
+
+You can check if this is your case reviewing the logs from the KEDA pods, if you are facing with this, you should see some errors in both pods (operator and metrics server)
```
tomkerkhove, line 10: What about this?
```suggestion
When KEDA has upstream errors to get scaler source information it will keep the current instance count of the workload unless the `fallback` section is defined.

This behavior might feel like the autoscaling is not happening, but in reality, it is because of problems related to the scaler source.

You can check if this is your case by reviewing the logs from the KEDA pods where you should see errors in both our operator and metrics server.
```

content/troubleshooting/scaledobject-not-scale.md
```diff
@@ -0,0 +1,11 @@
++++
+title = "Why is my `ScaledObject` paused?"
+weight = 1
++++
+
+When KEDA has upstream errors to get scaler source information it will keep the current instance count of the workload unless the `fallback` section is defined.
+
+This behavior might feel like the autoscaling is not happening, but in reality, it is because of problems related to the scaler source.
+
+You can check if this is your case by reviewing the logs from the KEDA pods where you should see errors in both our operator and metrics server.
```
zroubalik, line 10:
```suggestion
You can check if this is your case by reviewing the logs from the KEDA pods where you should see errors in both our Operator and Metrics server. You can also check a status of the ScaledObject (`READY` and `ACTIVE` condition) by running following command:
```bash
$ kubectl get scaledobject MY-SCALED_OBJECT
```

tomkerkhove (approved):

content/troubleshooting/scaledobject-not-scale.md
```diff
@@ -0,0 +1,14 @@
...
++++
+
+When KEDA has upstream errors to get scaler source information it will keep the current instance count of the workload unless the `fallback` section is defined.
+
+This behavior might feel like the autoscaling is not happening, but in reality, it is because of problems related to the scaler source.
+
+You can check if this is your case by reviewing the logs from the KEDA pods where you should see errors in both our Operator and Metrics server. You can also check a status of the ScaledObject (`READY` and `ACTIVE` condition) by running following command:
+
+```bash
+$ kubectl get scaledobject MY-SCALED_OBJECT
```
tomkerkhove, line 13: Nit
```suggestion
$ kubectl get scaledobject MY-SCALED-OBJECT
```

PR #655, opened 2022-02-11
Datadog: Add Required min, max and interval for HPA

rr-paras-patel:
Without this change HPA wouldn't work.
 Better to provide complete example rather than assuming others will read lot of other common docs

Signed-off-by: Paras  Patel <parasprince2001@gmail.com>

### DIFF
--- a/content/docs/2.6/scalers/datadog.md
+++ b/content/docs/2.6/scalers/datadog.md
@@ -81,6 +81,15 @@ metadata:
 spec:
   scaleTargetRef:
     name: worker
+  minReplicaCount: 1
+  maxReplicaCount: 3
+  pollingInterval: 5
+  cooldownPeriod: 10
+  advanced:
+    horizontalPodAutoscalerConfig:
+      behavior:
+        scaleDown:
+          stabilizationWindowSeconds: 10
   triggers:
   - type: datadog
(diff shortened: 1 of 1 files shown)


tomkerkhove:
What change made it work for you? None of these values should be required.

rr-paras-patel:
Is there any default value for `maxReplicaCount` and `minReplicaCount` ? 
 Without this it was scaling replicas to 0.

tomkerkhove:
Yes, they are documented on https://keda.sh/docs/latest/concepts/scaling-deployments/#scaledobject-spec but `0` is indeed the default for `minReplicaCount`.

tomkerkhove:
Just out of curiosity - What was the issue you've had? Because you shouldn't have to add all of this configuration to get it working.

rr-paras-patel:
> Just out of curiosity - What was the issue you've had? Because you shouldn't have to add all of this configuration to get it working.

if someone use metric(e.g. RPS) then service need to be there at least with replica. otherwise technically it's not possible to generate traffic and report metric.

tomkerkhove:
THat is true, but the sample uses Redis so I don't really see why the addition is an added value over our typical ScaledObject docs?

PR #656, opened 2022-02-11
docs: KEDA v2.5+ requires Kubernetes v1.17

tomkerkhove:
KEDA v2.5+ requires Kubernetes v1.17

### Checklist

Relates to https://github.com/kedacore/charts/issues/241


tomkerkhove:
@zroubalik What about both? Cluster operators would go to operators while requirements are still in FAQ?

zroubalik:
> @zroubalik What about both? Cluster operators would go to operators while requirements are still in FAQ?

Yeah, that would be the best option!

zroubalik (approved):
LGTM

PR #661, opened 2022-02-17
Google analytics: make it work for our GA4 property

chalin:
- Contributes to #594 
- This is take 2 to get the GA4 property to work using the current version of Hugo (v0.67.0).
  If this doesn't work, we'll either need to **upgrade Hugo** (preferred) or use a custom template.
- Also, added template code so that only the production build gets tracked.

/cc @nate-double-u

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,8 @@
+build: clean
+	hugo -e development -DFE
+
 clean:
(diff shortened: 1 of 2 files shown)


chalin:
This change wasn't enough to get GA to work. Next step is #662.

PR #685, opened 2022-03-02
Added documentation for Azure Pipelines Demands

Eldarrin:
Signed-off-by: mortx <mortxbox@live.com>

Documents using parents to support demands in ADO

### Checklist

Relates to https://github.com/kedacore/keda/issues/2328
Relates to https://github.com/kedacore/keda/pull/2795


content/docs/2.7/scalers/azure-pipelines.md
```diff
@@ -37,6 +39,7 @@ triggers:
 - `personalAccessTokenFromEnv` - Name of the environment variable that provides the personal access token (PAT) for Azure DevOps. Learn more about how to create one [in the official docs](https://docs.microsoft.com/en-us/azure/devops/organizations/accounts/use-personal-access-tokens-to-authenticate?view=azure-devops&tabs=preview-page).
 - `targetPipelinesQueueLength` - Target value for the amount of pending jobs in the queue to scale on. (Default: `1`, Optional)
   - Example - If one pod can handle 10 jobs, set the queue length target to 10. If the actual number of jobs in the queue is 30, the scaler scales to 3 pods.
+- `parent` - Put the name of the ADO agent that matched the ScaledJob. e.g. mavenagent-scaledjob may have an initial deployment called "mavenagent-keda-template"; this is the deployment that is made offline.
```
tomkerkhove, line 42: Frankly, I don't really understand what is expected here or how it works.

tomkerkhove: Isn't it as simple to just provide a list of capabilities?

JorTurFer: I think that capabilities is the best approach because in fact, it's MS approach

Eldarrin: My first couple of attempts both use capabilities in the Job. This method means there is then an extra API call to check if the agent can fulfil the demand. Inside the actual JobRequest along with demands is a field called matchedAgent which are the agents that ADO has automatically matched. This means ADO does the evaluation for us and already tells us what can fulfil it.
The other part is this separates the "ADO" specific stuff into the Dockerfile and has KEDA just doing the KEDA stuff.

Eldarrin: This is the other way of doing it and considering ADO is rate limited could cause more issues IMHO
`				var demands = ""
				for _, v := range v["demands"].([]interface{}) {
					demands = demands + "," + strings.Trim(v.(string), " ")
				}
				// tidy for querysting and make url safe
				demands = strings.ReplaceAll(strings.TrimLeft(demands, ","), " ", "%20")

				// determine if this agent can satisfy the job
				urlAgent := fmt.Sprintf("%s/_apis/distributedtask/pools/%d/agents?agentName=%s&demands=%s", s.metadata.organizationURL, s.metadata.poolID, s.metadata.parent, demands)
				bodyAgent, err := getAzurePipelineRequest(ctx, urlAgent, s.metadata, s.httpClient)
				if err != nil {
					return -1, err
				}

				var resultAgent map[string]interface{}
				err = json.Unmarshal(bodyAgent, &resultAgent)
				if err != nil {
					return -1, err
				}

				countAgent := int(resultAgent["count"].(float64))
				if countAgent > 0 {
					count++
				}`

JorTurFer: I'd add a list of capabilities inside metadata and I'd check them parsing the response. I have done the request and this is a sample of the response:
```json
{
   "count":1,
   "value":[
      {
         "requestId":1493,
         "queueTime":"............................",
         "serviceOwner":"...........................",
         "hostId":"..............................",
         "scopeId":".......................",
         "planType":"Build",
         "planId":"..................",
         "jobId":"..................",
         "demands":[
            "Agent.Version -gtVersion 2.144.0"
         ],
         "matchedAgents":[
            {
               "_links":{
                  "self":{
                     "href":"https://dev.azure.com/...................../_apis/distributedtask/pools/13/agents/161"
                  },
                  "web":{
                     "href":"https://dev.azure.com/...................../_settings/agentpools?view=jobs&poolId=13&agentId=161"
                  }
               },
               "id":161,
               "name":"test-deployment-8474f4bd57-2gcj2",
               "version":"2.196.2",
               "enabled":true,
               "status":"offline",
               "provisioningState":"Provisioned"
            }
         ],
         "definition":{
            "_links":{
               "web":{
                  "href":"https://dev.azure.com/..................................../_build/definition?definitionId=29"
               },
               "self":{
                  "href":"https://dev.azure.com/...................................../_apis/build/Definitions/29"
               }
            },
            "id":29,
            "name":"KEDA-CI"
         },
         "owner":{
            "_links":{
               "web":{
                  "href":"https://dev.azure.com/................................/_build/results?buildId=433"
               },
               "self":{
                  "href":"https://dev.azure.com/............................./_apis/build/Builds/433"
               }
            },
            "id":433,
            "name":"433"
         },
         "data":{
            "ParallelismTag":"Public",
            "IsScheduledKey":"False"
         },
         "poolId":13,
         "orchestrationId":"eee66f1e-5e66-4003-a1c3-a9da4b6e844f.job_1.__default",
         "matchesAllAgentsInPool":true,
         "priority":0
      }
   ]
}
```

I'm not 100% sure because I couldn't test it but I think that each element (job) contains the demands under the section `demands`. Maybe we could iterate the result and calculate how many pending jobs demand the supported capabilities.
This approach doesn't add any extra request to the Api so the limit rate should not be a problem.
WDYT?

Eldarrin: That was actually where I started too; and it does work. The problem is that you have to check the Job's demands from the request against the named demands in your metadata and you cannot be sure they are accurate. If you log into `ADO -> Project Settings -> Agent Pools -> PoolName -> Agents -> AgentName -> Capabilities`.
This is the list of demands that the Agent actually supports. These cannot all be codified into your ScaledJob metadata as anytime an agent's environment variables change, so do the demands.
This does work if you have a specific named demand to target an agent, such as the ScaledJob has a demand called `"keda-maven"` and your job has a specific demand of the same. Evaluating all the others is the issue.
**This was attempt 1.**

**Attempt 2:**
As above, but using the Job Request to ask ADO which agents support the requested demands; this used the parent methodology as in the PR but adds an extra API call to interrogate ADO.

**Attempt 3:**
ADO has already told you the list of agents that can do the job. This is in the matchedAgents field. As long as 1 of those agents can be related to the ScaledJob (I use a template agent and the parent metadata) then you know which to spin up.

JorTurFer: I get your point, but I still have doubts about this approach.
Any thoughts @kedacore/keda-contributors ?

JorTurFer: What about using deployments instead of statefulset for agents? I mean, the agent register itself using the hostname if I'm not wrong. How will this affect in that case?

JorTurFer: any  update here?

Eldarrin: good point, the environment variable "AZP_NAME" has to be set. I'll update the docs

JorTurFer: Any thoughts https://github.com/orgs/kedacore/teams/keda-contributors ?

JorTurFer: Any thoughts @kedacore/keda-contributors ?

tomkerkhove: I think that's a fair requirement to have `AZP_NAME`. However, does that have to be on the agent running (end-user deployment) or something we can do from KEDA?

I think it's the first but double-checking.

Eldarrin: Its the agent as that is what get passed into the job call that ADO actually makes and it reads it from its registered list inside the ADO project

content/docs/2.8/scalers/azure-pipelines.md
```diff
@@ -60,7 +66,24 @@ The URL should be similar to `https://dev.azure.com/{organization}/_settings/age
 
 Finally, it is also possible get the pool ID from the response of a HTTP request by calling the `https://dev.azure.com/{organizationName}/_apis/distributedtask/pools?poolname={agentPoolName}` endpoint in the key `value[0].id`.
 
-### Example
+### Supporting Demands in Agents ###
```
tomkerkhove, line 69:
```suggestion
### Supporting demands in agents
```

content/docs/2.8/scalers/azure-pipelines.md
```diff
@@ -60,7 +66,24 @@ The URL should be similar to `https://dev.azure.com/{organization}/_settings/age
...
+Demand List: KEDA will determine which agents can fulfil the job based on the demands provided. The demands are provided as a comma separated list and must be a subset of the actual capabilities of the agent. (Note: Agent.Version is ignored)
+
+Evaluation order is: 
+1) If neither parent nor demands is set in the ScaledJob is will scale up the ScaledJob to fulfil the job. 
+2) If parent is set in the ScaledJob then KEDA will interrogate the job request to determine if the parent is able to fulfil the job. If the parent is able to complete the job it scales up the ScaledJob to fulfil it. 
+3) Finally, if the demands are set in the ScaledJob then KEDA will determine which agents can fulfil the job based on the demands provided.
+
+> Note: If more than one ScaledJob is able to fulfil the demands of the job then they will both spin up an agent; this is as each ScaledJob is independent of each other.
+
+#### How it works under the hood.
```
tomkerkhove, line 82:
```suggestion
#### How it works under the hood
```

content/docs/2.8/scalers/azure-pipelines.md
```diff
@@ -60,7 +66,24 @@ The URL should be similar to `https://dev.azure.com/{organization}/_settings/age
...
+
+Azure Devops has a Job Request API with returns a list of all agents which can potentially match up to the Queued Job. KEDA will scan this request to find any matching template from the defined parent in the ScaledJob definition. Once it finds it, it will spin up a ScaledJob that matched the definition and Azure DevOps will assign to that agent.
```
tomkerkhove, line 84: Can we add a link to the API?

content/docs/2.8/scalers/azure-pipelines.md
```diff
@@ -60,7 +66,24 @@ The URL should be similar to `https://dev.azure.com/{organization}/_settings/age
 
 Finally, it is also possible get the pool ID from the response of a HTTP request by calling the `https://dev.azure.com/{organizationName}/_apis/distributedtask/pools?poolname={agentPoolName}` endpoint in the key `value[0].id`.
 
-### Example
+### Supporting Demands in Agents ###
+
+Parent Template: Azure Devops is able to determine which agents can match any job its is waiting for. If you specify a parent template then KEDA will further interrogate the job request to determine if the parent is able to fulfil the job. If the parent is able to complete the job it scales up the ScaledJob to fulfil it. The parent template is generally offline must stays in the Pool's Agent list.
+
+Demand List: KEDA will determine which agents can fulfil the job based on the demands provided. The demands are provided as a comma separated list and must be a subset of the actual capabilities of the agent. (Note: Agent.Version is ignored)
```
tomkerkhove, line 73:
```suggestion
TODO: Add some intro here before listing approaches:

- **Using Parent:** Azure DevOps is able to determine which agents can match any job it is waiting for. If you specify a parent template then KEDA will further interrogate the job request to determine if the parent is able to fulfill the job. If the parent is able to complete the job it scales the workload fulfill the request. The parent template that is generally offline must stay in the Pool's Agent list.

- **Using demands:** KEDA will determine which agents can fulfill the job based on the demands provided. The demands are provided as a comma-separated list and must be a subset of the actual capabilities of the agent. (For example `someexamplehere`. Note: `Agent.Version` is ignored)
```

Can we add a link to the official parent & demands doc per entry?

content/docs/2.8/scalers/azure-pipelines.md
```diff
@@ -60,7 +66,24 @@ The URL should be similar to `https://dev.azure.com/{organization}/_settings/age
...
+
+Evaluation order is: 
+1) If neither parent nor demands is set in the ScaledJob is will scale up the ScaledJob to fulfil the job. 
+2) If parent is set in the ScaledJob then KEDA will interrogate the job request to determine if the parent is able to fulfil the job. If the parent is able to complete the job it scales up the ScaledJob to fulfil it. 
+3) Finally, if the demands are set in the ScaledJob then KEDA will determine which agents can fulfil the job based on the demands provided.
+
+> Note: If more than one ScaledJob is able to fulfil the demands of the job then they will both spin up an agent; this is as each ScaledJob is independent of each other.
```
tomkerkhove, line 80:
```suggestion
KEDA will use the following evaluation order is: 
1) If neither parent nor demands are defined in the scaling definition, it will scale the workload to fulfill the job. 
2) If `parent` is set,  KEDA will interrogate the job request to determine if the parent is able to fulfill the job. If the parent is able to complete the job it scales the workload to fulfill the request. 
3) Finally, if the demands are set in the scaling definition then KEDA will determine which agents can fulfill the job based on the demands provided.

> Note: If more than one scaling definition is able to fulfill the demands of the job then they will both spin up an agent.
```

tomkerkhove (changes requested):
Made some editorial suggestions to make it easier to read and remove the direct references to ScaledJob since you can also use it with ScaledObject

content/docs/2.8/scalers/azure-pipelines.md
Eldarrin: I can add links for demands, but parent is an internal concept to support this model, its not Microsoft's

content/docs/2.8/scalers/azure-pipelines.md
Eldarrin: This is a bit of aa weird one. I'm using the same API that keda has always used, but this API is actually not disclosed by Microsoft. There API stops at /pool/id; there is a deeper internal API which is /pool/id/JobRequests. This is undocumented on the MS site I believe

content/docs/2.8/scalers/azure-pipelines.md
```diff
@@ -37,6 +41,8 @@ triggers:
 - `personalAccessTokenFromEnv` - Name of the environment variable that provides the personal access token (PAT) for Azure DevOps. Learn more about how to create one [in the official docs](https://docs.microsoft.com/en-us/azure/devops/organizations/accounts/use-personal-access-tokens-to-authenticate?view=azure-devops&tabs=preview-page).
 - `targetPipelinesQueueLength` - Target value for the amount of pending jobs in the queue to scale on. (Default: `1`, Optional)
   - Example - If one pod can handle 10 jobs, set the queue length target to 10. If the actual number of jobs in the queue is 30, the scaler scales to 3 pods.
+- `parent` - Put the name of the ADO agent that matched the ScaledJob. e.g. mavenagent-scaledjob may have an initial deployment called "mavenagent-keda-template"; this is the deployment that is made offline. This name is provided to the initial deployment as the environment variable "AZP_NAME"
```
JorTurFer, line 44: I'd not say `ScaledJob` because this trigger can be used to in ScaledObject

content/docs/2.8/scalers/azure-pipelines.md
```diff
@@ -37,6 +41,8 @@ triggers:
...
+- `demands` - Put the demands string that was provided to the ScaledJob. This MUST be a subset of the actual capability list the agent has. e.g. `maven,docker`
```
JorTurFer, line 45: Same as a above

content/docs/2.8/scalers/azure-pipelines.md
```diff
@@ -97,6 +146,32 @@ spec:
     metadata:
       poolID: "1"
       organizationURLFromEnv: "AZP_URL"
+      parent: "example-keda-template"
+      demands: "maven,docker"
     authenticationRef:
      name: pipeline-trigger-auth
 ```
+###Example for Parent Deployment or StatefulSet
```
JorTurFer, line 154: I'm not sure about having the requirement of an agent for doing nothing TBH. I prefer to register the agent manually in the AzDO UI than creating a deployment only for that.
WDYT @kedacore/keda-contributors ?

JorTurFer: My concern is that you have your own business logic to manage this and it works with your use case, but it's managed by your side, not by KEDA. IMO,  we should be most agnostics as possible and use the upstream tools for doing the upstream things. Obviously, we can add a "tip" suggesting the approach, having to deploy (and potentially delete) the deployment only to upgrade the agent in AzDO could be a complex action if you only have 2-3 different agents.

Maybe I'm overthinking about this and that's why I'd like to get the opinion of other @kedacore/keda-contributors but it's more a thing about how we document the feature than an issue with the feature itself (In fact, it's already approved :) ).

JorTurFer:
@Eldarrin 
We plan to do a release next week, do you think that you can address this comments before it?

Eldarrin:
@JorTurFer Conversation 1, I can resolve if your happy that that the API is undocumented on the MS side. Conversation 2, I thought you were waiting for comments from other contributors? But can resolve

JorTurFer:
Let's solve the first point, and we can open another issue to improve this in the future, but at least next release will have this feature

JorTurFer:
Please, solve also the signing issue in your commits 🙏

Eldarrin:
.... abandon pr and start over?

JorTurFer:
> .... abandon pr and start over? followed the instructions to rebase and it blew up

Maybe squashing the branch, you could solve it. If not, feel free to close this and open another

Eldarrin:
Closing to re-add just azure_pipelines.md as new PR

PR #688, opened 2022-03-04
Add Stilingue as KEDA user

luistilingue:
Hi.

We're using KEDA in our production environment \o/


JorTurFer (approved):
Awesome, thank you!

tomkerkhove (approved):

tomkerkhove:
Woop, woop - Thanks for using KEDA!

tomkerkhove:
@luistilingue We are looking for KEDA end-users who want to be interviewed by CNCF for our CNCF Graduation proposal.

PR #690, opened 2022-03-04
Netlify CLI: add framework flag

chalin:
When invoking `npm run serve`, the [Netlify-CLI dev command](https://cli.netlify.com/commands/dev) prompts us to choose a command, but it shouldn't because we explicitly provide a command via `-c`. We seem to be hitting the following issue: https://github.com/netlify/cli/issues/410. This PR implements the recommended workaround.

/cc @nate-double-u


JorTurFer (approved):

JorTurFer:
Thanks!!!

PR #701, opened 2022-03-07
Enable link checking for preview and production builds, adjust Netlify config

chalin:
- **Enable link checking** for preview and production builds.
- Make targets using `hugo` were (potentially) picking up the wrong version of hugo; the proper command would be `npx hugo`, but we don't need those targets anymore since we have NPM scripts for each of them.


chalin:
This is ready for review @tomkerkhove.

tomkerkhove (approved):

PR #733, opened 2022-03-27
ActivationValue for RabbitMQ scaler

adborroto:
Signed-off-by: Alejandro Dominguez <adborroto90@gmail.com>

Adding activation threshold (minMetricValue) for Rabbitmq. 

Feature PR: [#2831](https://github.com/kedacore/keda/pull/2831)

### Checklist

Fixes #

### DIFF
--- a/content/docs/2.7/scalers/rabbitmq-queue.md
+++ b/content/docs/2.7/scalers/rabbitmq-queue.md
@@ -19,6 +19,7 @@ triggers:
     protocol: auto # Optional. Specifies protocol to use, either amqp or http, or auto to autodetect based on the `host` value. Default value is auto.
     mode: QueueLength # QueueLength or MessageRate
     value: "100" # message backlog or publish/sec. target per instance
+    minMetricValue: "0.0" # Optional. Activation threshold 
     queueName: testqueue
     vhostName: / # Optional. If not specified, use the vhost in the `host` connection string.
     # Alternatively, you can use existing environment variables to read configuration from:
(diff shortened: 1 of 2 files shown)


content/docs/2.7/scalers/rabbitmq-queue.md
```diff
@@ -32,6 +33,7 @@ triggers:
 - `queueName` - Name of the queue to read message from.
 - `mode` - QueueLength to trigger on number of messages in the queue. MessageRate to trigger on the publish rate into the queue. (Values: `QueueLength`, `MessageRate`)
 - `value` - Message backlog or Publish/sec. rate to trigger on.
+- `activationValue` - Min value to activate the scaler. (Default: `0`, Optional)
```
JorTurFer, line 36: "Activate the scaler" could be confusing, I guess, WDYT about `Message backlog or Publish/sec. rate to scale from/to zero`?

zroubalik (commented):
I think we should have a generic section on `activation threshold` in the docs. Where we describe the behavior (current and the new). And here reference this doc.
WDYT? @kedacore/keda-contributors

adborroto:
> I think we should have a generic section on `activation threshold` in the docs. Where we describe the behavior (current and the new). And here reference this doc. WDYT? @kedacore/keda-contributors

I like it, since, is going to be a generic concept for all scalers.

JorTurFer:
+1 to a specific section

adborroto:
> I think we should have a generic section on `activation threshold` in the docs. Where we describe the behavior (current and the new). And here reference this doc. WDYT? @kedacore/keda-contributors

What do you think about add the section in a different PR/ticket. Since this is only affecting a single scaler. So we don't block this PR and also because not sure where this new section should be placed.

zroubalik:
A new PR is fine, but this general section should be added before this feature lands.

@kedacore/keda-contributors ideas on the docs section?

JorTurFer:
Maybe we can explain it under [Scaling Deployments, StatefulSets & Custom Resources](https://keda.sh/docs/2.6/concepts/scaling-deployments/) due to this only applies to ScaledObjects.
Something like "Differences between activation and scaling"

content/docs/2.7/scalers/rabbitmq-queue.md
JorTurFer, line 36:
```suggestion
- `activationValue` - Target value for activating the scaler. Learn more about activation [here](./../concepts/scaling-deployments.md#activating-and-scaling-thresholds).(Default: `0`, Optional, This value can be a float)
```

JorTurFer: For the documentation, we are using the same definition for all the scalers, I have suggested it

JorTurFer (approved):

PR #749, opened 2022-04-20
adding corsHeader documentation for keda issue #2884

xphyr:
Signed-off-by: Mark DeNeve <xphyr@users.noreply.github.com>

Small documentation update for ActiveMQ and Artemis documentation to call out the CoreHeader configuration item. This documentation change addresses the new configuration item "corsHeader" that is a part of the ActiveMQ scaler, and adds documentation for the _existing_ Artemis configuration item of the same name.

### Checklist

Fixes # Keda Issue #2884


xphyr:
I just realized that I added this doc update to 2.7, without knowing where the change will be released. If the actual change will be in a point release of 2.6 let me know and I will add the change to the 2.6 docs as well.

xphyr:
> Could you please update the artemis docs for the previous versions as well? (the cors header is present there).

@zroubalik - Done. It appears that the corsHeader was added to Artemis back in release 2.3 so I have updated the docs going all the way back to 2.3. I also fixed a small formatting error in the 2.7 docs where I had one two many spaces in the list.

zroubalik (approved):

PR #751, opened 2022-04-25
docs: Provide "Why does KEDA use external metrics and not custom metrics instead?" FAQ

tomkerkhove:
Provide long overdue "Why does KEDA use external metrics and not custom metrics instead?" FAQ to explain difference between external/custom metrics and why we use external.

### Checklist


zroubalik (commented):
Looking good, though I would add a note, that custom vs external is just an implementation detail, it has the same functionality.
So using just external metrics doesn't have any impact.

JorTurFer (approved):

PR #754, opened 2022-05-04
Update Helm uninstall procedure

adrianparks:
Update Helm uninstall procedure to cover the removal of ScaledObjects before (or after) the Helm chart is removed, as discussed in https://github.com/kedacore/keda/discussions/2301. I've made an update to cover 2.6, but I note docs for 2.7 are in place now too. If the proposed changes are acceptable, I can also duplicate them to the 2.7 deployment doc, if desired.

### DIFF
--- a/content/docs/2.6/deploy.md
+++ b/content/docs/2.6/deploy.md
@@ -43,10 +43,19 @@ Deploying KEDA with Helm is very simple:
 
 ### Uninstall
 
-If you want to remove KEDA from a cluster you can run one of the following:
+If you want to remove KEDA from a cluster, you first need to remove any ScaledObjects that you have created. Once that is done, the Helm chart can be uninstalled:
 
 ```sh
+kubectl delete $(kubectl get scaledobject -oname)
 helm uninstall keda -n keda
+```
+
+Note: if you uninstall the Helm chart without first deleting the ScaledObjects, they will become orphaned. In this situation, you will need to patch the ScaledObjects to remove their finalizers. Once this is done, they should automatically be removed:
+
+```sh
+for i in $(kubectl get scaledobject | awk '{ print $1 }' | awk 'NR!=1 {print}');
+do kubectl patch scaledobject $i -p '{"metadata":{"finalizers":null}}' --type=merge
+done
 ```
 
 ## Deploying with Operator Hub {#operatorhub}
--- a/content/docs/2.6/deploy.md
+++ b/content/docs/2.6/deploy.md
@@ -43,18 +43,22 @@ Deploying KEDA with Helm is very simple:
 
 ### Uninstall
 
-If you want to remove KEDA from a cluster, you first need to remove any ScaledObjects that you have created. Once that is done, the Helm chart can be uninstalled:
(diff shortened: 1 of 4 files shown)


zroubalik (approved):
Looking good! Could you please duplicate the same for docs 2.4-2.7? Thanks!

content/docs/2.6/deploy.md
```diff
@@ -43,12 +43,21 @@ Deploying KEDA with Helm is very simple:
 
 ### Uninstall
 
-If you want to remove KEDA from a cluster you can run one of the following:
+If you want to remove KEDA from a cluster, you first need to remove any ScaledObjects that you have created. Once that is done, the Helm chart can be uninstalled:
```
JorTurFer, line 46: You have to remove ALL KEDA custom resources, ScaledObjects, ScaledJobs, TriggerAuthentication and ClusterTriggerAuthentication. 
Should we add all of them?
@zroubalik

content/docs/2.6/deploy.md
```diff
@@ -43,12 +43,21 @@ Deploying KEDA with Helm is very simple:
...
 
 ```sh
+kubectl delete $(kubectl get scaledobject -oname)
 helm uninstall keda -n keda
 ```
 
+Note: if you uninstall the Helm chart without first deleting the ScaledObjects, they will become orphaned. In this situation, you will need to patch the ScaledObjects to remove their finalizers. Once this is done, they should automatically be removed:
```
JorTurFer, line 53: Same as above

content/docs/2.6/deploy.md
adrianparks: I didn't find that to be the case, @JorTurFer - although with that said, in my cluster I did only have TriggerAuthentication objects in addition to my ScaledObjects. The TriggerAuthentication resources I created after installing the Helm chart were cleanly removed when the Helm chart was uninstalled.

adrianparks: ...although on reflection, I suppose it does no harm to clean them out first anyway, I can update the PR accordingly

JorTurFer: you are right, not all resources have finalizers, only ScaledObject and ScaledJob, so we only need to add those 2

content/docs/2.6/deploy.md
```diff
@@ -43,12 +43,21 @@ Deploying KEDA with Helm is very simple:
 
 ### Uninstall
 
-If you want to remove KEDA from a cluster you can run one of the following:
+If you want to remove KEDA from a cluster, you first need to remove any ScaledObjects that you have created. Once that is done, the Helm chart can be uninstalled:
 
 ```sh
+kubectl delete $(kubectl get scaledobject -oname)
```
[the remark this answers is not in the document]
adrianparks: updated PR accordingly

content/docs/2.6/deploy.md
adrianparks: updated PR accordingly

JorTurFer (commented):
LGTM ❤️ ❤️ ❤️ 
Only one thing, we have released v2.7 some hours ago, could you rebase main branch and add the changes also to version v2.8?

adrianparks:
Closing pending a new PR

PR #780, opened 2022-06-13
docs: explain the differences between activation and scaling values

JorTurFer:
Signed-off-by: Jorge Turrado <jorge_turrado@hotmail.es>

### Checklist

Related https://github.com/kedacore/keda/issues/2800

### DIFF
--- a/content/docs/2.8/concepts/scaling-deployments.md
+++ b/content/docs/2.8/concepts/scaling-deployments.md
@@ -228,14 +228,14 @@ The presensce of this annotation will pause autoscaling no matter what number of
 
 ### Activating and Scaling thresholds
 
-There are user cases where the activating value (0-1 and 1-0) is totally different than 0, ie: workloads scaled with prometheus scaler where the values go from -X to X. KEDA brings support to specify different values for each case:
+There are use-cases where the activating value (0-1 and 1-0) is totally different than 0, such as workloads scaled with the Prometheus scaler where the values go from -X to X. KEDA allows you to specify different values for each scenario:
 
-- Activation: This value will be used to consider the scaler active or not and scale from/to 0 based on it.
-- Scaling: This value will be passed to the HPA as the target value and the HPA controller will try to reach it.
+- **Activation:** Defines when the scaler is active or not and scales from/to 0 based on it.
+- **Scaling:** Defines the target value to scale the workload from 1 to _n_ instances and vice versa. To achieve this, KEDA passes the target value to the Horizontal Pod Autoscaler (HPA) and the built-in HPA controller will handle all the autoscaling.
 
 > ⚠️ **NOTE:** If the minimum replicas is >= 1, the scaler is always active and the Activation value will be ignored.
 
-Each scaler defines in its own docs the proper parameters for each case, but the key for specifying the activation value will be always the same as the scaling value, appending the prefix `activation` (ie: `threshold` for scaling and `activationThreshold` for activation).
+Each scaler defines parameters for their use-cases, but the activation can always be the same as the scaling value, appended by the prefix `activation` (ie: `threshold` for scaling and `activationThreshold` for activation).
 
 There are some important topics to take into account:
 
--- a/content/docs/2.8/concepts/scaling-deployments.md
+++ b/content/docs/2.8/concepts/scaling-deployments.md
@@ -226,9 +226,24 @@ metadata:
 
 The presensce of this annotation will pause autoscaling no matter what number of replicas is provided. The above annotation will scale your current workload to 0 replicas and pause autoscaling. You can set the value of replicas for an object to be paused at to any arbitary number. To enable autoscaling again, simply remove the annotation from the `ScaledObject`definition.
 
-### `Activating` and `Scaling` thresholds
+### `Activating` and `Scaling`
 
-There are use-cases where the activating value (0-1 and 1-0) is totally different than 0, such as workloads scaled with the Prometheus scaler where the values go from -X to X. KEDA allows you to specify different values for each scenario:
+There are use-cases where the activating value (0-1 and 1-0) is totally different than 0, such as workloads scaled with the Prometheus scaler where the values go from -X to X. To give a consisten solution to this problem, KEDA has 2 different phases during the autoscaling process.
+
+- **Activation phase**
+- **Scaling phase**
+
+#### **Activating pahse**
+
+The activating (or deactivating) phase is the moment when KEDA (operator) has to decide if the workload should be scaled from/to zero. KEDA is the responsible for this action based on the result of the scaler `IsActive` function and only applies in from/to 0 scaling.
+
+#### **Scaling pahse**
+
+The scaling phase is the moment when KEDA has decided to scale out to 1 instance and now is the HPA controller who takes the scaling decisions based on the configuration defined in the generated HPA (from ScaledObject data) and the metrics exposed by KEDA (metrics server). This phase applies the scaling from 1-N and N-1.
+
+#### **Thresholds**
+
+KEDA allows you to specify different values for each scenario:
 
 - **Activation:** Defines when the scaler is active or not and scales from/to 0 based on it.
 - **Scaling:** Defines the target value to scale the workload from 1 to _n_ instances and vice versa. To achieve this, KEDA passes the target value to the Horizontal Pod Autoscaler (HPA) and the built-in HPA controller will handle all the autoscaling.
--- a/content/docs/2.8/concepts/scaling-deployments.md
+++ b/content/docs/2.8/concepts/scaling-deployments.md
@@ -233,14 +233,14 @@ There are use-cases where the activating value (0-1 and 1-0) is totally differen
 - **Activation:** Defines when the scaler is active or not and scales from/to 0 based on it.
 - **Scaling:** Defines the target value to scale the workload from 1 to _n_ instances and vice versa. To achieve this, KEDA passes the target value to the Horizontal Pod Autoscaler (HPA) and the built-in HPA controller will handle all the autoscaling.
 
-> ⚠️ **NOTE:** If the minimum replicas is >= 1, the scaler is always active and the Activation value will be ignored.
+> ⚠️ **NOTE:** If the minimum replicas is >= 1, the scaler is always active and the activation value will be ignored.
 
 Each scaler defines parameters for their use-cases, but the activation can always be the same as the scaling value, appended by the prefix `activation` (ie: `threshold` for scaling and `activationThreshold` for activation).
 
 There are some important topics to take into account:
 
-- Opposite to Scaling value, the Activation value is always optional and the default value is 0.
-- The Activation value has more priority than the Scaling value in case of different decisions for each. ie: `threshold: 10` and `activationThreshold: 50`, in case of 40 messages the scaler is not active and it'll be scaled to zero even the HPA requires 4 instances.
+- Opposite to scaling value, the activation value is always optional and the default value is 0.
+- The activation value has more priority than the scaling value in case of different decisions for each. ie: `threshold: 10` and `activationThreshold: 50`, in case of 40 messages the scaler is not active and it'll be scaled to zero even the HPA requires 4 instances.
 
 ## Long-running executions
 
--- a/content/docs/2.8/concepts/scaling-deployments.md
+++ b/content/docs/2.8/concepts/scaling-deployments.md
@@ -242,7 +242,6 @@ There are some important topics to take into account:
 - Opposite to Scaling value, the Activation value is always optional and the default value is 0.
 - The Activation value has more priority than the Scaling value in case of different decisions for each. ie: `threshold: 10` and `activationThreshold: 50`, in case of 40 messages the scaler is not active and it'll be scaled to zero even the HPA requires 4 instances.
 
-
 ## Long-running executions
 
 One important consideration to make is how this pattern can work with long running executions.  Imagine a deployment triggers on a RabbitMQ queue message.  Each message takes 3 hours to process.  It's possible that if many queue messages arrive, KEDA will help drive scaling out to many replicas - let's say 4.  Now the HPA makes a decision to scale down from 4 replicas to 2.  There is no way to control which of the 2 replicas get terminated to scale down.  That means the HPA may attempt to terminate a replica that is 2.9 hours into processing a 3 hour queue message.
--- a/content/docs/2.8/concepts/scaling-deployments.md
+++ b/content/docs/2.8/concepts/scaling-deployments.md
@@ -232,11 +232,11 @@ metadata:
 
 The presensce of this annotation will pause autoscaling no matter what number of replicas is provided. The above annotation will scale your current workload to 0 replicas and pause autoscaling. You can set the value of replicas for an object to be paused at to any arbitary number. To enable autoscaling again, simply remove the annotation from the `ScaledObject`definition.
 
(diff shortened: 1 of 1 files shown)


content/docs/2.8/concepts/scaling-deployments.md
```diff
@@ -226,6 +226,22 @@ metadata:
 
 The presensce of this annotation will pause autoscaling no matter what number of replicas is provided. The above annotation will scale your current workload to 0 replicas and pause autoscaling. You can set the value of replicas for an object to be paused at to any arbitary number. To enable autoscaling again, simply remove the annotation from the `ScaledObject`definition.
 
+### Activating and Scaling thresholds
```
tomkerkhove, line 229: Would you mind updating all versions please?

content/docs/2.8/concepts/scaling-deployments.md
```diff
@@ -226,6 +226,22 @@ metadata:
...
+
+There are user cases where the activating value (0-1 and 1-0) is totally different than 0, ie: workloads scaled with prometheus scaler where the values go from -X to X. KEDA brings support to specify different values for each case:
+
+- Activation: This value will be used to consider the scaler active or not and scale from/to 0 based on it.
+- Scaling: This value will be passed to the HPA as the target value and the HPA controller will try to reach it.
```
tomkerkhove, line 234: What about this?
```suggestion
- **Activation:** Defines when the scaler is active or not and scales from/to 0 based on it.
- **Scaling:** Defines the target value to scale the workload from 1 to _n_ instances and vice versa. To achieve this, KEDA passes the target value to the Horizontal Pod Autoscaler (HPA) and the built-in HPA controller will handle all the autoscaling.
```

content/docs/2.8/concepts/scaling-deployments.md
```diff
@@ -226,6 +226,22 @@ metadata:
...
+
+> ⚠️ **NOTE:** If the minimum replicas is >= 1, the scaler is always active and the Activation value will be ignored.
+
+Each scaler defines in its own docs the proper parameters for each case, but the key for specifying the activation value will be always the same as the scaling value, appending the prefix `activation` (ie: `threshold` for scaling and `activationThreshold` for activation).
```
tomkerkhove, line 238: What about this? However, not fully sure as this sentence is a bit confusing.
```suggestion
Each scaler defines parameters for their use-cases, but the activation can always be the same as the scaling value, appended by the prefix `activation` (ie: `threshold` for scaling and `activationThreshold` for activation).
```

content/docs/2.8/concepts/scaling-deployments.md
tomkerkhove, line 229: We should maybe use code formatting for configuration names

content/docs/2.8/concepts/scaling-deployments.md
```diff
@@ -226,6 +226,22 @@ metadata:
 
 The presensce of this annotation will pause autoscaling no matter what number of replicas is provided. The above annotation will scale your current workload to 0 replicas and pause autoscaling. You can set the value of replicas for an object to be paused at to any arbitary number. To enable autoscaling again, simply remove the annotation from the `ScaledObject`definition.
 
+### Activating and Scaling thresholds
+
+There are user cases where the activating value (0-1 and 1-0) is totally different than 0, ie: workloads scaled with prometheus scaler where the values go from -X to X. KEDA brings support to specify different values for each case:
```
tomkerkhove, line 231: Before we dive in to this, I would conceptually explain activation vs scaling, or point to our existing docs on this

content/docs/2.8/concepts/scaling-deployments.md
JorTurFer: No no, this change applies only after next version because requires changes on every scaler

content/docs/2.8/concepts/scaling-deployments.md
JorTurFer: Isn't enough the explanation below? I can expand it

tomkerkhove: No because it only comes later on, so people who are new to this might have no idea what we are talking about

content/docs/2.8/concepts/scaling-deployments.md
tomkerkhove: This comment was more for config names in this section, though. I think "Activating and Scaling thresholds" as a title as perfect

content/docs/2.8/concepts/scaling-deployments.md
zroubalik: Agree with Tom on this one. We should start with explaining what is Activation and what is Scaling, then we can follow up with a different options.

JorTurFer:
PTAL @zroubalik @tomkerkhove ,
I have refactored it a bit from the last review to explain better what means activating and what means scaling

content/docs/2.8/concepts/scaling-deployments.md
tomkerkhove: Can we rename it back?

content/docs/2.8/concepts/scaling-deployments.md
```diff
@@ -226,6 +226,37 @@ metadata:
...
+- **Activation phase**
+- **Scaling phase**
+
+#### **Activating pahse**
+
+The activating (or deactivating) phase is the moment when KEDA (operator) has to decide if the workload should be scaled from/to zero. KEDA is the responsible for this action based on the result of the scaler `IsActive` function and only applies in from/to 0 scaling.
+
+#### **Scaling pahse**
+
+The scaling phase is the moment when KEDA has decided to scale out to 1 instance and now is the HPA controller who takes the scaling decisions based on the configuration defined in the generated HPA (from ScaledObject data) and the metrics exposed by KEDA (metrics server). This phase applies the scaling from 1-N and N-1.
```
tomkerkhove, line 242: Let's merge all of this in just the list to keep it simpler, maybe?

content/docs/2.8/concepts/scaling-deployments.md
```diff
@@ -226,6 +226,37 @@ metadata:
...
+
+#### **Thresholds**
```
tomkerkhove, line 244: No need for making it bold
```suggestion
#### Managing Activation & Scaling Thresholds
```

content/docs/2.8/concepts/scaling-deployments.md
```diff
@@ -226,6 +226,37 @@ metadata:
...
+
+KEDA allows you to specify different values for each scenario:
+
+- **Activation:** Defines when the scaler is active or not and scales from/to 0 based on it.
+- **Scaling:** Defines the target value to scale the workload from 1 to _n_ instances and vice versa. To achieve this, KEDA passes the target value to the Horizontal Pod Autoscaler (HPA) and the built-in HPA controller will handle all the autoscaling.
```
tomkerkhove, line 249: We should use the configuration values here, for example:
```suggestion
- `activationThreshold`:  Defines when the scaler is active or not and scales from/to 0 based on it.
- `threshold`:  Defines the target value to scale the workload from 1 to _n_ instances and vice versa. To achieve this, KEDA passes the target value to the Horizontal Pod Autoscaler (HPA) and the built-in HPA controller will handle all the autoscaling.
```

JorTurFer: The problem is that we don't have specific values for that, they depend on the scaler, they can be queueLenght, threshold, value, lag... The constraint is that the activation value name will be the scaling value appending the prefix `activantion`, that's why I preferred to use these names

JorTurFer: But we can use `threshold` as the word for explaining the concept. WDYT?

zroubalik: yeah, let's keep the **Activation** and **Scaling** here

zroubalik (commented):
Looking good, just a couple of nits.

content/docs/2.8/concepts/scaling-deployments.md
JorTurFer: No no, don't worry, this will be the base for the change so it's better to discuss it and take the best possible approach.
I prefer to make this clearest as possible before merging it and starting with these changes

content/docs/2.8/concepts/scaling-deployments.md
```diff
@@ -226,6 +226,29 @@ metadata:
 
 The presensce of this annotation will pause autoscaling no matter what number of replicas is provided. The above annotation will scale your current workload to 0 replicas and pause autoscaling. You can set the value of replicas for an object to be paused at to any arbitary number. To enable autoscaling again, simply remove the annotation from the `ScaledObject`definition.
 
+### `Activating` and `Scaling`
+
+There are use-cases where the activating value (0-1 and 1-0) is totally different than 0, such as workloads scaled with the Prometheus scaler where the values go from -X to X. To give a consisten solution to this problem, KEDA has 2 different phases during the autoscaling process.
+
+- **Activation phase:**: The activating (or deactivating) phase is the moment when KEDA (operator) has to decide if the workload should be scaled from/to zero. KEDA is the responsible for this action based on the result of the scaler `IsActive` function and only applies in from/to 0 scaling.
```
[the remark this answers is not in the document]
JorTurFer: done

content/docs/2.8/concepts/scaling-deployments.md
```diff
@@ -226,6 +226,29 @@ metadata:
...
+- **Scaling phase:** The scaling phase is the moment when KEDA has decided to scale out to 1 instance and now is the HPA controller who takes the scaling decisions based on the configuration defined in the generated HPA (from ScaledObject data) and the metrics exposed by KEDA (metrics server). This phase applies the scaling from 1-N and N-1.
```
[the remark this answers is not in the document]
JorTurFer: done

content/docs/2.8/concepts/scaling-deployments.md
JorTurFer: To clarify this, I have made it bold the paragraph where this is explained. I think that now it's clearer

content/docs/2.8/concepts/scaling-deployments.md
```diff
@@ -226,6 +226,29 @@ metadata:
 
 The presensce of this annotation will pause autoscaling no matter what number of replicas is provided. The above annotation will scale your current workload to 0 replicas and pause autoscaling. You can set the value of replicas for an object to be paused at to any arbitary number. To enable autoscaling again, simply remove the annotation from the `ScaledObject`definition.
 
+### `Activating` and `Scaling`
+
+There are use-cases where the activating value (0-1 and 1-0) is totally different than 0, such as workloads scaled with the Prometheus scaler where the values go from -X to X. To give a consisten solution to this problem, KEDA has 2 different phases during the autoscaling process.
```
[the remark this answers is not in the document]
JorTurFer: I think that done

content/docs/2.8/concepts/scaling-deployments.md
```diff
@@ -232,6 +232,29 @@ metadata:
 
 The presensce of this annotation will pause autoscaling no matter what number of replicas is provided. The above annotation will scale your current workload to 0 replicas and pause autoscaling. You can set the value of replicas for an object to be paused at to any arbitary number. To enable autoscaling again, simply remove the annotation from the `ScaledObject`definition.
 
+### `Activating` and `Scaling`
+
+To give a consisten solution to this problem, KEDA has 2 different phases during the autoscaling process.
+
+- **Activation phase:**: The activating (or deactivating) phase is the moment when KEDA (operator) has to decide if the workload should be scaled from/to zero. KEDA is the responsible for this action based on the result of the scaler`IsActive` function and only applies to 0<->1 scaling. There are use-cases where the activating value (0-1 and 1-0) is totally different than 0, such as workloads scaled with the Prometheus scaler where the values go from -X to X.
```
tomkerkhove, line 239:
```suggestion
- **Activation phase:** The activating (or deactivating) phase is the moment when KEDA (operator) has to decide if the workload should be scaled from/to zero. KEDA is the responsible for this action based on the result of the scaler`IsActive` function and only applies to 0<->1 scaling. There are use-cases where the activating value (0-1 and 1-0) is totally different than 0, such as workloads scaled with the Prometheus scaler where the values go from -X to X.
```

content/docs/2.8/concepts/scaling-deployments.md
```diff
@@ -232,6 +232,29 @@ metadata:
 
 The presensce of this annotation will pause autoscaling no matter what number of replicas is provided. The above annotation will scale your current workload to 0 replicas and pause autoscaling. You can set the value of replicas for an object to be paused at to any arbitary number. To enable autoscaling again, simply remove the annotation from the `ScaledObject`definition.
 
+### `Activating` and `Scaling`
```
tomkerkhove, line 235: This is still not renamed as suggested here: https://github.com/kedacore/keda-docs/pull/780#discussion_r899774978

I think it would be better to revert to your original name given it's more approachable to new people and less technical focussed.

It also feels a bit odd to see this in conceptual docs:
image

content/docs/2.8/concepts/scaling-deployments.md
```diff
@@ -232,6 +232,29 @@ metadata:
...
+#### Managing Activation & Scaling Thresholds
+
+KEDA allows you to specify different values for each scenario:
+
+- **Activation:** Defines when the scaler is active or not and scales from/to 0 based on it.
+- **Scaling:** Defines the target value to scale the workload from 1 to _n_ instances and vice versa. To achieve this, KEDA passes the target value to the Horizontal Pod Autoscaler (HPA) and the built-in HPA controller will handle all the autoscaling.
+
+> ⚠️ **NOTE:** If the minimum replicas is >= 1, the scaler is always active and the activation value will be ignored.
+
+**Each scaler defines parameters for their use-cases, but the activation will always be the same as the scaling value, appended by the prefix `activation`** (ie: `threshold` for scaling and `activationThreshold` for activation).
```
tomkerkhove, line 251: I don't think we need to put this in bold as it looks odd in the rendered docs IMO; but who am I!
```suggestion
Each scaler defines parameters for their use-cases, but the activation will always be the same as the scaling value, appended by the prefix `activation` (ie: `threshold` for scaling and `activationThreshold` for activation).
```

tomkerkhove (approved):

content/docs/2.8/concepts/scaling-deployments.md
JorTurFer: I didn't see it, omw

content/docs/2.8/concepts/scaling-deployments.md
JorTurFer: renamed it again to `Activating and Scaling thresholds`

content/docs/2.8/concepts/scaling-deployments.md
tomkerkhove: No worries at all - That's what reviews are for - Thanks!

content/docs/2.8/concepts/scaling-deployments.md
JorTurFer: addressed

content/docs/2.8/concepts/scaling-deployments.md
JorTurFer: updated

content/docs/2.8/concepts/scaling-deployments.md
JorTurFer: reverted

PR #783, opened 2022-06-14
doc: Add a note on scalers which support float target value

JorTurFer:
Signed-off-by: Jorge Turrado <jorge_turrado@hotmail.es>

After merging the PR to support `mili` unit in core, there are some scalers that support `float` as target value (because the scaler logic already supported float64 and it made sense). 

### Checklist

Related https://github.com/kedacore/keda/issues/3135

### DIFF
--- a/content/docs/2.8/scalers/aws-cloudwatch.md
+++ b/content/docs/2.8/scalers/aws-cloudwatch.md
@@ -24,8 +24,8 @@ triggers:
     # Optional: Expression query
     expression: SELECT MAX("ApproximateNumberOfMessagesVisible") FROM "AWS/SQS" WHERE QueueName = 'keda'
     metricName: ApproximateNumberOfMessagesVisible
-    targetMetricValue: "2"
-    minMetricValue: "0"
+    targetMetricValue: "2.1"
+    minMetricValue: "1.5"
     # Required: region
     awsRegion: "eu-west-1"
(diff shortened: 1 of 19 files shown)


tomkerkhove (commented):
As an end-user, how much do I care about this? If I do, might need some more background on what this means maybe? Scaler typically does the heavy lifting on this and in some scalers this does not make sense - For example, I won't be scaling for 30.5 messages.

Maybe it's better to update the scalers where this is useful?

JorTurFer:
> As an end-user, how much do I care about this? If I do, might need some more background on what this means maybe? Scaler typically does the heavy lifting on this and in some scalers this does not make sense - For example, I won't be scaling for 30.5 messages.
> 
> Maybe it's better to update the scalers where this is useful?

you are right, on it

tomkerkhove:
They do support it, but does it make sense? :D Anyway it's up to them but for SB it's less relevant

tomkerkhove (approved):

JorTurFer:
Yes, the scalers inclued in this PR support float because it's relevant.
I mean, you can get a rate from 0 to 1 using Azure Monitor, or database scalers, etc.
The only case where could not be relevant is RabbitMQ using queues, but there is a remark explaining that this only applies in case of message rate

JorTurFer:
@zroubalik , does this match with your expectations too?

zroubalik (approved):
Looking good, thanks for this!

PR #790, opened 2022-06-21
docs: Permission segregation for Azure AD Pod & Workload Identity.

v-shenoy:
Signed-off-by: Vighnesh Shenoy <vighneshq@gmail.com>

https://github.com/kedacore/keda/pull/3216

### Checklist

Relates to https://github.com/kedacore/keda/issues/2656

### DIFF
--- a/content/docs/2.8/concepts/authentication.md
+++ b/content/docs/2.8/concepts/authentication.md
@@ -93,7 +93,7 @@ metadata:
 spec:
   podIdentity:
       provider: none | azure | azure-workload | aws-eks | aws-kiam  # Optional. Default: none
-      identityID: <identity-id>                         # Optional. Only used by azure & azure-workload providers.
+      identityId: <identity-id>                         # Optional. Only used by azure & azure-workload providers.
   secretTargetRef:                                      # Optional.
   - parameter: {scaledObject-parameter-name}            # Required.
     name: {secret-name}                                 # Required.
@@ -272,7 +272,7 @@ Currently we support the following:
 ```yaml
 podIdentity:
   provider: none | azure | azure-workload | aws-eks | aws-kiam  # Optional. Default: none
-  identityID: <identity-id>                                     # Optional. Only used by azure & azure-workload providers.
+  identityId: <identity-id>                                     # Optional. Only used by azure & azure-workload providers.
 ```
 
 #### Azure Pod Identity
@@ -284,13 +284,13 @@ You can tell KEDA to use Azure AD Pod Identity via `podIdentity.provider`.
 ```yaml
 podIdentity:
   provider: azure           # Optional. Default: none
-  identityID: <identity-id> # Optional. Default: Identity linked with the label set when installing KEDA. Refer to Pod Identity docs, and the docs below.
+  identityId: <identity-id> # Optional. Default: Identity linked with the label set when installing KEDA.
 ```
 
 Azure AD Pod Identity will give access to containers with a defined label for `aadpodidbinding`.  You can set this label on the KEDA operator deployment.  This can be done for you during deployment with Helm with `--set podIdentity.activeDirectory.identity={your-label-name}`.
 
-You can override the identity set while installing KEDA by specifying an `identityID` in the `podIdentity` field. This can be used to prevent an Azure Managed Identity from having permissions to multiple Azure resources, due to security
-concerns.
+You can override the identity that was assigned to KEDA during installation, by specifying an `identityId` parameter under
+the `podIdentity` field. This allows end-users to use different identities to access various resources which is more secure than using a single identity that has access to multiple resources.
 
 #### Azure Workload Identity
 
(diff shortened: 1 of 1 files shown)


content/docs/2.8/concepts/authentication.md
```diff
@@ -93,6 +93,7 @@ metadata:
 spec:
   podIdentity:
       provider: none | azure | azure-workload | aws-eks | aws-kiam  # Optional. Default: none
+      identityID: <identity-id>                         # Optional. Only used by azure & azure-workload providers.
```
tomkerkhove, line 96: Nit, we typically use `Id` instead of `ID`
```suggestion
      identityId: <identity-id>                         # Optional. Only used by azure & azure-workload providers.
```

tomkerkhove: Same for other places, obviously.

content/docs/2.8/concepts/authentication.md
```diff
@@ -281,11 +283,15 @@ You can tell KEDA to use Azure AD Pod Identity via `podIdentity.provider`.
 
 ```yaml
 podIdentity:
-  provider: azure # Optional. Default: none
+  provider: azure           # Optional. Default: none
+  identityID: <identity-id> # Optional. Default: Identity linked with the label set when installing KEDA. Refer to Pod Identity docs, and the docs below.
```
tomkerkhove, line 287: This is already the pod identity section so would leave this part out
```suggestion
  identityID: <identity-id> # Optional. Default: Identity linked with the label set when installing KEDA.
```

content/docs/2.8/concepts/authentication.md
```diff
@@ -281,11 +283,15 @@ You can tell KEDA to use Azure AD Pod Identity via `podIdentity.provider`.
...
 ```
 
 Azure AD Pod Identity will give access to containers with a defined label for `aadpodidbinding`.  You can set this label on the KEDA operator deployment.  This can be done for you during deployment with Helm with `--set podIdentity.activeDirectory.identity={your-label-name}`.
 
+You can override the identity set while installing KEDA by specifying an `identityID` in the `podIdentity` field. This can be used to prevent an Azure Managed Identity from having permissions to multiple Azure resources, due to security
+concerns.
```
tomkerkhove, line 293: How about this?
```suggestion
You can override the identity that was assigned to KEDA while installing it by specifying an `identityID` in the `podIdentity` field. This allows end-users to use different identities that have access to various resources which is more secure compared to only using one identity that has access to multiple Azure resources.
```

content/docs/2.8/concepts/authentication.md
```diff
@@ -307,6 +314,9 @@ following flags -
 2. `--set podIdentity.azureWorkload.clientId={azure-ad-client-id}`
 3. `--set podIdentity.azureWorkload.tenantId={azure-ad-tenant-id}`
 
+You can override the clientId set while installing KEDA by specifying an `identityID` in the `podIdentity` field. This can be used to prevent an Azure AD Application from having permissions to multiple Azure resources, due to security
```
tomkerkhove, line 317: See comment above

tomkerkhove (commented):
LGTM, just a few suggestions + tweaks

content/docs/2.8/concepts/authentication.md
```diff
@@ -284,13 +284,13 @@ You can tell KEDA to use Azure AD Pod Identity via `podIdentity.provider`.
...
-  identityID: <identity-id> # Optional. Default: Identity linked with the label set when installing KEDA. Refer to Pod Identity docs, and the docs below.
+  identityId: <identity-id> # Optional. Default: Identity linked with the label set when installing KEDA.
 ```
 
 Azure AD Pod Identity will give access to containers with a defined label for `aadpodidbinding`.  You can set this label on the KEDA operator deployment.  This can be done for you during deployment with Helm with `--set podIdentity.activeDirectory.identity={your-label-name}`.
 
-You can override the identity set while installing KEDA by specifying an `identityID` in the `podIdentity` field. This can be used to prevent an Azure Managed Identity from having permissions to multiple Azure resources, due to security
-concerns.
+You can override the identity that was assigned to KEDA during installation, by specifying an `identityId` parameter under
+the `podIdentity` field. This allows end-users to use different identities to access various resources which is more secure than using a single identity that has access to multiple resources.
```
tomkerkhove, line 293: No need to break lines.
 ```suggestion
You can override the identity that was assigned to KEDA during installation, by specifying an `identityId` parameter under the `podIdentity` field. This allows end-users to use different identities to access various resources which is more secure than using a single identity that has access to multiple resources.
```

File: content/docs/2.8/concepts/authentication.md [hunk 6]
```diff
@@ -314,8 +314,8 @@ following flags -
 2. `--set podIdentity.azureWorkload.clientId={azure-ad-client-id}`
 3. `--set podIdentity.azureWorkload.tenantId={azure-ad-tenant-id}`
 
-You can override the clientId set while installing KEDA by specifying an `identityID` in the `podIdentity` field. This can be used to prevent an Azure AD Application from having permissions to multiple Azure resources, due to security
-concerns.
+You can override the identity that was assigned to KEDA during installation, by specifying an `identityId` parameter under
+the `podIdentity` field. This allows end-users to use different identities to access various resources which is more secure than using a single identity that has access to multiple resources.
```
tomkerkhove, line 318: No need to break lines.
 ```suggestion
You can override the identity that was assigned to KEDA during installation, by specifying an `identityId` parameter under the `podIdentity` field. This allows end-users to use different identities to access various resources which is more secure than using a single identity that has access to multiple resources.
```

tomkerkhove (approved):

zroubalik (approved):

JorTurFer (approved):

PR #797, opened 2022-06-29
Update cpu.md documentation page

arnoldyahad:
Signed-off-by: Arnold Yahad <arnoldyahad@gmail.com>

### Checklist

Fixes #

Hey guys,
I have a suggestion for something in the CPU docs https://keda.sh/docs/2.0/scalers/cpu
I think the way it looks here:
image
everything in this snippet (which is white) can be copy-pasted
if something can't be copy pasted should be a comment in my opinion 

or like this:
image
just with `{AverageValue or Utilization}`

### DIFF
--- a/content/docs/2.0/scalers/cpu.md
+++ b/content/docs/2.0/scalers/cpu.md
@@ -20,7 +20,7 @@ triggers:
 - type: cpu
   metadata:
     # Required
-    type: Utilization/ AverageValue
+    type: Utilization # or AverageValue
     value: "60"
 ```
 


content/docs/2.0/scalers/cpu.md
```diff
@@ -20,7 +20,7 @@ triggers:
 - type: cpu
   metadata:
     # Required
-    type: Utilization/ AverageValue
+    type: Utilization # or AverageValue
```
tomkerkhove, line 23:
```suggestion
    type: Utilization # Allowed types are 'Utilization' or 'AverageValue'
```

tomkerkhove (changes requested):
Fine by me, but we should update all versions - Can you do that?

content/docs/2.0/scalers/cpu.md
byronic: I've made this change as part of a separate PR, in case that helps: https://github.com/kedacore/keda-docs/pull/909/files

tomkerkhove:
Abandoning given https://github.com/kedacore/keda-docs/pull/909 fixes this.

Thanks anyway @arnoldyahad!

Issue #412, opened 2021-03-31
Improve scaler catalog to be more user-friendly


tomkerkhove:
Over time, our scaler catalog has grown a lot over the last couple of releases where we have more than 30 of them today.

As a user, it's not that easy anymore to find the scalers you need and the navigation is not able to coop with it.

It would be good if we can:
- Make the navigation be more user-friendly in this case
- Integrate external scalers from the community to create a central place for all scalers (built-in or 3rd party)
- Provide filtering on the current overview, examples are:
  - Categories (ie messaging, databases, CI/CD)
  - Cloud (Alibaba, AWS, Azure, etc.)
  - Built-in / External
  - Maintainer (ie KEDA, community, ...)
  - Support for (cluster) trigger authentication?

chalin:
Do you consider this done or work-in-progress @tomkerkhove?
Should it be assigned to @thisisobate?

/cc @nate-double-u

tomkerkhove:
No because the filtering capabilities are not there yet.

I'm OK with assigning it to @thisisobate but I'm leaving it up to your team as I don't want to push work on your team so feel free to assign how you want to.

tomkerkhove:
And https://github.com/kedacore/keda-docs/issues/845 & https://github.com/kedacore/keda-docs/issues/857 are also related to this one

thisisobate:
Do we consider this as done? @tomkerkhove 
We've fixed all related issues so I think we should close this one

tomkerkhove:
Not really, we are still missing this
> * Provide filtering on the current overview, examples are:
>   
>   * Categories (ie messaging, databases, CI/CD)
>   * Cloud (Alibaba, AWS, Azure, etc.)
>   * ~Built-in / External~
>   * Maintainer (ie KEDA, community, ...)

thisisobate:
I originally thought the search bar already solves the filtering capabilities but it seems I'm wrong

tomkerkhove:
No, the idea is to have something similar to this:
![image](https://user-images.githubusercontent.com/4345663/190070506-0eaa6c4c-6d3f-4cbe-b7bc-ff426be2beca.png)

thisisobate:
@tomkerkhove I noticed that there are no "category" nor "cloud" label/tag in the scaler's metadata. This affects all the scaler docs for every version.

If you want to list these labels in the filter, then I would advise you open a PR that updates the docs metadata accordingly. For now, I will focus on listing just the "maintainer" label.

[...]

PR #805, opened 2022-07-07
feat: Provide scaler search and list external scalers from Artifact Hub

### LINKED ISSUE #577: Provide overview of all external scalers
Provide overview of all external scalers that are available.

This information should be based on Artifact Hub - https://artifacthub.io/packages/search?kind=8&sort=relevance&page=1

Kept is using a similar approach: https://keptn.sh/docs/integrations/

Relates to #412

thisisobate:
After analyzing the issue, I realized it's best to avoid creating an entirely new page for external scalers. Both built-in and external scalers fall under the same **scaler** category and thus, I think it's best to have both scaler types present on one landing page. By so doing, we maintain a clean information architecture that the user loves.

### Checklist

- [x] Create tabs for switching between built-in and external scalers
- [x] Embed widget to show external scaler hosted in artifacthub
- [x] Add search input
- [x] Style search input
- [x] Create search index using [lunr](https://lunrjs.com/)
- [x] Implement search functionality
- [x] Hide search result count when viewed in external scalers
- [x] Fix search result link for built in scalers
- [x] Show total scaler count by default
- [x] Filter by input change instead of submit
- [x] revert title to page name instead
- [x] Disable external scaler button when built in scaler is searched
- [x] Add a section for wildcard search in faq
- [x] clean up code

Fixes #577
Relates to #412


thisisobate:
@chalin @nate-double-u 
Any thoughts?

nate-double-u:
I like that you're splitting them up like that, the external page looks especially clean. Let's discuss if there is a way to break up the built-in set into categories.

tomkerkhove:
@kedacore/keda-maintainers Thoughts on this PR?

My current thinking is that I love it, but believe the initial request was to have a separate page for external scalers. In theory having them in the same with a filter is OK; but given their layout is different I'm doubting if we should go with that separate page.

With regards to the built-in scalers, though, we were recently discussing the way we display them on Scalers page vs our landing page and believe that the current Scalers page is more ideal given the amount of scalers that we have which makes it easier to see all of them at a glance without scrollign

JorTurFer:
I ❤️  how it looks, the only feedback I have it's: Could we make the scaler cards a bit smaller to introduce 3 columns per row? Now The scroll is huge. makes sense?

thisisobate:
@nate-double-u Splitting built-in scalers wouldn't be bad at all but I think it's best we make it a last resort. Tbh, we can explore other easier approach like @JorTurFer's suggestion

thisisobate:
@tomkerkhove @JorTurFer I agree with you. The scroll will become too long as we add more scalers.
Would try the 3-columns approach instead

JorTurFer:
What do you propose as a category filter? I mean, maintainer it's not a useful search value. I think that a search filter by name it's a good option. WDYT?

thisisobate:
The category filter I'm talking about is something like a list with checkboxes showing all the scalers maintained by a specific organization. When a user clicks on one of the categories (e.g Microsoft), it shows all the scalers maintained by Microsoft.

A search filter by name isn't bad either. I think would work perfectly.

tomkerkhove:
Thanks for adding the search! However the box was a bit hard to find and didn't realize I had to press enter 😅

thisisobate:
Lol...I know
Added styling and UX to the todo list in the PR description above

thisisobate:
The search actually supports wildcard search. All you need to do is append asterisks (e.g "**active***") at the end of the search query and it returns all the appropriate matches.

You can read more [here](https://lunrjs.com/guides/searching.html#wildcards)

@v-shenoy

v-shenoy:
Ah, that makes sense. Just one thing though, as a user the natural assumption (at least I think it to be) is that search is fuzzy. 

So, maybe just a line mentioning the usage and linking to this doc could be added?

thisisobate:
Sure...would add something to the faq for future reference

thisisobate:
@JorTurFer @nate-double-u @v-shenoy @chalin 
Please feel free to take a look

zroubalik:
Scalers page keeps loading and I don't see any scalers in the list there:

Snímek obrazovky 2022-07-25 v 10 05 15

thisisobate:
oh wow...taking a look at this

tomkerkhove:
> The search actually supports wildcard search. All you need to do is append asterisks (e.g "**active***") at the end of the search query and it returns all the appropriate matches.

That's good to know, but can we remove this requirement so that we search for exact + as suffix/prefix? The UX for needing to think about wildcards is less ideal imo

tomkerkhove:
If you would be able to fetch this data from:
image

As raw data, would it be hard to align the formatting with the built-in scalers to have a unified UI?

thisisobate:
I tried running this on local and it works. I also reloaded the deploy page and it works as expected. This is a weird behavior but to be on the safe side, I would refactor the code to use the [fetch API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API) instead of the old [Ajax](https://tr.javascript.info/xmlhttprequest) method

thisisobate:
> That's good to know, but can we remove this requirement so that we search for exact + as suffix/prefix? The UX for needing to think about wildcards is less ideal IMO

What do you mean exactly? I assume you mean we should remove the **website search** section on the faq page since it's less ideal for a user to use it?

thisisobate:
> As raw data, would it be hard to align the formatting with the built-in scalers to have a unified UI?

It depends. From a UI standpoint, implementation is feasible. if we should proceed with this, then it would be a separate PR.

zroubalik:
> No I would definitely keep that but was more suggesting the scaler search to work without "*" need. So if I search for "active" it should automatically show "ActiveMQ scaler"

Yeah, this is a must from my POV.

thisisobate:
> No I would definitely keep that but was more suggesting the scaler search to work without "*" need. So if I search for "active" it should automatically show "ActiveMQ scaler"

I see. Given my little experience with Lunr, I don't know for sure if Lunr gives us the capability to turn off wildcard search.

But since we use Javascript, anything is possible 😊

We can solve this by performing these wildcard searches under the hood whenever the user inputs a query in the search.

Would push a fix right away

thisisobate:
Would go ahead to remove the **website search** section on the faq page since this behavior is already handled under the hood

tomkerkhove:
> I assume you mean we should remove the website search section on the faq page since it's less ideal for a user to use it?

I would keep this for sake of consistency imo

thisisobate:
@tomkerkhove @zroubalik Please feel free to take a look

zroubalik:
@thisisobate it took me more than a minute to load the list of scalers, when I opened the page. Not sure if it is a problem of the development preview or if it is something in the page itself 🤔

thisisobate:
That is very strange...it loads up pretty fast for me.
Screenshot 2022-07-26 at 12 34 05

zroubalik:
It was probably some transient problem (it didn't worked in Chrome and Firefox), but seems like it is working for me now :)

thisisobate:
What I'm looking to do in a follow-up PR is to download and save the lunr.js package locally within the website directory. By so doing, we don't have to make lots of network calls for the search to work. This will significantly improve the load time and prevent us from being affected by a breaking change

thisisobate:
Follow up issues:
- #839 
- #840

tegioz:
> > As raw data, would it be hard to align the formatting with the built-in scalers to have a unified UI?
> 
> It depends. From a UI standpoint, implementation is feasible. if we should proceed with this, then it would be a separate PR.

Hi @thisisobate 👋 

If you need any help interacting with the [Artifact Hub API](https://artifacthub.io/docs/api/) to get the external scalers data, please let @cynthia-sg and me know, we'd be happy to help 🙂

JorTurFer:
It seems that it's failing yet
image

This is the line in the script: 
```html
<script>window.addEventListener("DOMContentLoaded",()=>{"use strict";let a;const d=window.location.pathname,l=String(d).split("/"),c=l[2],h=document.getElementById("search"),s=document.getElementById("search-input"),e=document.querySelector(".is-search-result"),r=document.querySelector(".results"),n=document.getElementById("is-search-template");let t=s.value.trim(),i={};t||o(),s.addEventListener("input",function(a){a.preventDefault();const i=s.value.trim(),r=document.querySelectorAll(".filterBtn");if(r.forEach(e=>e.disabled=!0),!i){const e=document.querySelectorAll(".filterBtn");e.forEach(e=>e.disabled=!1)}for(t=i.length>0?`${i}*`:"",o();e.firstChild.nextSibling;)e.removeChild(n.nextSibling.nextElementSibling)},!1);async function o(){const e=await fetch("/index.json",{method:"GET"}).then(e=>e.json()).then(e=>e).catch(e=>console.error("error:",e));a=lunr(function(){const t=e;this.ref("title"),this.field("title",{boost:20,usePipeline:!0,wildcard:lunr.Query.wildcard.TRAILING,presence:lunr.Query.presence.REQUIRED}),this.field("maintainer",{boost:15}),this.field("href",{boost:15}),this.field("description",{boost:10}),this.field("availability",{boost:5}),t.forEach(function(e){e.version===c&&(this.add(e),i[e.title]={href:e.href,title:e.title,maintainer:e.maintainer,description:e.description,availability:e.availability})},this)}),u(t)}function u(s){const t=a.search(s);if("content"in n){const o=document.createElement("h3");o.id="search-results",o.className="subtitle is-size-3",t.length==0?o.textContent=`No results found for "${s}"`:t.length==1?o.textContent=`Found one result for "${s}"`:t.length>1&&s===""?o.textContent=`${t.length} scalers available`:o.textContent=`Found ${t.length} results for "${s}"`,r.replaceChildren(o),t.forEach(function(o){const t=i[o.ref],s=n.content.cloneNode(!0);s.querySelector(".scaler-title").textContent=t.title,s.querySelector(".scaler-title").setAttribute("href",t.href),t.description&&(s.querySelector(".description").textContent=t.description),t.maintainer&&(s.querySelector(".maintainer").textContent=t.maintainer),t.availability&&(s.querySelector(".availability").textContent=t.availability),e.appendChild(s)},this)}}},!1)</script>

```

thisisobate:
Great catch! @JorTurFer 
Will fix up right away

JorTurFer:
I ❤️  how it looks!
Only one small thing (maybe it's intentional), the page looks smaller and more empty. I mean, The header size has decreased and also the "content column" is smaller, keeping a bigger empty size for the menu on the left side.
image

JorTurFer (approved):

layouts/partials/content.html
```diff
@@ -24,9 +24,27 @@
...
+        <button id="btn-focus-on-mount" class="button filterBtn is-focused" value="built-in-scalers">Built-in</button>
+        <button class="button filterBtn" value="external-scalers">External</button>
+      </div>
+      <div class="results"></div>
+      <div id="built-in-scalers" style="margin-top:20px;" class="columns is-multiline is-search-result">
+        <template id="is-search-template">
+        {{ .Render "scaler-template" }}
+        </template>
+      </div>
+      <div id="external-scalers" class="artifacthub-widget-group" style="display: none;" data-url="https://artifacthub.io/packages/search?kind=8&sort=relevance&page=1" data-theme="light" data-header="false" data-stars="true" data-color="#405273" data-responsive="true"  data-loading="true"></div>
```
tomkerkhove, line 44: We should add a section here that explains that external scaler information is pulled from Artifact Hub.

External scaler maintainers can learn more on https://artifacthub.io/docs/topics/repositories/#keda-scalers-repositories

tomkerkhove: Reactivating this as this is just a comment but the idea was to show this to our end-users

thisisobate: Do you mean something like a notice? @tomkerkhove

tomkerkhove: Just a sentence would be enough; just like the introduction.

thisisobate: If so, I will suggest we create this as a separate issue as this PR is growing out of its original scope. @tomkerkhove

thisisobate: > Just a sentence would be enough; just like the introduction.
Oh...I see. Alright then, will push something

tomkerkhove:
Is there also a way to reduce the spacing here? (not new issue though, so separate PR is ok)

image

thisisobate:
> Is there also a way to reduce the spacing here? (not new issue though, so separate PR is ok)

Yes...it's ok to have this as a separate PR

tomkerkhove (approved):

tomkerkhove:
Thanks a ton!

tomkerkhove:
I'll revert this PR because it also messed up our community page - Sorry!

https://keda.sh/community/#users

Before it was like this:
Screenshot_1

thisisobate:
> it also messed up our community page

Was able to trace this issue. would add a fix to this branch and open as a separate PR

tomkerkhove:
Thanks! It's also OK to do a new PR as I didn't complete the revert yet

thisisobate:
@tomkerkhove My suggestion: I think it's best to open separate PRs for each issue you highlighted instead of reverting this PR. I already created a fix for each so it's safe to have them as follow-up PRs. Wdyt?

tomkerkhove:
👌

tomkerkhove:
> Turns out there is a problem because the Redis scalers are not shown: image
> 
> Also, can we remove the "*" from the message given this is internal details.

@thisisobate Just FYI that the main problem here is still open though. If you type in "Redis"/"Azure"/"AWS" it does not show any of our scalers.

thisisobate:
Yeah...would take a look at it today

layouts/_default/scaler-template.html
```diff
@@ -0,0 +1,26 @@
...
+          class="has-text-weight-bold availability"
+        ></span>
+      </div>
+
+      <div class="card-footer-item card-is-wrapped">
+        Maintainer:&nbsp;<span class="has-text-weight-bold maintainer"></span>
+      </div>
+    </footer>
+  </div>
+</div>
```
chalin, line 26: This shouldn't be under `_default`. Did you try putting it under `partials`?

layouts/partials/content.html
```diff
@@ -24,9 +24,30 @@
...
+      <div class="results"></div>
+      <div id="built-in-scalers" style="margin-top:20px;" class="columns is-multiline is-search-result">
+        <template id="is-search-template">
+        {{ .Render "scaler-template" }}
+        </template>
+      </div>
+      <div id="external-scalers" style="display: none;">
+        <p>External scaler information is pulled from Artifact Hub. External scaler maintainers can learn more on <a href="https://artifacthub.io/docs/topics/repositories/#keda-scalers-repositories">Artifact Hub docs</a></p>
+        <div class="artifacthub-widget-group" data-url="https://artifacthub.io/packages/search?kind=8&sort=relevance&page=1" data-theme="light" data-header="false" data-stars="true" data-color="#405273" data-responsive="true"  data-loading="true"></div>
+        <script async src="https://artifacthub.io/artifacthub-widget.js"></script>
```
chalin, line 47: This would be better placed into a partial and included via the `partial` command as was done before.

config.toml
```diff
@@ -56,7 +56,18 @@ mediaType = "text/netlify"
 baseName = "_redirects"
 
 [outputs]
-home = [ "HTML", "REDIRECTS" ]
+home = [ "HTML", "RSS", "REDIRECTS", "Lunr" ]
```
chalin, line 59: Why activate "RSS"?

chalin (commented):
See inline comments for questions and remarks. I know it's late, but better (a bit) late than never :)

config.toml
tomkerkhove: For our blog? :)

chalin: Ok

layouts/_default/scaler-template.html
thisisobate: oh wow...just saw this now. 
Would try making it a partial instead. 
Thanks for bringing this to my attention!

chalin: No worries. Great job overall! Please cc me on your refactoring-into-partial PR. Thanks @thisisobate!

layouts/partials/javascript.html
```diff
@@ -1,24 +1,205 @@
...
+      let index;
+      const currentUrl = window.location.pathname;
+      const currentLocation = String(currentUrl).split("/");
+      const currentVersion = currentLocation[2];
+      const form = document.getElementById("search");
+      const input = document.getElementById("search-input");
+      const target = document.querySelector(".is-search-result");
+      const searchResultCount = document.querySelector(".results");
+      const template = document.getElementById("is-search-template");
+      let query = input.value.trim();
```
chalin, line 79: This line seems to be the cause of:

- #873

PR #863, opened 2022-08-03
chore: Clarification about activation

JorTurFer:

Some scalers like CPU/Memory, Cron, etc don't implement activation feature because it doesn't make sense for them. This PR adds a note clarifying that if the scaler page doesn't define the parameter for the scaler it is because the scaler doesn't support activation

### Checklist

Fixes #

### DIFF
--- a/content/docs/2.8/concepts/scaling-deployments.md
+++ b/content/docs/2.8/concepts/scaling-deployments.md
@@ -255,7 +255,7 @@ There are some important topics to take into account:
 - Opposite to scaling value, the activation value is always optional and the default value is 0.
 - The activation value has more priority than the scaling value in case of different decisions for each. ie: `threshold: 10` and `activationThreshold: 50`, in case of 40 messages the scaler is not active and it'll be scaled to zero even the HPA requires 4 instances.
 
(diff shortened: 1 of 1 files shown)


content/docs/2.8/concepts/scaling-deployments.md
```diff
@@ -255,6 +255,8 @@ There are some important topics to take into account:
 - Opposite to scaling value, the activation value is always optional and the default value is 0.
 - The activation value has more priority than the scaling value in case of different decisions for each. ie: `threshold: 10` and `activationThreshold: 50`, in case of 40 messages the scaler is not active and it'll be scaled to zero even the HPA requires 4 instances.
 
+> ⚠️ **NOTE:** If scaler's page doesn't define the specific parameter is because that specific scaler doesn't support activation value and is always 0.
```
zroubalik, line 258:
```suggestion
> ⚠️ **NOTE:** If a scaler doesn't define "activation" parameter (a property that starts with `activation` prefix), then this specific scaler doesn't support configurable activation value and the activation value is always 0.
```

zroubalik (approved):

tomkerkhove (approved):

Issue #869, opened 2022-08-04
Consider not listing scalers at the end of each individual scaler page


chalin:
Each individual scaler page should be about the scaler itself and nothing else. Why list all scalers at the end of the page? This adds content irrelevant to the scaler which shows up in, e.g., the printed version of the page.

chalin:
@tomkerkhove - any comments about this?

Related:

- #870

tomkerkhove:
I'd say this simplifies navigation but if they are listed in the side navigation as per https://github.com/kedacore/keda-docs/issues/860 then I'm ok with removing them

tomkerkhove:
PS: Add all of these issues to "Documentation Improvements" please :)

chalin:
> I'd say this simplifies navigation but if they are listed in the side navigation as per #860 then I'm ok with removing them

A user can click on **Scalers** in the top-nav (or the nav under mobile) to view all scalers. (I'd vote against #860, as I commented there, since the list of scalers is too long for a nav.)

tomkerkhove:
Too many clicks :) I should go straight from scaler A to B imo.

PR #871, opened 2022-08-04
Drop scalers-related dead code

chalin:
- This PR reduces technical debt by eliminating scaler-related code that is no longer relevant.
- Best viewed by ignoring whitespace changes
- No change in generated site files (other than differences in whitespace):
  ```console
  $ pwd
  .../keda-docs
  $ npm run check-links
  ...
  $ (cd public && git diff -bw --ignore-blank-lines)  
  $
  ```

Preview, see for example:

-

### DIFF
--- a/archetypes/scaler.md
+++ b/archetypes/scaler.md
@@ -3,7 +3,6 @@ title = "{{ replace .Name "-" " " | title }}"
 availability = ""
 maintainer = ""
 description = "Insert description here"
-layout = "scaler"
 +++
 
 ### Trigger Specification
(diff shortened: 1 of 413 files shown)


tomkerkhove:
This is probably related to the recent scaler catalog change I presume?

image

We should revert this and make sure it uses the previous formatting of simple buttons imo

chalin:
> We should revert this and make sure it uses the previous formatting of simple buttons imo

I'd vote instead for:

- #869

chalin:
@tomkerkhove et all: anything holding back the merging of this cleanup PR?

tomkerkhove:
But I forgot it's not related to this change, sorry

tomkerkhove (approved):

PR #899, opened 2022-08-16
chore: move scaler layout to partial directory instead

### LINKED ISSUE #892: Factor our scalers partial
_Originally posted by @chalin in https://github.com/kedacore/keda-docs/pull/805#discussion_r937771660_:

> This would be better placed into a partial and included via the `partial` command as was done before.

Specifically, lines 28 - 48 should be in a partial just like they were before the change (i.e., the old line 28):

> image

Hmm, I'm wondering if it would make sense to include the `h2` heading too? Actually, I think that it makes sense to move line 2 and from line 21 to the `{{ end }}` of the `{{ if }}`. WDYT?

thisisobate:

### Checklist

Fixes #892

### DIFF
--- a/layouts/partials/scaler-layout.html
+++ b/layouts/partials/scaler-layout.html
@@ -1,0 +1,58 @@
+<br />
+<br />
+
+<h2 id="available-scalers">Currently available scalers for KEDA</h2>
+<div class="buttons" style="display: flex; justify-content: flex-end">
+  <form
+    class="control has-icons-left has-extra-sm-right-margin"
+    id="search"
+    role="search"
+  >
+    <input
+      class="input is-rounded"
+      type="search"
+      placeholder="Search for scalers"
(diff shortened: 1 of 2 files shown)


thisisobate:
cc: @chalin

layouts/partials/scaler-layout.html
```diff
@@ -0,0 +1,58 @@
+<br />
```
tomkerkhove, line 1: Do we need these?

Think we can drop those:
image

thisisobate: yeah...we don't. But I think it's best I fix it in a separate PR since there is an open issue for it already

tomkerkhove: Ok, I forgot about the issue, thanks!

tomkerkhove (approved):

PR #917, opened 2022-08-28
feat: Support regex usage in Azure Service Bus scaler.

v-shenoy:
Signed-off-by: Vighnesh Shenoy <vshenoy@microsoft.com>

### Checklist

Relates to https://github.com/kedacore/keda/pull/3607

### DIFF
--- a/content/docs/2.9/scalers/azure-service-bus.md
+++ b/content/docs/2.9/scalers/azure-service-bus.md
@@ -41,8 +41,8 @@ triggers:
 - `subscriptionName` - Name of the Azure Service Bus queue to scale on. (Optional*, Required when `topicName` is specified)
 - `namespace` - Name of the Azure Service Bus namespace that contains your queue or topic. (Optional*, Required when pod identity is used)
 - `connectionFromEnv` - Name of the environment variable your deployment uses to get the connection string of the Azure Service Bus namespace. (Optional)
-- `useRegex` - This parameter allows the usage of regex (in the `queueName` or `subscriptionName` parameters). (Values: `true`, `false`, Default: `false`, Optional)
-- `operation` - Operation that will be used to compute the number of messages when `useRegex` is set to `true`. (Values: `sum`, `max`, or `avg`, Default: `sum`, Optional).
+- `useRegex` - Provides indication whether or not a regex is used in the `queueName` or `subscriptionName` parameters. (Values: `true`, `false`, Default: `false`, Optional)
+- `operation` - Defines how to compute the number of messages when `useRegex` is set to `true`. (Values: `sum`, `max`, or `avg`, Default: `sum`, Optional).
 - `cloud` - Name of the cloud environment that the service bus belongs to. Must be a known Azure cloud environment, or `Private` for Azure Stack Hub or Air Gapped clouds. (valid values: `AzurePublicCloud`, `AzureUSGovernmentCloud`, `AzureChinaCloud`, `AzureGermanCloud`, `Private`; default: `AzurePublicCloud`)
 
 When `cloud` is set to `Private`, the `endpointSuffix` parameter is required. Otherwise, it is automatically generated based on the cloud environment. `endpointSuffix` represents the service bus endpoint suffix of the cloud environment that the service bus belongs to, e.g. `servicebus.usgovcloudapi.net` for `AzureUSGovernmentCloud`.
--- a/content/docs/2.9/scalers/azure-service-bus.md
+++ b/content/docs/2.9/scalers/azure-service-bus.md
@@ -42,7 +42,7 @@ triggers:
 - `namespace` - Name of the Azure Service Bus namespace that contains your queue or topic. (Optional*, Required when pod identity is used)
 - `connectionFromEnv` - Name of the environment variable your deployment uses to get the connection string of the Azure Service Bus namespace. (Optional)
 - `useRegex` - This parameter allows the usage of regex (in the `queueName` or `subscriptionName` parameters). (Values: `boolean`, Default: `false`, Optional)
-- `operation` - Operation that will be used to compute the number of messages when `useRegex` is set to `true`. Required when `useRegex` is `true`. (Values: `sum`, `max`, or `avg`, Default: `sum`, Optional).
+- `operation` - Operation that will be used to compute the number of messages when `useRegex` is set to `true`. (Values: `sum`, `max`, or `avg`, Default: `sum`, Optional).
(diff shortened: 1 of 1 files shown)


content/docs/2.9/scalers/azure-service-bus.md
```diff
@@ -41,6 +41,8 @@ triggers:
 - `subscriptionName` - Name of the Azure Service Bus queue to scale on. (Optional*, Required when `topicName` is specified)
 - `namespace` - Name of the Azure Service Bus namespace that contains your queue or topic. (Optional*, Required when pod identity is used)
 - `connectionFromEnv` - Name of the environment variable your deployment uses to get the connection string of the Azure Service Bus namespace. (Optional)
+- `useRegex` - This parameter allows the usage of regex (in the `queueName` or `subscriptionName` parameters). (Values: `boolean`, Default: `false`, Optional)
+- `operation` - Operation that will be used to compute the number of messages when `useRegex` is set to `true`. Required when `useRegex` is `true`. (Values: `sum`, `max`, or `avg`, Default: `sum`, Optional).
```
tomkerkhove, line 45: why not just use entityNameRegex or so?

tomkerkhove: I think it's safe to assume that when the regex field was provided, that a regex should be used

v-shenoy: I aligned this with the way RabbitMQ supports regex, but what you're suggesting is also fair.

tomkerkhove: @JorTurFer @zroubalik wdyt?

content/docs/2.9/scalers/azure-service-bus.md
```diff
@@ -41,6 +41,8 @@ triggers:
 - `subscriptionName` - Name of the Azure Service Bus queue to scale on. (Optional*, Required when `topicName` is specified)
 - `namespace` - Name of the Azure Service Bus namespace that contains your queue or topic. (Optional*, Required when pod identity is used)
 - `connectionFromEnv` - Name of the environment variable your deployment uses to get the connection string of the Azure Service Bus namespace. (Optional)
+- `useRegex` - This parameter allows the usage of regex (in the `queueName` or `subscriptionName` parameters). (Values: `boolean`, Default: `false`, Optional)
```
JorTurFer, line 44: I'd say that in other places we have this
```suggestion
- `useRegex` - This parameter allows the usage of regex (in the `queueName` or `subscriptionName` parameters). (Values: `true`, `false`, Default: `false`, Optional)
```

content/docs/2.9/scalers/azure-service-bus.md
JorTurFer: it's true that in RabbitMQ we have `useRegex` as a boolean and `queueName` but there we don't have 2 different parameters (`queueName` and `topicName`) only one (`queueName`). Maybe in this context we can use `entityNameRegex` as single field for both.
BTW, isn't `topicName` instead of `subscriptionName` or I'm wrong?

v-shenoy: I didn't want to have a field entity name regex because then we would need another field to determine which entity it is (queue or subscription). 

Although, I guess if it's for a queue the metadata will only contain `entityNameRegex`, while if it's for a subscription, the metadata will contain `topicName` and `entityNameRegex`.

v-shenoy: The presence of a `queueName`, or `subscriptionName` already tells us what entity it is.

JorTurFer: are we going to apply the regex in the subscription name instead of the topic name?

v-shenoy: Yes, there can be multiple subscriptions within a topic, so this will make it easier to specify multiple of them.
And that was also the original ask - https://github.com/kedacore/keda/issues/1624

JorTurFer: oh, thanks for the clarification. Once again, the problem is that I'm a noobie with Azure Service Bus xD

content/docs/2.9/scalers/azure-service-bus.md
v-shenoy: I'll change this. Any other changes besides this?

tomkerkhove:
This PR looks a bit outdated with the code, or am I wrong?

v-shenoy:
> This PR looks a bit outdated with the code, or am I wrong?

What do you mean? If you mean regarding the changes @JorTurFer  is doing, yes. I will have to rebase and make changes once Jorge's commit with the new service bus package is merged.

tomkerkhove:
No, with regards to the changes for this feature as the code introduces these options:
image

But the docs only list 2 of them.

v-shenoy:
> No, with regards to the changes for this feature as the code introduces these options: image
> 
> But the docs only list 2 of them.

Oh no, `entityNameRegex` is just for internal use. Not for configuration.

content/docs/2.9/scalers/azure-service-bus.md
```diff
@@ -41,6 +41,8 @@ triggers:
 - `subscriptionName` - Name of the Azure Service Bus queue to scale on. (Optional*, Required when `topicName` is specified)
 - `namespace` - Name of the Azure Service Bus namespace that contains your queue or topic. (Optional*, Required when pod identity is used)
 - `connectionFromEnv` - Name of the environment variable your deployment uses to get the connection string of the Azure Service Bus namespace. (Optional)
+- `useRegex` - This parameter allows the usage of regex (in the `queueName` or `subscriptionName` parameters). (Values: `true`, `false`, Default: `false`, Optional)
```
tomkerkhove, line 44:
```suggestion
- `useRegex` - Provides indication whether or not a regex is used in the `queueName` or `subscriptionName` parameters. (Values: `true`, `false`, Default: `false`, Optional)
```

content/docs/2.9/scalers/azure-service-bus.md
```diff
@@ -41,6 +41,8 @@ triggers:
...
+- `operation` - Operation that will be used to compute the number of messages when `useRegex` is set to `true`. (Values: `sum`, `max`, or `avg`, Default: `sum`, Optional).
```
tomkerkhove, line 45:
```suggestion
- `operation` - Defines how to compute the number of messages across entities when `useRegex` is set to `true`. (Values: `sum`, `max`, or `avg`, Default: `sum`, Optional).
```

tomkerkhove (approved):
LGTM, but made some suggestions

PR #926, opened 2022-09-02
Update helm install steps

thepaulmacca:
Signed-off-by: Paul McDonald <pm@thepaulmacca.com>

Just a quick docs update to the `helm install` commands

### Checklist

### DIFF
--- a/content/docs/2.8/deploy.md
+++ b/content/docs/2.8/deploy.md
@@ -21,7 +21,7 @@ Deploying KEDA with Helm is very simple:
 1. Add Helm repo
 
     ```sh
-    helm repo add kedacore https://kedacore.github.io/charts
+    helm repo add keda https://kedacore.github.io/charts
     ```
 
 2. Update Helm repo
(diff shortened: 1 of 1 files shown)


JorTurFer:
hey!
thanks for the contribution! What would you like to achieve with this change? I mean, the organization is kedacore.

tomkerkhove:
I think both options are OK, but think kedacore is better to show that it's more than keda itself

JorTurFer:
Yes, I think KEDA is also correct, but as you say, the repo has more charts than KEDA, thanks for the clarification

thepaulmacca:
I just found it a bit confusing, as when I go to `Releases`, the commands are different

It's up to you both really

JorTurFer:
I'll review the releases then, maybe that's wrong. Thanks for pointing it

JorTurFer:
you are totally right. I'm going to solve it right now.
Thanks again ❤️

JorTurFer:
I have updated the already existing releases to show the correct commands and I have created another PR fixing [the release template](https://github.com/kedacore/charts/pull/314). I'll close this PR because it's not needed, but thanks again for pointing the issue ❤️❤️

PR #942, opened 2022-09-20
SASL/OAuthbearer Implementation Documentation

DevinGillman:
Added parameter documentation for  #3681
Fixed a spelling mistake.

Relates To #3681


zroubalik (commented):
Could you please add example ScaledObject and TA that uses `oauthbearer`, the same way it is done for other auth modes?

JorTurFer (approved):

zroubalik (approved):

PR #954, opened 2022-10-07
Document CRD totals Prometheus metrics exposed by the operator

v-shenoy:
Signed-off-by: Vighnesh Shenoy <vshenoy@microsoft.com>

### Checklist

Relates to https://github.com/kedacore/keda/pull/3719


content/docs/2.9/operate/prometheus.md
```diff
@@ -11,6 +11,7 @@ weight = 200
 The KEDA Operator exposes Prometheus metrics which can be scraped on port `8080` at `/metrics`. The following metrics are being gathered:
 
 - `keda_operator_trigger_totals` - Total number of triggers per trigger type.
+- `keda_operator_crd_totals` - Total number of KEDA custom resources per namespace for each resource.
```
JorTurFer, line 14:
```suggestion
- `keda_operator_crd_totals` - Total number of KEDA custom resources per namespace for each resource.
- `keda_operator_trigger_totals` - Total number of triggers per trigger type.
```
let's sort them alphabetically 😄

v-shenoy: Done.

JorTurFer (approved):
❤️

content/docs/2.9/operate/prometheus.md
```diff
@@ -10,6 +10,7 @@ weight = 200
 
 The KEDA Operator exposes Prometheus metrics which can be scraped on port `8080` at `/metrics`. The following metrics are being gathered:
 
+- `keda_operator_crd_totals` - Total number of KEDA custom resources per namespace for each resource.
```
tomkerkhove, line 13: Ok I'm late to the game but this metric name is odd @v-shenoy @JorTurFer because this refers to our CRD which means custom resource *definition*.

If we want to use something along that lines, we should use `keda_operator_resource_totals` or so IMO.

Also, is this ScaledObjects, ScaledJobs, TriggerAuthentication and/or ClusterTriggerAuthentication?

My suggestion would be to have these metrics IMO:
```suggestion
- `keda_operator_scaledobject_totals` - Total number of KEDA ScaledObjects per namespace resources.
- `keda_operator_scaledjobs_totals` - Total number of KEDA ScaledJobs per namespace resources.
- `keda_operator_triggerauthentication_totals` - Total number of KEDA TriggerAuthentication per namespace resources.
- `keda_operator_clustertriggerauthentication_totals` - Total number of KEDA ClusterTriggerAuthentication resources.
```

The current metric is not end-user friendly.

What do you think @zroubalik?

tomkerkhove (changes requested):

content/docs/2.9/operate/prometheus.md
v-shenoy: I thought having a single metric with the resource type as a dimension made sense. Do you think a cleaner name than `keda_operator_crd_totals` makes more sense, or do you think we need to have them separate?

tomkerkhove: I highly encourage to split them as we are shoving 4 scenarios in 1 metric for same of CX.

Thoughts on above proposal @zroubalik  @JorTurFer ?

JorTurFer: Based on my short experience with Prometheus, this multi dimension metrics are really common, so I'd keep them together, if you want to check only one type, it's enough adding the label in the query.

tomkerkhove: Yes, but to a point. If your metric cardinality is too big it will cause performance issues:
https://grafana.com/blog/2022/02/15/what-are-cardinality-spikes-and-why-do-they-matter/

Also I think the majority of times the use-case will be different and will only add noise by combining them while you can easily plot the individual metrics in one chart if you want all data.

If everyone wants this then I'll just resolve but curious to know what @zroubalik  thinks

JorTurFer: I get it, but I think we are in the low cardinality level (4 CRDs atm) and when we introduce more CRDs, we will introduce a few more, but I'd say that we will never have >40 CRDs (and I'm giving a huge safety margin). 

From plotting pov, I think that having multiple dimensions can improve the maintenance of the dashboards because end users can just group by CRD type and the charts will be updated automatically when new labels are added, something like `sum by (type) (keda_operator_crd_totals)`

But being honest, I don't have problems splitting them if you think that it's better

zroubalik: I tend to agree with what @JorTurFer points, although I am not completely against splitting, my vote would be on keeping the one metric.

tomkerkhove:
@v-shenoy can you elaborate on the labels that are provided on this metric next to resource type? For example, do we have namespace?

content/docs/2.9/operate/prometheus.md
tomkerkhove: Fine by me, seems like I'm the only one

v-shenoy:
> @v-shenoy can you elaborate on the labels that are provided on this metric next to resource type? For example, do we have namespace?

We have `type` and `namespace`

content/docs/2.9/operate/prometheus.md
tomkerkhove, line 13: The current phrasing makes it feel like it's per resource instead of CRD
```suggestion
- `keda_operator_crd_totals` - Total number of KEDA custom resources per namespace for each custom resource type (CRD).
```

tomkerkhove (approved):

content/docs/2.9/operate/prometheus.md
v-shenoy: Incorporated the suggestion. Also modified the metric name so that it's slightly more intuitive.

content/docs/2.9/operate/prometheus.md
```diff
@@ -10,6 +10,7 @@ weight = 200
 
 The KEDA Operator exposes Prometheus metrics which can be scraped on port `8080` at `/metrics`. The following metrics are being gathered:
 
+- `keda_operator_cr_totals` - Total number of KEDA custom resources per namespace for each custom resource type (CRD).
```
tomkerkhove, line 13: Here I am again to be a pain but `cr` is really a bad name.

Can we please rename it to `keda_operator_resource_totals` or `keda_operator_customresource_totals` please?
```suggestion
- `keda_operator_resources_totals` - Total number of KEDA custom resources per namespace for each custom resource type (CRD).
```

That way people know what it means when not looking at this page

tomkerkhove (changes requested):

zroubalik (approved):

tomkerkhove (approved):

PR #968, opened 2022-11-02
Small clarification to New Relic scaler docs

zanehala-wk:
Signed-off-by: zanehala-wk <zane.hala@workiva.com>

We've recently had some devs ask about the metric result when the `nodata` param is set to false and the query returns `nodata`. Looking in the scaler code it looks like the result will be `0`. I'm sure other people have wondered this as well so I figured it might be nice to call it out in the docs.

### Checklist


JorTurFer (approved):
Thanks for the clarification!

PR #977, opened 2022-11-11
Apache Pulsar: Document support for new auth modes

michaeljmarshall:
Document the new support for Bearer Token and Basic Auth Authentication methods.

### Checklist

Fixes https://github.com/kedacore/keda/issues/3844
Relates to https://github.com/kedacore/keda/pull/3845

### DIFF
--- a/content/docs/2.9/scalers/pulsar.md
+++ b/content/docs/2.9/scalers/pulsar.md
@@ -29,7 +29,7 @@ triggers:
 - `subscription` - Name of the topic subscription
 - `msgBacklogThreshold` - Average target value to trigger scaling actions. (default: 10)
 - `activationMsgBacklogThreshold` - Target value for activating the scaler. Learn more about activation [here](./../concepts/scaling-deployments.md#activating-and-scaling-thresholds).(Default: `0`, Optional)
-- `authModes` - a comma separated list of authentication modes to use. Default `""`. Valid options are `bearer`, `tls`, and `basic`. Note that `tls,bearer` or `tls,basic` are valid combinations and would indicate mutual TLS to secure the connection and then `bearer` or `basic` headers should be added to the HTTP request.
+- `authModes` - a comma separated list of authentication modes to use. (Values: `bearer`, `tls`,`basic`, Default: `""`, Optional, `tls,bearer` or `tls,basic` are valid combinations and would indicate mutual TLS to secure the connection and then `bearer` or `basic` headers should be added to the HTTP request)
 
 ### Authentication Parameters
 
--- a/content/docs/2.9/scalers/pulsar.md
+++ b/content/docs/2.9/scalers/pulsar.md
@@ -33,7 +33,7 @@ triggers:
 
 ### Authentication Parameters
 
-  The `authModes` are specified on the `metadata` field of the `ScaledObject`, as shown above. The associated configuration parameters are specified in the `TriggerAuthentication` spec.  
+The authentication is defined in `authModes`. The associated configuration parameters are specified in the `TriggerAuthentication` spec.  
 
 **Bearer Auth**
 
--- a/content/docs/2.9/scalers/pulsar.md
+++ b/content/docs/2.9/scalers/pulsar.md
@@ -19,6 +19,7 @@ triggers:
     subscription: sub1
     msgBacklogThreshold: '5'
     activationMsgBacklogThreshold: '2'
+    authModes: ""
 ```
 
 **Parameter list:**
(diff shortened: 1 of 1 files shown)


content/docs/2.9/scalers/pulsar.md
```diff
@@ -28,22 +29,42 @@ triggers:
 - `subscription` - Name of the topic subscription
 - `msgBacklogThreshold` - Average target value to trigger scaling actions. (default: 10)
 - `activationMsgBacklogThreshold` - Target value for activating the scaler. Learn more about activation [here](./../concepts/scaling-deployments.md#activating-and-scaling-thresholds).(Default: `0`, Optional)
+- `authModes` - a comma separated list of authentication modes to use. Default `""`. Valid options are `bearer`, `tls`, and `basic`. Note that `tls,bearer` or `tls,basic` are valid combinations and would indicate mutual TLS to secure the connection and then `bearer` or `basic` headers should be added to the HTTP request.
```
JorTurFer, line 32:
```suggestion
- `authModes` - a comma separated list of authentication modes to use. (Values: `bearer`, `tls`,`basic`, Default: `""`, Optional, `tls,bearer` or `tls,basic` are valid combinations and would indicate mutual TLS to secure the connection and then `bearer` or `basic` headers should be added to the HTTP request)
```

content/docs/2.9/scalers/pulsar.md
```diff
@@ -28,22 +29,42 @@ triggers:
...
 - `key`: Key for client authentication. 
 
-### Example
 
-Your pulsar cluster no TLS auth:
+### TLS with custom CA Certificates
+
+When configuring a trusted root CA that is not well known, it is sufficient to specify the `ca` field on the `TriggerAuthentication` resource. See [Bearer Token with TLS via custom trusted CA Certificate](#bearer-token-with-tls-via-custom-trusted-ca-certificate) for an example.
+
+Before 2.9.0, it was necessary to configure `tls: enable` on the `ScaledObject`. That was removed in a backwards compatible way, so you can remove that field now.
```
JorTurFer, line 63: I'm not sure if we should specify this here because this is 2.9 indeed, so changes are clear. We can/should add a deprecation message in the log, but I'm not sure if here is the best place.
@kedacore/keda-docs ?

zroubalik: I think that it is okay to keep it here, probably marked as 
```
> **Note:**
>
```

I am not sure if we want to have a deprecation message in the log? as the config will still work, it won't be changed in the future

JorTurFer: do we want to support the old `tls` approach having the new one?

zroubalik: @JorTurFer I am not sure I follow. @michaeljmarshall mentioned that the change is done in a backwards compatible way. Ie -> before it was needed to specify `tls: enable`, now you don't have to do so, but you can keep the setting there. No change in the functionality in the end.

JorTurFer: Maybe I misunderstood the change, but I supposed this will be the new "desired" way for the scaler. Am I right? @michaeljmarshall ?

JorTurFer: @kedacore/keda-docs , we should merge this soon because the feature was merged by error before having the docs ready.
Do you have any thoughts you'd like to share?

content/docs/2.9/scalers/pulsar.md
tomkerkhove: @JorTurFer Shouldn't the values/default be at the end?

content/docs/2.9/scalers/pulsar.md
```diff
@@ -28,22 +29,42 @@ triggers:
 - `subscription` - Name of the topic subscription
 - `msgBacklogThreshold` - Average target value to trigger scaling actions. (default: 10)
 - `activationMsgBacklogThreshold` - Target value for activating the scaler. Learn more about activation [here](./../concepts/scaling-deployments.md#activating-and-scaling-thresholds).(Default: `0`, Optional)
+- `authModes` - a comma separated list of authentication modes to use. Default `""`. Valid options are `bearer`, `tls`, and `basic`. Note that `tls,bearer` or `tls,basic` are valid combinations and would indicate mutual TLS to secure the connection and then `bearer` or `basic` headers should be added to the HTTP request.
 
 ### Authentication Parameters
 
-  If TLS is required you should set `tls` to `enable`. If required for your Pulsar configuration, you may also provide a `ca`, `cert` and `key`. `ca`, `cert` and `key` must be specified together.
+  The `authModes` are specified on the `metadata` field of the `ScaledObject`, as shown above. The associated configuration parameters are specified in the `TriggerAuthentication` spec.
```
tomkerkhove, line 36:
```suggestion
The authentication is defined in `authModes`. The associated configuration parameters are specified in the `TriggerAuthentication` spec.  
```

tomkerkhove (approved):
LGTM in addition to open items

content/docs/2.9/scalers/pulsar.md
JorTurFer: I have checked it, the style is 

- name - Description. (Values: x, y, z, Default: y, Optional, Extra Info)

So I think that my suggestion is correct

content/docs/2.9/scalers/pulsar.md
tomkerkhove: I don't see what the problem is with this PR to be honest.

The security fix by @michaeljmarshall seems unrelated to this and does not provide any configuration options anyway; but I might be missing something.

zroubalik (approved):

content/docs/2.9/scalers/pulsar.md
zroubalik: @JorTurFer @tomkerkhove so what about this?

tomkerkhove: Resolving

PR #982, opened 2022-11-22
support disable keep alive http connection

penghuazhou:
Signed-off-by: penghuazhou <1021556870@qq.com>

support disable keep alive http connection

### Checklist

Fixes #https://github.com/kedacore/keda/issues/3874

### DIFF
--- a/content/docs/2.9/operate/cluster.md
+++ b/content/docs/2.9/operate/cluster.md
@@ -75,6 +75,23 @@ $env:KEDA_HTTP_DEFAULT_TIMEOUT=1000
 
 All applicable scalers will use this timeout. Setting a per-scaler timeout is currently unsupported.
 
+## HTTP connection disable keep alive
+
+Keep alive behaviour is enabled by default for every HTTP connection, this could stack a huge amount of connections (one per scaler) in some scenarios. 
+
+You can override this default setting `KEDA_HTTP_DISABLE_KEEP_ALIVE` environment variable to true. For example, on Linux/Mac/Windows WSL2 operating systems, you'd use this command to set to true:
+
+```shell
+export KEDA_HTTP_DISABLE_KEEP_ALIVE=true
+```
+
+And on Windows Powershell, you'd use this command:
+
+```shell
+$env:KEDA_HTTP_DISABLE_KEEP_ALIVE=true
+```
+All applicable scalers will use this keep alive behaviour. Setting a per-scaler keep alive behaviour is currently unsupported.
+
 ## HTTP Proxies
 
 Some scalers issue HTTP requests to external servers (i.e. cloud services). As certain companies require external servers to be accessed by proxy servers, adding the relevant environment variables to both the KEDA Operator, and KEDA Metrics Server deployments (HTTP_PROXY, HTTPS_PROXY, NO_PROXY, etc.) would allow the scaler to connect via the desired proxy.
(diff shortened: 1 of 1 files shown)


content/docs/2.7/operate/cluster.md
```diff
@@ -65,6 +65,19 @@ $env:KEDA_HTTP_DEFAULT_TIMEOUT=1000
 
 All applicable scalers will use this timeout. Setting a per-scaler timeout is currently unsupported.
 
+## HTTP connection disable keep alive
+You can override this default by setting the `DISABLE-KEEP-ALIVES` environment variable to true. For example, on Linux/Mac/Windows WSL2 operating systems, you'd use this command to set to true:
```
JorTurFer, line 69:
```suggestion
## HTTP Keep alives

Keep alive behaviour is enabled by default for every HTTP connection, this could stack a huge amount of connections (one per scaler) in some scenarios. You can override this default setting `KEDA_HTTP_DISABLE_KEEP_ALIVE` environment variable to true. For example, on Linux/Mac/Windows WSL2 operating systems, you'd use this command to set to true:
```

content/docs/2.7/operate/cluster.md
```diff
@@ -65,6 +65,19 @@ $env:KEDA_HTTP_DEFAULT_TIMEOUT=1000
...
+```shell
+$env:DISABLE-KEEP-ALIVES=true
```
JorTurFer, line 71:
```suggestion
export KEDA_HTTP_DISABLE_KEEP_ALIVE=true
```

content/docs/2.7/operate/cluster.md
```diff
@@ -65,6 +65,19 @@ $env:KEDA_HTTP_DEFAULT_TIMEOUT=1000
...
+```
+
+And on Windows Powershell, you'd use this command:
+
+```shell
+$env:DISABLE-KEEP-ALIVES=true
```
JorTurFer, line 77:
```suggestion
$env:KEDA_HTTP_DISABLE_KEEP_ALIVE=true
```

JorTurFer (commented):
Please, don't change v2.7 file, this change will be applied in v2.9, move the changes there please 🙏 
When you have some time, please sign all the commits, it's required for us

content/docs/2.7/operate/cluster.md
penghuazhou: @JorTurFer, I have change it, please review again, thanks

zroubalik (approved):

content/docs/2.9/operate/cluster.md
```diff
@@ -75,6 +75,23 @@ $env:KEDA_HTTP_DEFAULT_TIMEOUT=1000
...
 All applicable scalers will use this timeout. Setting a per-scaler timeout is currently unsupported.
 
+## HTTP connection disable keep alive
+
+Keep alive behaviour is enabled by default for every HTTP connection, this could stack a huge amount of connections (one per scaler) in some scenarios. 
+
+You can override this default setting `KEDA_HTTP_DISABLE_KEEP_ALIVE` environment variable to true. For example, on Linux/Mac/Windows WSL2 operating systems, you'd use this command to set to true:
+
+```shell
+export KEDA_HTTP_DISABLE_KEEP_ALIVE=true
```
zroubalik, line 85: We should point users here to env variables in Kubernetes Deployments not on their local machines, similar way it is documented for `HTTP Proxies` section below. 

I see that you took the reference from `HTTP Timeouts` section, which is misleading and should be changed the same way, could you please change that as well?

zroubalik (changes requested):

content/docs/2.9/operate/cluster.md
penghuazhou: @JorTurFer, I have change it, please review again, thanks

JorTurFer (approved):

zroubalik (approved):

PR #999, opened 2022-12-06
Fix identation error on ScaledObject definition

1nvk3r:
The authenticationRef parameter was wrongly configured as a child of metadata and causing the following error: error validating data: ValidationError(ScaledObject.spec.triggers[0].metadata.authenticationRef): invalid type for sh.keda.v1alpha1.ScaledObject.spec.triggers.metadata: got "map", expected "string"; if you choose to ignore these errors, turn validation off with --validate=false.

This new identation should fix the problem.

Signed-off-by: AJPG <85803236+1nvk3r@users.noreply.github.com>

### Checklist

Fixes #

### DIFF
--- a/content/docs/2.8/scalers/new-relic.md
+++ b/content/docs/2.8/scalers/new-relic.md
@@ -99,6 +99,6 @@ spec:
         nrql: "SELECT average(duration) from Transaction where appName='SITE' TIMESERIES"
         noDataError: "true"
         threshold: 1000
-        authenticationRef:
-          name: keda-trigger-auth-new-relic
+      authenticationRef:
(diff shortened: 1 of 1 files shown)


JorTurFer (commented):
Nice catch! Thanks for the fix 🙇 
Could you update v2.9 with this same fix?

JorTurFer:
> No problem, bro. Just did it :)

Thanks, in the same PR would be enough, but I'll merge both. Thanks! 🙇

JorTurFer (approved):

1nvk3r:
Sorry about that @JorTurFer. I was multitasking and ended up doing things on auto.

PR #1027, opened 2022-12-30
docs: validating webhook

JorTurFer:
New validation hook to check if scale target is already managed

### Checklist

Relates to https://github.com/kedacore/keda/pull/4001
Relates to https://github.com/kedacore/keda/issues/3755
Relates to https://github.com/kedacore/keda/issues/3087
Relates to https://github.com/kedacore/keda/issues/3083
Relates to https://github.com/kedacore/keda-docs/issues/1030


tomkerkhove:
@JorTurFer Is this doc related to https://github.com/kedacore/keda/pull/4001#discussion_r1060006432?

tomkerkhove:
It looks like it

content/docs/2.10/concepts/_index.md
```diff
@@ -12,10 +12,11 @@ weight = 1
...
+2. **Metrics** — KEDA acts as a [Kubernetes metrics server](https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/#support-for-custom-metrics) that exposes rich event data like queue length or stream lag to the Horizontal Pod Autoscaler to drive scale out.  It is up to the Deployment to consume the events directly from the source.  This preserves rich event integration and enables gestures like completing or abandoning queue messages to work out of the box.  The metric serving is the primary role of the `keda-operator-metrics-apiserver` container that runs when you install KEDA.
+3. **Validation Webhook** - KEDA validates all the incoming ScaledObjects to ensure that the monitored workload isn't already autoscaled by other ScaledObject or HPA. This prevents unintended behaviours due to conficts between multple HPAs autoscaling the same workload.
```
tomkerkhove, line 19: I think it's best to align our naming since we currently use "Validation Webhook" and "Webhooks" in this PR; while the Kubernetes standard naming for this is Admission Controller.

I'd love to align with that since it's a known concept which will help resonate with people more

```suggestion
3. **Admission Controller** - Automatically validate resource changes to prevent misconfiguration and enforce best practices by using an [admission controller](https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/). As an example, it will prevent multiple ScaledObjects to target the same scale target.
```

content/docs/2.10/operate/_index.md
```diff
@@ -9,4 +9,5 @@ We provide guidance & requirements around various areas to operate KEDA:
 - Cluster ([link](./cluster))
 - Integrate with Prometheus ([link](./prometheus))
 - Kubernetes Events ([link](./events))
-- KEDA Metrics Server ([link](./metrics-server))
\ No newline at end of file
+- KEDA Metrics Server ([link](./metrics-server))
+- Validating Webhook ([link](./validating-webhook))
```
tomkerkhove, line 13:
```suggestion
- Admission Controller ([link](./admission-controller))
```

content/docs/2.10/operate/prometheus.md
```diff
@@ -18,6 +18,13 @@ The KEDA Operator exposes Prometheus metrics which can be scraped on port `8080`
 - `keda_trigger_totals` - Total number of triggers per trigger type.
 - Metrics exposed by the `Operator SDK` framework as explained [here](https://sdk.operatorframework.io/docs/building-operators/golang/advanced-topics/#metrics).
 
+### Webhooks
+
+The KEDA Webhooks expose Prometheus metrics which can be scraped on port `8080` at `/metrics`. The following metrics are being gathered:
+
+- `scaled_object_validating_total`- The current value for scaled object validations.
+- `scaled_object_validating_errors` - The number of validating errors.
```
tomkerkhove, line 26:
```suggestion
### Admission Controller

The KEDA Admission Controller expose Prometheus metrics which can be scraped on port `8080` at `/metrics`. The following metrics are being gathered:

- `scaled_object_validation_total`- The current value for scaled object validations.
- `scaled_object_validation_errors` - The number of validation errors.
```

content/docs/2.10/operate/validating-webhook.md
```diff
@@ -0,0 +1,16 @@
++++
+title = "Validating Webhook"
```
tomkerkhove, line 2:
```suggestion
title = "Admission Controller"
```

content/docs/2.10/operate/validating-webhook.md
```diff
@@ -0,0 +1,16 @@
...
+description = "Details on KEDA Validating Webhook"
```
tomkerkhove, line 3:
```suggestion
description = "Automatically validate resource changes to prevent misconfiguration and enforce best practices"
```

content/docs/2.10/operate/validating-webhook.md
```diff
@@ -0,0 +1,16 @@
...
+weight = 100
++++
+
+## Validating Webhooks
```
tomkerkhove, line 7:
```suggestion
## Admission Controller

> 💡 The Admission Controller is an opt-in feature and will become an opt-out feature as of KEDA v2.12.

```

content/docs/2.10/operate/validating-webhook.md
```diff
@@ -0,0 +1,16 @@
...
+
+Webhooks require a valid certificates to expose the server for the api-server. To achieve this KEDA generates it's own certificates and stores them in a secret. This solves the problem but some users could want to have their own certificates generated from their own CA in order to improve the security. This can be doable disabling the certificate rotation via setting the console argument `disable-cert-rotation=true` and placing the `tls.crt` and `tls.key` files (these names are required by the framework and can't be changed) in the path `/certs`. The path is also configurable using the console argument `webhooks-cert-dir=PATH-FOR-CERTS`.
```
tomkerkhove, line 9:
```suggestion
The KEDA Admission Controller requires a valid certificate to expose the server for the api-server. To achieve this KEDA generates its own certificates and stores them in a secret. 

While this is a good starting point, some end-users may want to use their own certificates which are generated from their own CA in order to improve security. This can be achieved by disabling the certificate rotation via setting the console argument `disable-cert-rotation=true` and mounting the `tls.crt` and `tls.key` files to `/certs`. The path is also configurable using the console argument `webhooks-cert-dir=PATH-FOR-CERTS`.
```

content/docs/2.10/operate/validating-webhook.md
```diff
@@ -0,0 +1,16 @@
...
+
+As default, the validating webhook is registered with `failurePolicy: Ignore`, this won't block the resources creation/update if the webhook isn't available. To ensure that the validation is always required, rejecting incoming ScaledObjects when the webhook isn't available, `failurePolicy: Fail` is required.
```
tomkerkhove, line 11:
```suggestion
By default, the admission controller is registered with `failurePolicy: Ignore`, this won't block the resources creation/update when the admission controller is not available. To ensure that the validation is always required and perform validation, setting `failurePolicy` to `Fail` is required.
```

content/docs/2.10/operate/validating-webhook.md
```diff
@@ -0,0 +1,16 @@
...
+
+### ScaledObject
```
tomkerkhove, line 13: Let's call them prevention rules as this will grow more in the future:
```suggestion
### Prevention Rules

#### ScaledObject
```

content/docs/2.10/operate/validating-webhook.md
```diff
@@ -0,0 +1,16 @@
...
+
+The KEDA registers a validating webhook for checking every incoming ScaledObject to ensure that them don't autoscale a workload already autoscaled by other sources (other ScaledObject or HPA), giving a message exlaining the conflict.
```
tomkerkhove, line 15:
```suggestion
KEDA will block all incoming changes to `ScaledObject` to ensure that they do not autoscale a workload that is already being autoscaled by another other sources (other ScaledObject or HPA).
```

tomkerkhove (changes requested):
I did a first skim.

I think it's a good start, but would:
1. Rename everything to "Admission Controller" which is a known concept
2. Restructure things a bit for upcoming rules that we already have on the backlog
3. Update the image to also use "Admission Controller"
4. Introduce a section under "Concepts" that talks about the feature on what it is and how it helps, while the "operate" section can be a listing of how to opt-in, configure certs and give detailed overview of prevention rules

JorTurFer:
> 4\. Introduce a section under "Concepts" that talks about the feature on what it is and how it helps, while the "operate" section can be a listing of how to opt-in, configure certs and give detailed overview of prevention rules

I'm not sure if this makes sense, I mean, the feature is too small and has a single responsibility, maybe in the future this takes more entity, but for the moment I wouldn't put at the same as those things

content/docs/2.10/operate/validating-webhook.md
JorTurFer: I'd modify this once we have more things, using an <h4> having a single element doesn't make sense IMO

content/docs/2.10/operate/validating-webhook.md
JorTurFer: Is an optional feature? Using helm we can make it optional, but with `make deploy` or using kubectl apply we can't do it because the manifests are the same. That's why I configure it to not fail in case of missing admission-controller, end users can't disable them just removing or scaling to zero the deployment

tomkerkhove: That's what we discussed on the issue; it should be an opt-in. Setting the validation to skip aligns with this here an we should make that clear or end-users will assume the validation blocks things

tomkerkhove:
> > 4. Introduce a section under "Concepts" that talks about the feature on what it is and how it helps, while the "operate" section can be a listing of how to opt-in, configure certs and give detailed overview of prevention rules
> 
> I'm not sure if this makes sense, I mean, the feature is too small and has a single responsibility, maybe in the future this takes more entity, but for the moment I wouldn't put at the same as those things

But yet it is a core component of KEDA so we should cover it imo

content/docs/2.10/operate/validating-webhook.md
JorTurFer: Update the installation steps to reflect both files :)

content/docs/2.10/operate/validating-webhook.md
JorTurFer: I have removed the `#### ScaledObject` in favor of a rule list because `<h4>` is too small

JorTurFer:
PTAL @tomkerkhove 
I have moved the page from operate to concepts. TBH, I don't have enough content to have a real page if I split into `concepts` and `operate`. 
For properly explaining how to use custom certs (in depth, not just the overview) I have created [this issue](https://github.com/kedacore/keda-docs/issues/1030) to track it because I have to change also the cert generation in the operator/ms to use the same approach as the webhooks and I prefer to do all at once (just to clarify, I'm going to do it after finishing the webhooks feature).

content/docs/2.10/concepts/admission-webhooks.md
```diff
@@ -0,0 +1,22 @@
++++
+title = "Admission Webhook"
+description = "Automatically validate resource changes to prevent misconfiguration and enforce best practices"
+weight = 100
```
tomkerkhove, line 4: Let's make sure it gets pushed down the list:
image

content/docs/2.10/concepts/admission-webhooks.md
```diff
@@ -0,0 +1,22 @@
...
++++
+
+## Admission Webhooks
+
+> 💡 The Admission Webhooks are an opt-in feature and will become an opt-out feature as of KEDA v2.12.
```
tomkerkhove, line 9: Let's remove this:
```suggestion
> 💡 The Admission Webhooks are an opt-in feature and will become an opt-out feature as of KEDA v2.12.
```

Because:
image

content/docs/2.10/concepts/_index.md
```diff
@@ -12,10 +12,11 @@ weight = 1
 
 ## How KEDA works
```
tomkerkhove, line 13: A nit on the new overview - All KEDA components are grouped, except for the  new one. IMO we should move it.

image

content/docs/2.10/concepts/admission-webhooks.md
```diff
@@ -0,0 +1,22 @@
...
+While this is a good starting point, some end-users may want to use their own certificates which are generated from their own CA in order to improve security. This can be achieved by disabling the certificate rotation via setting the console argument `enable-cert-rotation=false` (or removing it) and mounting the `tls.crt` and `tls.key` files to `/certs`. The path is also configurable using the console argument `webhooks-cert-dir=PATH-FOR-CERTS`.
+
+By default, the admission webhooks are registered with `failurePolicy: Ignore`, this won't block the resources creation/update when the admission controller is not available. To ensure that the validation is always required and perform validation, setting `failurePolicy` to `Fail` is required.
+
+### Prevention Rules
+
+KEDA will block all incoming changes to `ScaledObject` that don't match these rules:
+
+- The scaled workload (`scaledobject.spec.scaleTargetRef`) is already autoscaled by another other sources (other ScaledObject or HPA).
+- CPU and/or Memory trigger are used and the scaled workload doesn't have the requests defined. **This rule doesn't apply to all the workload types, only to `Deployment` and `StatefulSet`.**
```
tomkerkhove, line 22: This content is good, but in my opinion this is a mix of two types of docs:
- **Concepts** - What is it doing and how does it help me?
- **Configuration/Ops** - How do I manage certs.

I believe we should move the certs to a new "operate" page for "admission webhooks" or extend it to our existing "cluster" doc (although I think that one is not the best place and will burry the info.

I personally think all we need to have in this doc is:

```suggestion

The Admission Webhooks are used to automatically validate resource changes to prevent misconfiguration and enforce best practices.

This is achieved by intercepting and validating all resource changes based on a set of [prevention rules](#prevention-rules).

<<Visual with interaction between control plane, webhooks and eventually operator processing CRD if applicable>>

### Prevention Rules

KEDA will block all incoming changes to `ScaledObject` that don't match these rules:

- The scaled workload (`scaledobject.spec.scaleTargetRef`) is already autoscaled by another other sources (other ScaledObject or HPA).
- CPU and/or Memory trigger are used and the scaled workload doesn't have the requests defined. **This rule doesn't apply to all the workload types, only to `Deployment` and `StatefulSet`.**

File: content/docs/2.10/deploy.md [hunk 15]
```diff
@@ -81,16 +81,22 @@ Locate installed KEDA Operator in `keda` namespace, then remove created `KedaCon
 
 If you want to try KEDA on [Minikube](https://minikube.sigs.k8s.io) or a different Kubernetes deployment without using Helm you can still deploy it with `kubectl`.
 
-- We provide sample YAML declaration which includes our CRDs and all other resources in a file which is available on the [GitHub releases](https://github.com/kedacore/keda/releases) page.
+- We provide sample YAML declaration which includes our CRDs and all other resources in a file which is available on the [GitHub releases](https://github.com/kedacore/keda/releases) page. There are 2 files available, with and without the admission webhooks (keda-2.xx.x.yaml and keda-2.xx.x-core.yaml respectively), you can chose the installation mode just using the correct file.
```
tomkerkhove, line 84: This is a bit long
image

Let's simplify
```suggestion
- We provide sample YAML declaration which includes our CRDs and all other resources in a file which is available on the [GitHub releases](https://github.com/kedacore/keda/releases) page.
  - We offer two options to deploy KEDA:
    - Use `keda-2.xx.x.yaml` that includes all features, including admission webhooks (recommended)
    - Use `keda-2.xx.x-core.yaml` that installs the minimal required KEDA components, without admission webhooks
```

What do you think?

tomkerkhove (commented):
Looking very good, thank you!

Added a few suggestions, but I believe we should split the concept document and move some information to a new operate doc.

tomkerkhove:
> PTAL @tomkerkhove Tom Kerkhove FTE I have moved the page from operate to concepts. TBH, I don't have enough content to have a real page if I split into `concepts` and `operate`. For properly explaining how to use custom certs (in depth, not just the overview) I have created [this issue](https://github.com/kedacore/keda-docs/issues/1030) to track it because I have to change also the cert generation in the operator/ms to use the same approach as the webhooks and I prefer to do all at once (just to clarify, I'm going to do it after finishing the webhooks feature).

That's OK and I like it more this way. I have made a suggestion, however, to split the pages anyway as they are targeting different audiences but I hope my example explains it better - Happy to discuss over call if it helps.

content/docs/2.10/concepts/admission-webhooks.md
JorTurFer: Yeah, I didn't know what the `weight` means, but now I got it! I move it between External Scalers and Troubleshooting

content/docs/2.10/concepts/admission-webhooks.md
JorTurFer: nice!

content/docs/2.10/concepts/_index.md
JorTurFer: I placed out because conceptually it's different from other pieces, but I can update the graphic and place it at the same level.

JorTurFer: done!

content/docs/2.10/deploy.md
JorTurFer: nice!

content/docs/2.10/concepts/admission-webhooks.md
JorTurFer: I have split into 2 different pages, but I feel both like too short :( 
PTAL @tomkerkhove

content/docs/2.10/operate/tls-certificates.md
```diff
@@ -0,0 +1,13 @@
++++
+title = "Use your own TLS Certificates"
```
tomkerkhove, line 2: Let's call this "Security" with "Use your own TLS Certificates" as a section so we can add more topics in the future.

That way, this document is more of a catch-all for security topics and we don't have to break the URL in the future (assuming we also rename this file to `security.md`)

content/docs/2.10/deploy.md
```diff
@@ -81,16 +81,26 @@ Locate installed KEDA Operator in `keda` namespace, then remove created `KedaCon
...
+    - Use `keda-2.xx.x.yaml` that includes all features, including admission webhooks (recommended)
+    - Use `keda-2.xx.x-core.yaml` that installs the minimal required KEDA components, without admission webhooks
+
 Run the following command (if needed, replace the version, in this case `2.10.0`, with the one you are using):
 
 ```sh
+// Including admission webhooks
 kubectl apply -f https://github.com/kedacore/keda/releases/download/v2.10.0/keda-2.10.0.yaml
+// Without admission webhooks
+kubectl apply -f https://github.com/kedacore/keda/releases/download/v2.10.0/keda-2.10.0-core.yaml
```
tomkerkhove, line 95: Comments in shell are using # instead of // which breaks the syntax highlighting
```suggestion
# Including admission webhooks
kubectl apply -f https://github.com/kedacore/keda/releases/download/v2.10.0/keda-2.10.0.yaml
# Without admission webhooks
kubectl apply -f https://github.com/kedacore/keda/releases/download/v2.10.0/keda-2.10.0-core.yaml
```

content/docs/2.10/deploy.md
```diff
@@ -81,16 +81,26 @@ Locate installed KEDA Operator in `keda` namespace, then remove created `KedaCon
...
 ```
 
 - Alternatively you can download the file and deploy it from the local path:
 ```sh
```
tomkerkhove, line 99: Comments in shell are using # instead of // which breaks the syntax highlighting

content/docs/2.10/deploy.md
```diff
@@ -106,13 +116,19 @@ VERSION=2.10.0 make deploy
 - In case of installing from released YAML file just run the following command (if needed, replace the version, in this case `2.10.0`, with the one you are using):
 
 ```sh
```
tomkerkhove, line 118: Comments in shell are using # instead of // which breaks the syntax highlighting

content/docs/2.10/deploy.md
```diff
@@ -106,13 +116,19 @@ VERSION=2.10.0 make deploy
...
+// Including admission webhooks
 kubectl delete -f https://github.com/kedacore/keda/releases/download/v2.10.0/keda-2.10.0.yaml
+// Without admission webhooks
+kubectl delete -f https://github.com/kedacore/keda/releases/download/v2.10.0/keda-2.10.0-core.yaml
 ```
 
 - If you have downloaded the file locally, you can run:
 
 ```sh
```
tomkerkhove, line 127: Comments in shell are using # instead of // which breaks the syntax highlighting

content/docs/2.10/operate/admission-webhooks.md
```diff
@@ -0,0 +1,7 @@
++++
+title = "Admission Webhooks"
+description = "Admission webhooks configurations guidance"
+weight = 100
++++
+
+By default, the admission webhooks are registered with `failurePolicy: Ignore`, this won't block the resources creation/update when the admission controller is not available. To ensure that the validation is always required and perform validation, setting `failurePolicy` to `Fail` is required.
```
tomkerkhove, line 7: Just a suggestion to create a header for this section; as we'll add more content in the future and we can deep-link to it

```suggestion
## Validation Enforcement

By default, the admission webhooks are registered with `failurePolicy: Ignore`, this won't block the resources creation/update when the admission controller is not available. To ensure that the validation is always required and perform validation, setting `failurePolicy` to `Fail` is required.
```

tomkerkhove (approved):
LGTM, just added some small nits.

Please also update https://github.com/kedacore/keda-docs/blob/main/schematics.pptx which we should use in favor of the one in the code repo as it feels more related here than there. Thoughts? I would remove it from code repo as well.

content/docs/2.10/operate/tls-certificates.md
JorTurFer: I like the idea

JorTurFer:
Done! PTAL

content/docs/2.10/operate/security.md
```diff
@@ -0,0 +1,15 @@
...
+title = "Security"
+description = "Guidance to configure security options"
+weight = 100
++++
+
+## Use your own TLS Certificates
+
+KEDA uses self-signed certificates for different things. These certificates are generated and rotated by the operator, storing them in a Kubernetes secret (`kedaorg-certs`) that it's mounted in other KEDA components in the (default) path `/certs` using the names `tls.crt` and `tls.key` for TLS certificate and `ca.crt` and `ca.key` for CA certificate. KEDA also patches Kubernetes resources to include the `caBundle`, making Kubernetes to trust in the CA.
+
+While this is a good starting point, some end-users may want to use their own certificates which are generated from their own CA in order to improve security. This can be done disabling the certificate generation/rotation in the operator and updating default values in other components (if required).
```
zroubalik, line 11:
```suggestion
While this is a good starting point, some end-users may want to use their own certificates which are generated from their own CA in order to improve security. This can be done by disabling the certificate generation/rotation in the operator and updating default values in other components (if required). 
```

content/docs/2.10/concepts/admission-webhooks.md
tomkerkhove: LGTM

content/docs/2.10/operate/security.md
```diff
@@ -0,0 +1,15 @@
...
+
+KEDA uses self-signed certificates for different things. These certificates are generated and rotated by the operator, storing them in a Kubernetes secret (`kedaorg-certs`) that it's mounted in other KEDA components in the (default) path `/certs` using the names `tls.crt` and `tls.key` for TLS certificate and `ca.crt` and `ca.key` for CA certificate. KEDA also patches Kubernetes resources to include the `caBundle`, making Kubernetes to trust in the CA.
+
+While this is a good starting point, some end-users may want to use their own certificates which are generated from their own CA in order to improve security. This can be done disabling the certificate generation/rotation in the operator and updating default values in other components (if required). 
+
+As the operator is the responsible of generating the certificate for all the services, this behaviour can be disabled removing the console argument `--enable-cert-rotation=true` or setting it to `false`. Once this is disabled, user given certs can be placed in the secret `kedaorg-certs` which is automatically mounted in all the components or they can be patched to use other secret (this can be done throught helm values too).
```
zroubalik, line 13:
```suggestion
The KEDA operator is responsible for generating certificates for all the services, this behaviour can be disabled removing the console argument `--enable-cert-rotation=true` or setting it to `false`. Once this setting is disabled, user given certs can be placed in the secret `kedaorg-certs` which is automatically mounted in all the components or they can be patched to use other secret (this can be done throught helm values too).
```

`or they can be patched to use other secret (this can be done throught helm values too)` - there should be an example how to do that.

content/docs/2.10/operate/cluster.md
```diff
@@ -25,10 +25,11 @@ As a reference, this compatibility matrix shows supported k8s versions per KEDA
...
 
-| Deployment     | CPU                     | Memory                        |
-| -------------- | ----------------------- | ----------------------------- |
-| Metrics Server | Limit: 1, Request: 100m | Limit: 1000Mi, Request: 100Mi |
-| Operator       | Limit: 1, Request: 100m | Limit: 1000Mi, Request: 100Mi |
+| Deployment         | CPU                     | Memory                        |
+| ------------------ | ----------------------- | ----------------------------- |
+| Metrics Server     | Limit: 1, Request: 100m | Limit: 1000Mi, Request: 100Mi |
+| Operator           | Limit: 1, Request: 100m | Limit: 1000Mi, Request: 100Mi |
+| Admission Webhooks | Limit: 1, Request: 100m | Limit: 1000Mi, Request: 100Mi |
```
tomkerkhove, line 32: Shouldn't we alphabetically sort?

tomkerkhove (approved):
LGTM, feel free to merge when feature is merged

tomkerkhove:
> Please also update [`main`/schematics.pptx](https://github.com/kedacore/keda-docs/blob/main/schematics.pptx?rgh-link-date=2023-01-09T06%3A46%3A27Z) which we should use in favor of the one in the code repo as it feels more related here than there. Thoughts? I would remove it from code repo as well.

@JorTurFer I think you did it the other way around :D I would keep the file here and remove it from kedacore/keda (code repo) since it is unlikely that we'll need it much there compared to here

content/docs/2.10/operate/security.md
```diff
@@ -0,0 +1,15 @@
...
+
+All the components read the folder `/certs` looking for the certificates inside it, but this folder can be also modifyied using the console argument `--cert-dir`, patching the manifests or using helm values. Due to these certs are also used for internal communication between KEDA components, the CA certificate is also required for registering it as a trusted CA inside KEDA components.
```
zroubalik, line 15:
```suggestion
All components inspect the folder `/certs` for any certificates inside it. Argument `--cert-dir` can be used to specify another folder to be used as a source for certificates, this argument can be patched in the manifests or using Helm values. Because these certificates are also used for internal communication between KEDA components, the CA is also required to be registered as a trusted CA inside KEDA components.
```

content/docs/2.10/deploy.md
```diff
@@ -81,16 +81,26 @@ Locate installed KEDA Operator in `keda` namespace, then remove created `KedaCon
 
 If you want to try KEDA on [Minikube](https://minikube.sigs.k8s.io) or a different Kubernetes deployment without using Helm you can still deploy it with `kubectl`.
 
-- We provide sample YAML declaration which includes our CRDs and all other resources in a file which is available on the [GitHub releases](https://github.com/kedacore/keda/releases) page.
+- We provide sample YAML declaration which includes our CRDs and all other resources in a file which is available on the [GitHub releases](https://github.com/kedacore/keda/releases) page. 
+  - We offer two options to deploy KEDA:
```
zroubalik, line 85: would be nice to add a link to the webhooks documentation here

JorTurFer:
> @JorTurFer I think you did it the other way around :D I would keep the file here and remove it from kedacore/keda (code repo) since it is unlikely that we'll need it much there compared to here

that's what I did, I updated the file here and removed from keda repo

tomkerkhove:
Nope:
image

tomkerkhove:
Nevermind me; it's a 95% delete and not a complete delete - Sorry.

JorTurFer:
> Nevermind me; it's a 95% delete and not a complete delete - Sorry.

Binary diffs xD

content/docs/2.10/operate/cluster.md
JorTurFer: yes we should

content/docs/2.2/concepts/_index.md
```diff
@@ -21,7 +21,7 @@ KEDA performs two key roles within Kubernetes:
 
 The diagram below shows how KEDA works in conjunction with the Kubernetes Horizontal Pod Autoscaler, external event sources, and Kubernetes' [etcd](https://etcd.io) data store:
 
-![KEDA architecture](/img/keda-arch.png)
+![KEDA architecture](/img/keda-arch-old.png)
```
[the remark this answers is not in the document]
JorTurFer: done

content/docs/2.10/operate/security.md
JorTurFer: I'll add the example when I have the helm change ready

content/docs/2.10/deploy.md
JorTurFer: sure!

PR #1058, opened 2023-02-02
chore: set VersionTLS12 as min default TLS version

JorTurFer:
### Checklist

Related to https://github.com/kedacore/keda/issues/3961

### DIFF
--- a/content/docs/2.10/operate/cluster.md
+++ b/content/docs/2.10/operate/cluster.md
@@ -105,7 +105,7 @@ Some scalers issue HTTP requests to external servers (i.e. cloud services). As c
 
 Our industry has seen an evolution of TLS versions and some are more secure than another. For example, TLS1.0 and TLS1.1 have known vulnerabilities.
 
-By default, KEDA uses TLS1.2 as a minimum TLS version given it is the lowest version without vulnerabilities. However, if you need to support and older version you can configure it by using the environment variable `KEDA_HTTP_MIN_TLS_VERSION`.
+By default, KEDA uses TLS1.2 as a minimum TLS version given it is the lowest version without vulnerabilities. However, if you need to support another version you can configure it by using the environment variable `KEDA_HTTP_MIN_TLS_VERSION`.
 
 For example:
 ```yaml
--- a/content/docs/2.10/operate/cluster.md
+++ b/content/docs/2.10/operate/cluster.md
@@ -103,8 +103,11 @@ Some scalers issue HTTP requests to external servers (i.e. cloud services). As c
 
 ## HTTP TLS min version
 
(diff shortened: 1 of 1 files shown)


content/docs/2.10/operate/cluster.md
```diff
@@ -101,6 +101,15 @@ Some scalers issue HTTP requests to external servers (i.e. cloud services). As c
     NO_PROXY: 10.0.0.0/8
 ```
 
+## HTTP TLS min version
+
+Nowadays, not all TLS versions are secure as TLS1.0 and TLS1.1 have know vulnerabilities. As default, KEDA uses TLS1.2 as minimum because it's minimum non broken version. If you have any service which needs older TLS versions, the minimum version can be configured using the environment variable `KEDA_HTTP_MIN_TLS_VERSION`. Valid values are: `TLS13`, `TLS12`, `TLS11` and `TLS10`.
+
```
tomkerkhove, line 107:
```suggestion
Our industry has seen an evolution of TLS versions and some are more secure than another. For example, TLS1.0 and TLS1.1 have known vulnerabilities.

By default, KEDA uses TLS1.2 as a minimum TLS version given it is the lowest version without vulnerabilities. However, if you need to support and older version you can configure it by using the environment variable `KEDA_HTTP_MIN_TLS_VERSION`.

For example:
```

content/docs/2.10/operate/cluster.md
```diff
@@ -101,6 +101,15 @@ Some scalers issue HTTP requests to external servers (i.e. cloud services). As c
...
+```yaml
+- env:
+    KEDA_HTTP_MIN_TLS_VERSION: TLS13
+```
+
```
tomkerkhove, line 112:
```suggestion

The following values are allowed: `TLS13`, `TLS12`, `TLS11` and `TLS10`.

```

content/docs/2.10/operate/cluster.md
```diff
@@ -101,6 +101,20 @@ Some scalers issue HTTP requests to external servers (i.e. cloud services). As c
     NO_PROXY: 10.0.0.0/8
 ```
 
+## HTTP TLS min version
+
+Our industry has seen an evolution of TLS versions and some are more secure than another. For example, TLS1.0 and TLS1.1 have known vulnerabilities.
+
+By default, KEDA uses TLS1.2 as a minimum TLS version given it is the lowest version without vulnerabilities. However, if you need to support and older version you can configure it by using the environment variable `KEDA_HTTP_MIN_TLS_VERSION`.
```
zroubalik, line 108:
```suggestion
By default, KEDA uses TLS1.2 as a minimum TLS version given it is the lowest version without vulnerabilities. However, if you need to support another version you can configure it by using the environment variable `KEDA_HTTP_MIN_TLS_VERSION`.
```

zroubalik (changes requested):

PR #1062, opened 2023-02-05
fix: Prometheus metric is `scaler_errors_total` and not `scaler_error_totals`

JorTurFer:
I have noticed that the `*_scaler_error_totals` isn't correct because in the code (for all the versions) is `*_scaler_errors_total`
For example, in [v2.0](https://github.com/kedacore/keda/blob/v2.0.0/pkg/metrics/prometheus_metrics.go#L18) or [v2.10](https://github.com/kedacore/keda/blob/main/pkg/prommetrics/prommetrics.go#L44)

### Checklist


zroubalik (approved):
good catch!

PR #1074, opened 2023-02-21
Deprecate explicitly setting `metricName` for all scalers

yardenshoham:
In v2.12 this setting will be internalized and should not be set by users
### Checklist

Docs part of https://github.com/kedacore/keda/pull/4235
Relates to https://github.com/kedacore/keda/issues/4220


JorTurFer:
> @JorTurFer @zroubalik Any thoughts on annotating previous version to create awareness on the deprecation?

IDK... I guess that adding the annotation only to the versions where it is deprecated is more consistent. I mean, we have introduced a warning in the code. Let's add the annotation only to versions where the message is logged but I don't have a strong opinion, if @zroubalik agrees with annotating other versions too, go ahead

zroubalik (approved):

PR #1077, opened 2023-02-28
GitHub Runner Scaler documentation for keda#1782

Eldarrin:
https://github.com/kedacore/keda/issues/1732

Added a github runner scaler for GitHub Actions

### Checklist

Fixes #

### DIFF
--- a/content/docs/2.10/scalers/github-runner.md
+++ b/content/docs/2.10/scalers/github-runner.md
@@ -1,0 +1,153 @@
++++
+title = "Github Runner Scaler"
+availability = "v2.10+"
+maintainer = "GitHub"
+description = "Scale GitHub Runners based on the number of queued jobs in GitHub Actions"
+go_file = "github_runner_scaler"
++++
+
+### Trigger Specification
+
+This specification describes the `github-runner-scaler` trigger that scales based on queued jobs in GitHub Actions.
+
+```yaml
+triggers:
+  - type: github-runner
+    metadata:
+      # Optional: The URL of the GitHub API, defaults to https://api.github.com
+      githubAPIURL: "https://api.github.com"
+      # Required: The owner of the GitHub repository, or the organization that owns the repository
+      owner: "{owner}"
+      # Required: The scope of the runner, can be either "org", "ent" or "repo"
+      runnerScope: "{runnerScope}"
+      # Required: The personal access token to access the GitHub API
+      personalAccessToken: "{personalAccessToken}"
+      # Optional: The list of repositories to scale, separated by comma
+      repos: "{repos}"
+      # Optional: The list of runner labels to scale on, separated by comma
+      labels: "{labels}"
+      # Optional: The target number of queued jobs to scale on
+      targetWorkflowQueueLength: "1" # Default 1
+      # Optional: The target number of queued jobs to achieve to activate the scaler
+      activationTargetWorkflowQueueLength: "0" # Default 0
+    authenticationRef:
+      name: personalAccessToken
+```
+
+**Parameter list:**
+
+- `githubAPIURL` - The URL of the GitHub API, defaults to https://api.github.com. You should only need to modify this if you have your own GitHub Appliance. (Optional)
+- `owner` - The owner of the GitHub repository, or the organization that owns the repository. (Required)
+- `runnerScope` - The scope of the runner, can be either "org", "ent" or "repo". (Required)
+- `personalAccessToken` - The personal access token to access the GitHub API. (Required)
+- `repos` - The list of repositories to scale, separated by comma. (Optional)
+- `labels` - The list of runner labels to scale on, separated by comma. (Optional)
+- `targetWorkflowQueueLength` - The target number of queued jobs to scale on. (Optional, Default: 1)
+- `activationTargetWorkflowQueueLength` - The target number of queued jobs to achieve to activate the scaler. (Optional, Default: 0)
+
+*Parameters from Environment Variables*
+
+You can access each parameter from above using environment variables. When you specify the parameter in metadata with a suffix of `FromEnv`, 
+the scaler will use the value from the environment variable. The environment variable must be available to the manifest. e.g. `labelsFromEnv: "RUNNER_LABELS"` will use the environment variable `RUNNER_LABELS` as the source fo the `labels` parameter.
+
+There are also a set of default environment variables that will be used if none of the above are specified directly.
+
+- `githubAPIURL` - `GITHUB_API_URL`
+- `runnerScope` - `RUNNER_SCOPE`
+- `personalAccessToken` - `ACCESS_TOKEN`
+- `labels` - `LABELS`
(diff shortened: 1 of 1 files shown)


content/docs/2.10/scalers/github-runner.md
```diff
@@ -0,0 +1,153 @@
...
+      activationTargetWorkflowQueueLength: "0" # Default 0
+    authenticationRef:
+      name: personalAccessToken
+```
+
+**Parameter list:**
+
+- `githubAPIURL` - The URL of the GitHub API, defaults to https://api.github.com. You should only need to modify this if you have your own GitHub Appliance. (Optional)
+- `owner` - The owner of the GitHub repository, or the organization that owns the repository. (Required)
+- `runnerScope` - The scope of the runner, can be either "org", "ent" or "repo". (Required)
```
tomkerkhove, line 41: You might want to elaborate on what `ent` means? Enterprise I presume?

content/docs/2.10/scalers/github-runner.md
```diff
@@ -0,0 +1,153 @@
...
+- `personalAccessTokenFromEnv` - The personal access token to access the GitHub API. (Required)
```
tomkerkhove, line 42: Is it really required if somebody uses TriggerAuthentication?

content/docs/2.10/scalers/github-runner.md
```diff
@@ -0,0 +1,153 @@
...
+- `repos` - The list of repositories to scale, separated by comma. (Optional)
+- `labels` - The list of runner labels to scale on, separated by comma. (Optional)
+- `targetWorkflowQueueLength` - The target number of queued jobs to scale on. (Optional, Default: 1)
```
tomkerkhove, line 45: I think we might just simplify this:
```suggestion
- `queueLength` - The target number of queued jobs to scale on. (Optional, Default: 1)
```

tomkerkhove: Ok, I think I'm not fully getting the difference with `activationTargetWorkflowQueueLength`?

content/docs/2.10/scalers/github-runner.md
```diff
@@ -0,0 +1,153 @@
...
+
+The scaler will query the GitHub API in the following order:
+
+If no repos are specified: Get the list of repos for the specified owner.
+
+For each repo: Get the list of workflows runs in the repo.
+
+For each queued workflow run: Get the list of jobs in the queued workflow run.
+
+For each job: if the scaler matches, increment the queue length for that scaler.
```
tomkerkhove, line 90:
```suggestion
- If no repos are specified: Get the list of repos for the specified owner.
- For each repo: Get the list of workflows runs in the repo.
- For each queued workflow run: Get the list of jobs in the queued workflow run.
- For each job: if the scaler matches, increment the queue length for that scaler.
```

content/docs/2.10/scalers/github-runner.md
```diff
@@ -0,0 +1,153 @@
...
+There are also a set of default environment variables that will be used if none of the above are specified directly.
+
+- `githubAPIURL` - `GITHUB_API_URL`
+- `runnerScope` - `RUNNER_SCOPE`
+- `personalAccessTokenFromEnv` - `ACCESS_TOKEN`
+- `labels` - `LABELS`
+- `owner` - No defaults, must be specified if runnerScope is not `org` or `ent`. Otherwise `ORG_NAME` or `ENTERPRISE_NAME` will be used.
+- `repos` - No defaults, must be specified or left empty
+- `targetWorkflowQueueLength` - No defaults, must be specified or left empty
+- `activationTargetWorkflowQueueLength` - No defaults, must be specified or left empty
```
tomkerkhove, line 62: I'm not sure this is how we want it, to be honest.

I would expect that either end-users explicitly specify them through ie `labelsFromEnv` or not as this can lead to tricky scenarios.

Because of that, I would propose not to load from ENV VAR by default, unless explicitly configured in the trigger metadata.

Thoughts @kedacore/keda-maintainers?

Also, let's add all the possible `FromEnv`  parameters to the list above as they are not mentioned except for `personalAccessTokenFromEnv`.

content/docs/2.10/scalers/github-runner.md
```diff
@@ -0,0 +1,153 @@
...
+
+As an alternative to using environment variables, you can authenticate with GitHub using a Personal Access Token via `TriggerAuthentication` configuration.
+
+**Personal Access Token Authentication:**
+
+- `personalAccessToken` - The Personal Access Token (PAT) for GitHub from your user.
+
+### How does it work?
+
+The scaler will query the GitHub API to get the number of queued jobs in the specified repositories. If the number of queued jobs is greater than the `targetWorkflowQueueLength`, the scaler will scale up. If the number of queued jobs is less than the `activationTargetWorkflowQueueLength`, the scaler will scale down.
```
tomkerkhove, line 74: Do we need 2 configuration values here? For typical queue-based workloads we only have 1.

content/docs/2.10/scalers/github-runner.md
```diff
@@ -0,0 +1,153 @@
...
+
+If no `repos` are specified, the scaler will query all repositories in the specified `owner`. This is useful if you want to scale on all repositories in an organization, but will result in a lot of API calls and affect the Rate Limit.
+
+The `labels` parameter is used to filter the runners that the scaler will scale. It uses the minimum applicable label for the runner. For example, if you have a runner with the labels `golang` and `helm`, and you specify `helm` in the `labels` field on the GitHub Action, the scaler will scale up that runner.
```
tomkerkhove, line 78: Let's format the content a bit better, maybe?

```suggestion
We provide various options to have granular control over what runners to scale:

- **Repository Filtering** - If no `repos` are specified, the scaler will query all repositories in the specified `owner`. This is useful if you want to scale on all repositories in an organization, but will result in a lot of API calls and affect the Rate Limit.
- **Label-based Filtering** - The `labels` parameter is used to filter the runners that the scaler will scale. It uses the minimum applicable label for the runner. For example, if you have a runner with the labels `golang` and `helm`, and you specify `helm` in the `labels` field on the GitHub Action, the scaler will scale up that runner.
```

content/docs/2.10/scalers/github-runner.md
Eldarrin: done

content/docs/2.10/scalers/github-runner.md
Eldarrin: removed meta element

content/docs/2.10/scalers/github-runner.md
Eldarrin: removed `activationTargetWorkflowQueueLength` but think we should stick with `targetWorkflowQueueLength` to differentiate from `queueLength` which is the length of the in-process queue as opposed to the spin up target

content/docs/2.10/scalers/github-runner.md
Eldarrin: activation removed

zroubalik (approved):

PR #1078, opened 2023-03-01
Change version # to 2.9.3 (was still at 2.9.2) in the "Deploying using the deployment YAML files" section.

frachel:
Change version # to 2.9.3 (was still at 2.9.2) in the "Deploying using the deployment YAML files" section.


JorTurFer (approved):

JorTurFer:
Thanks for the fix!
Could you sign your commit? DCO check is failing

frachel:
> Thanks for the fix! Could you sign your commit? DCO check is failing

Where do I do that now that I've already done a PR? When I committed to my fork I did have the DCO in my commit, but I guess when I did the PR I had to do it there as well? (We've only switched to git recently after 20+ years on SVN so its a little foreign to me).

JorTurFer:
Hey,
No worries, it isn't something obvious. You need to sign you commit locally and push it. In this case, as you have already pushed it without signing it, you will need to modify that commit and push again your branch, forcing the push because the branch will change.
We have a doc section [about it](https://github.com/kedacore/keda/blob/main/CONTRIBUTING.md#i-didnt-sign-my-commit-now-what)  (because as I said, it isn't something obvious)

PR #1097, opened 2023-03-24
Azure managed Prometheus doc update

raggupta-ms:
1. adding info about what persmission the identity needs to access Azure Monitor Workspace
2. updating azure managed prometheus service name to reflect the official service name.
3. removing other clouds (other than Azure public cloud) support since managed prometheus service is only present in public cloud as of now. When it will be preesnt in other clouds, will update the doc.

### Checklist

Fixes #

### DIFF
--- a/content/docs/2.10/scalers/prometheus.md
+++ b/content/docs/2.10/scalers/prometheus.md
@@ -26,10 +26,6 @@ triggers:
     customHeaders: X-Client-Id=cid,X-Tenant-Id=tid,X-Organization-Id=oid # Optional. Custom headers to include in query. In case of auth header, use the custom authentication or relevant authModes.
     ignoreNullValues: false # Default is `true`, which means ignoring the empty value list from Prometheus. Set to `false` the scaler will return error when Prometheus target is lost
     unsafeSsl: "false" #  Default is `false`, Used for skipping certificate check when having self signed certs for Prometheus endpoint    
-    # Valid when using Azure managed service for Prometheus
-    cloud: Private # Default is `AzurePublicCloud`
-    # Required when cloud = Private
-    azureManagedPrometheusResourceURL: https://prometheus.monitor.azure.airgap/.default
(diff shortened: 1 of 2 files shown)


content/docs/2.10/scalers/prometheus.md
```diff
@@ -76,13 +70,12 @@ You can use `TriggerAuthentication` CRD to configure the authentication. It is p
 
 > 💡 **NOTE:**It's also possible to set the CA certificate regardless of the selected `authModes` (also without any authentication). This might be useful if you are using an enterprise CA.
 
-**Azure managed service for Prometheus**
+**Azure monitor managed service for Prometheus**
```
tomkerkhove, line 73: Nit - It's a product name
```suggestion
**Azure Monitor managed service for Prometheus**
```

tomkerkhove (commented):
LGTM, but can you also update v2.11 please?

raggupta-ms:
> LGTM, but can you also update v2.11 please?

not sure if we want that. Treat this doc update as a bug in the documentation for KEDA 2.10. Now I am fixing this doc bug. There are no code changes assosiated with this doc update and the updated documentation is valid for KEDA 2.10.
Does that make sense?

content/docs/2.10/scalers/prometheus.md
raggupta-ms: yes, updated.

tomkerkhove:
Not really as everything from 2.10 is also in 2.11 or above so we have to align it, unfortunately

raggupta-ms:
> Not really as everything from 2.10 is also in 2.11 or above so we have to align it, unfortunately

Ahh.. I think I misread your earlier comment. I though you said make this change on 2.11 and not on 2.10. But you said make this change on **2.11 as well.** Makes sense, updated.

tomkerkhove (approved):

PR #1119, opened 2023-04-25
Welcome LeanIX, a new KEDA user!

dan-leanix:
Added a new KEDA user.


tomkerkhove (approved):
Thank you very much for using KEDA and opening the PR @dan-leanix!

Issue #1112, opened 2023-04-23
Provide documentation on setting up authentication with GCP workload identity


raghu-manne:
A clear documentation on how to set up workload identity

Would be good if a [reference](https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity#authenticating_to) or step by step guide on how to set up workload identity is added.

It is also unclear on which service account of keda setup should be used for setting up workload identity though we have only one service account. It can easily be understandable to a new comer if these things are explicitly mentioned in the documentation.

Referring to this link: https://keda.sh/docs/2.10/authentication-providers/gcp-workload-identity/

raghu-manne:
@tomkerkhove 
Would love to work on this part of the documentation.

tomkerkhove:
Wonderful, thank you!

raghu-manne:
Made changes in PR #1120 and are merged to main.
Closing this issue.

tomkerkhove:
Thank you!

PR #1120, opened 2023-04-25
documentation for setting up GCP workload identity

raghu-manne:
Added step by step documentation of setting up workload identity for GCP.
### Checklist

Fixes issue [1112](https://github.com/kedacore/keda-docs/issues/1112)

Added steps by referring GCP [Documentation](https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity#authenticating_to).

### DIFF
--- a/content/docs/2.10/authentication-providers/gcp-workload-identity.md
+++ b/content/docs/2.10/authentication-providers/gcp-workload-identity.md
@@ -10,3 +10,61 @@ You can tell KEDA to use GCP Workload Identity via `podIdentity.provider`.
 podIdentity:
   provider: gcp # Optional. Default: none
 ```
+
+### Steps to setup Workload Identity
+If you are using podIdentity provider as `gcp`, you need to setup workload identity as below and your GKE cluster must have Workload Identity enabled.
+
+* You need to create a GCP IAM service account with proper permissions to retrive metrics for particular scalers.
+
+  ```shell
+  gcloud iam service-accounts create GSA_NAME \
+  --project=GSA_PROJECT
+  ```
+    
+  Replace the following: \
+  GSA_NAME: the name of the new IAM service account.\
+  GSA_PROJECT: the project ID of the Google Cloud project for your IAM service account.
+
+
+* Ensure that your IAM service account has the [roles](https://cloud.google.com/iam/docs/understanding-roles) you need. You can grant additional roles using the following command:
+
+  ```shell
+  gcloud projects add-iam-policy-binding PROJECT_ID \
+  --member "serviceAccount:GSA_NAME@GSA_PROJECT.iam.gserviceaccount.com" \
+  --role "ROLE_NAME"
+  ```
+
+  Replace the following:
+
+  PROJECT_ID: your Google Cloud project ID. \
+  GSA_NAME: the name of your IAM service account. \
+  GSA_PROJECT: the project ID of the Google Cloud project of your IAM service account. \
+  ROLE_NAME: the IAM role to assign to your service account, like roles/monitoring.viewer.
+
+* Allow the Kubernetes service account to impersonate the IAM service account by adding an IAM policy binding between the two service accounts. This binding allows the Kubernetes service account to act as the IAM service account.
+  ```shell
+  gcloud iam service-accounts add-iam-policy-binding GSA_NAME@GSA_PROJECT.iam.gserviceaccount.com \
+      --role roles/iam.workloadIdentityUser \
+      --member "serviceAccount:PROJECT_ID.svc.id.goog[NAMESPACE/KSA_NAME]"
+  ```
+  Replace the following:
+
+  PROJECT_ID: your Google Cloud project ID. \
+  GSA_NAME: the name of your IAM service account. \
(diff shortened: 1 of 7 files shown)


content/docs/2.10/authentication-providers/gcp-workload-identity.md
```diff
@@ -10,3 +10,56 @@ You can tell KEDA to use GCP Workload Identity via `podIdentity.provider`.
...
+  PROJECT_ID: your Google Cloud project ID. \
+  GSA_NAME: the name of your IAM service account. \
+  GSA_PROJECT: the project ID of the Google Cloud project of your IAM service account. \
+  ROLE_NAME: the IAM role to assign to your service account, like roles/monitoring.viewer.
+
+* Allow the Kubernetes service account to impersonate the IAM service account by adding an IAM policy binding between the two service accounts. This binding allows the Kubernetes service account to act as the IAM service account.
+  ```shell
+  gcloud iam service-accounts add-iam-policy-binding GSA_NAME@GSA_PROJECT.iam.gserviceaccount.com \
+      --role roles/iam.workloadIdentityUser \
+      --member "serviceAccount:PROJECT_ID.svc.id.goog[keda/keda-operator]"
```
JorTurFer, line 48: is `keda/keda-operator` the namespace/serviceAccountName? If yes, we should add them as parameters because both are configurable.

JorTurFer (commented):
Awesome! Nice improvement ❤️ 
I have left a comment inline and another 2 small things:
- should we add somewhere the link to the official docs? https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity
- could you update other versions too? (KEDA >= v2.5)

content/docs/2.10/authentication-providers/gcp-workload-identity.md
raghu-manne: Added as suggested.

raghu-manne:
@JorTurFer 

- Added official GCP Documentation link for reference.
- Will add to other versions of the documentation once this is finalised to avoid reworking on all the other versions.
- Also, GCP Workload Identity needs a nodeSelector to be added to the keda-operator workload
   ```yaml
    spec:
      serviceAccountName: KSA_NAME
      nodeSelector:
        iam.gke.io/gke-metadata-server-enabled: "true"
    ``` 
    I think we should add a condition in helm chart which will include this nodeSelector when [podIdentity](https://github.com/kedacore/charts/blob/main/keda/values.yaml#L140) is set to gcp.
Thoughts?

JorTurFer:
> * I think we should add a condition in helm chart which will include this nodeSelector when [podIdentity](https://github.com/kedacore/charts/blob/main/keda/values.yaml?rgh-link-date=2023-04-26T04%3A48%3A06Z#L140) is set to gcp.

Definitively, we already have that option for Azure and AWS, I think that we can add it for GCP (to add the node selector but also to annotate the service account, not having to do it manually after)

JorTurFer:
> * Will add to other versions of the documentation once this is finalised to avoid reworking on all the other versions.

Nice! I think that now looks nice to replicate in other versions :)

raghu-manne:
> > * I think we should add a condition in helm chart which will include this nodeSelector when [podIdentity](https://github.com/kedacore/charts/blob/main/keda/values.yaml?rgh-link-date=2023-04-26T04%3A48%3A06Z#L140) is set to gcp.
> 
> Definitively! we already have that option for Azure and AWS, I think that we can add it for GCP (to add the node selector but also to annotate the service account, not having to do it manually after)

Thanks, for annotating service account we need to have GCP IAM Service account name and Project ID. Not sure how can we handle that if we try to annotate from helm chart itself.





> > * Will add to other versions of the documentation once this is finalised to avoid reworking on all the other versions.
> 
> Nice! I think that now looks nice to replicate in other versions :)

Will raise separate PR for that.

JorTurFer:
> Thanks, for annotating service account we need to have GCP IAM Service account name and Project ID. Not sure how can we handle that if we try to annotate from helm chart itself.
> 

Yes if you do the other steps before update KEDA chart. I mean, you know the namespace where you will deploy KEDA, and you also know the service account name, you could do the other steps before, and then upgrade KEDA in the cluster



> > > * Will add to other versions of the documentation once this is finalised to avoid reworking on all the other versions.
> > 
> > 
> > Nice! I think that now looks nice to replicate in other versions :)
> 
> Will raise separate PR for that.

Please, update all the versions as part of this PR 🙏

raghu-manne:
>Yes if you do the other steps before update KEDA chart. I mean, you know the namespace where you will deploy KEDA, and you also know the service account name, you could do the other steps before, and then upgrade KEDA in the cluster
>
Makes sense.
>Please, update all the versions as part of this PR
>
Added for all versions starting from 2.5

JorTurFer (approved):
Thanks a lot for the imporvement! ❤️

JorTurFer:
> > Yes if you do the other steps before update KEDA chart. I mean, you know the namespace where you will deploy KEDA, and you also know the service account name, you could do the other steps before, and then upgrade KEDA in the cluster
> 
> Makes sense.

Would you open a PR in kedacore/charts with the changes?

PR #1126, opened 2023-05-04
fix misleading doc that states cpu/mem scalers use limit

### LINKED ISSUE #1028: Incorrect CPU Scaler documentation
# Which page requires fixing ?

https://keda.sh/docs/2.9/scalers/cpu/#prerequisites

# Why this page requires fixing ?

https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/#how-does-a-horizontalpodautoscaler-work

> the controller calculates the utilization value as a percentage of the equivalent [resource request](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#requests-and-limits) on the containers in each Pod

Limits don't take part in HPA autoscaling. Limits are used just to limit the number of resources specific container can use.
That's why we can set for example 200% CPU utilization for target of HPA..

Adarsh-verma-14:
Signed-off-by: Adarsh-verma-14 <t_adarsh.verma@india.nec.com>

These scalers(cpu/mem) use resource requests not limits.
### Checklist

Fixes #1028

### DIFF
--- a/content/docs/2.0/scalers/cpu.md
+++ b/content/docs/2.0/scalers/cpu.md
@@ -13,19 +13,19 @@ go_file = "cpu_memory_scaler"
 
 ### Prerequisites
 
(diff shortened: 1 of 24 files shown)


JorTurFer (approved):

JorTurFer:
Thanks for the fix!

PR #1137, opened 2023-05-30
Admission Webhooks won't be mandatory

zroubalik:
### Checklist

Relates: https://github.com/kedacore/charts/pull/453

### DIFF
--- a/content/docs/2.10/concepts/admission-webhooks.md
+++ b/content/docs/2.10/concepts/admission-webhooks.md
@@ -4,8 +4,6 @@ description = "Automatically validate resource changes to prevent misconfigurati
 weight = 600
 +++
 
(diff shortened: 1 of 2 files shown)


JorTurFer (approved):

tomkerkhove:
@JorTurFer @zroubalik Care to share why we suddenly changed our mind? Making it an opt-out is different from making it mandatory IMO since it is configurable.

I presume you mean in terms of the YAML that we provide?

JorTurFer:
I think that we discussed it by slack (but I could be wrong or maybe we did it in other channel) all together, but the summary is that there are cases where users are not allowed to deploy cluster scoped resources (at least in my mind)

As default the feature will be activated, but the option in the chart will be maintained

zroubalik:
Yeah, let's keep it a default option, but keep an option to disable for a specific cases

PR #1144, opened 2023-06-06
Update scaling-jobs.md

eugen-nw:
Explained the minReplicaCount scale-out behavior in response to incoming events.

_I explained the minReplicaCount-dictated scale-out behavior in response to incoming events._

### Checklist

Fixes #

### DIFF
--- a/content/docs/2.8/concepts/scaling-jobs.md
+++ b/content/docs/2.8/concepts/scaling-jobs.md
@@ -107,7 +107,9 @@ ___
   minReplicaCount: 10 # Optional. Default: 0
 ```
 
-The min number of jobs that is created by default. This can be useful to avoid bootstrapping time of new jobs. If minReplicaCount is greater than maxReplicaCount, minReplicaCount will be set to maxReplicaCount. 
+The min number of jobs that is created by default. This can be useful to avoid bootstrapping time of new jobs. If minReplicaCount is greater than maxReplicaCount, minReplicaCount will be set to maxReplicaCount.  
+  
+All messages will create new jobs - within the limits imposed by maxReplicaCount - in order to reach the state where minReplicaCount jobs are always running.  For example, if one sets minReplicaCount to 2 then there will be 2 jobs running permanently.  If 3 new messages are sent, 2 of those messages will be processed on the already running jobs.  A minimum of 2 new jobs will be created in order to fulfill the desired state dictated by the minReplicaCount parameter that is set to 2.
 ___
 
 ---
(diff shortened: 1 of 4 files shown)


content/docs/2.8/concepts/scaling-jobs.md
```diff
@@ -107,7 +107,9 @@ ___
   minReplicaCount: 10 # Optional. Default: 0
 ```
 
-The min number of jobs that is created by default. This can be useful to avoid bootstrapping time of new jobs. If minReplicaCount is greater than maxReplicaCount, minReplicaCount will be set to maxReplicaCount. 
+The min number of jobs that is created by default. This can be useful to avoid bootstrapping time of new jobs. If minReplicaCount is greater than maxReplicaCount, minReplicaCount will be set to maxReplicaCount.  
+  
+All messages will create new jobs - within the limits imposed by maxReplicaCount - in order to reach the state where minReplicaCount jobs are always running.  For example, if one sets minReplicaCount to 2 then there will be 2 jobs running permanently.  If 3 new messages are sent, 2 of those messages will be processed on the already running jobs.  A minimum of 2 new jobs will be created in order to fulfill the desired state dictated by the minReplicaCount parameter that is set to 2.
```
JorTurFer, line 112:
```suggestion
All messages will create new jobs - within the limits imposed by maxReplicaCount - in order to reach the state where minReplicaCount jobs are always running.  For example, if one set a minReplicaCount to 2 then there will be 2 jobs running permanently. With a targetValue of 1, if 3 new messages are sent, 2 of those messages will be processed on the already running jobs but another 3 jobs will be created in order to fulfill the desired state dictated by the minReplicaCount parameter that is set to 2.
```

WDYT? I reckon that it's more clear about what happens

JorTurFer (commented):
Could you add this change (once it's ready) to other versions? This feature was released on 2.8 but it's still present (v2.9, v2.10 and v2.11)

content/docs/2.8/concepts/scaling-jobs.md
tomkerkhove: Agreed!

eugen-nw:
I just added a new commit.  Please review and if it's OK, I'll port the change to the other versions of the doc as well.

tomkerkhove:
@eugen-nw Thanks! But can you also update v2.9, 2.10 & 2.11 please?

eugen-nw:
Done.

tomkerkhove (approved):

tomkerkhove:
Thanks!

PR #1161, opened 2023-06-19
Added doc to use oauth for pulsar scaler

mingmcb:
Added doc to use oauth for pulsar scaler

### Checklist

Related to https://github.com/kedacore/keda/pull/4709


content/docs/2.10/scalers/pulsar.md
```diff
@@ -32,7 +32,11 @@ triggers:
...
 - `activationMsgBacklogThreshold` - Target value for activating the scaler. Learn more about activation [here](./../concepts/scaling-deployments.md#activating-and-scaling-thresholds).(Default: `0`, Optional)
-- `authModes` - a comma separated list of authentication modes to use. (Values: `bearer`, `tls`,`basic`, Default: `""`, Optional, `tls,bearer` or `tls,basic` are valid combinations and would indicate mutual TLS to secure the connection and then `bearer` or `basic` headers should be added to the HTTP request)
+- `authModes` - a comma separated list of authentication modes to use. (Values: `bearer`, `tls`, `basic`, `oauth`, Default: `""`, Optional, `tls,bearer` or `tls,basic` are valid combinations and would indicate mutual TLS to secure the connection and then `bearer` or `basic` headers should be added to the HTTP request)
+- `oauthTokenURI` - Token endpoint for your OAuth provider
+- `grantType` - only `client_credentials` is supported
+- `scopes` - space delimited oauth scopes(Optional). 
+- `clientID` - clientID from your OAuth provider.  It will be ignored if clientID is also provided from `authenticationRef`
```
zroubalik, line 39: I think we should move this to TriggerAuthentication. We tend to not put credentials related stuff into trigger config. 

@tomkerkhove @JorTurFer WDYT

zroubalik: You can check Kafka scaler: https://keda.sh/docs/2.10/scalers/apache-kafka/#authentication-parameters

content/docs/2.10/scalers/pulsar.md
```diff
@@ -32,7 +32,11 @@ triggers:
 - `subscription` - Name of the topic subscription
 - `msgBacklogThreshold` - Average target value to trigger scaling actions. (default: 10)
 - `activationMsgBacklogThreshold` - Target value for activating the scaler. Learn more about activation [here](./../concepts/scaling-deployments.md#activating-and-scaling-thresholds).(Default: `0`, Optional)
-- `authModes` - a comma separated list of authentication modes to use. (Values: `bearer`, `tls`,`basic`, Default: `""`, Optional, `tls,bearer` or `tls,basic` are valid combinations and would indicate mutual TLS to secure the connection and then `bearer` or `basic` headers should be added to the HTTP request)
+- `authModes` - a comma separated list of authentication modes to use. (Values: `bearer`, `tls`, `basic`, `oauth`, Default: `""`, Optional, `tls,bearer` or `tls,basic` are valid combinations and would indicate mutual TLS to secure the connection and then `bearer` or `basic` headers should be added to the HTTP request)
+- `oauthTokenURI` - Token endpoint for your OAuth provider
+- `grantType` - only `client_credentials` is supported
```
zroubalik, line 37: if `client_credentials` is the only supported type, then we can make it a default one, right? And we don't have add a property for it in here.

content/docs/2.10/scalers/pulsar.md
JorTurFer: I don't think that `clientID` and `oauthTokenURI` are secrets, I mean, AFAIK, only the `clientSecret` is a sensitive data, clientId usually is part of the query string

JorTurFer: but we can move it into TriggerAuthentication as it's related with the credentials, but they aren't sensitive data

zroubalik: yeah, this is what I mean. It logically belongs there imho. Kafka is done this way also.

mingmcb: The clientID is already part of `TirggerAuthentication`. see the code example:  https://github.com/kedacore/keda-docs/pull/1161/files#diff-36e3935f303a0a808c6ee2bc08e18d62f54d555079a822c101882f3aec0198a6R268-R280

The idea is to have the additional `clientID` in the metadata is to mainly for ease of use. with[ OAuth RFC8705](https://tools.ietf.org/html/rfc8705), it is possible to request the access token without the secret. In this case, developer can just have everything in the metadata without secret. see example for this case:  https://github.com/kedacore/keda-docs/pull/1161/files#diff-36e3935f303a0a808c6ee2bc08e18d62f54d555079a822c101882f3aec0198a6R308-R328

content/docs/2.10/scalers/pulsar.md
mingmcb: yes, I can make it as default if works better.

content/docs/2.10/scalers/pulsar.md
zroubalik: Okay make sense. Could you please then add also the `scopes` and `oauthTokenURI` to TriggerAuth as well? The same way `clientID` is done.

content/docs/2.10/scalers/pulsar.md
zroubalik: yeah and remove the property from the config pls

content/docs/2.10/scalers/pulsar.md
```diff
@@ -58,6 +61,15 @@ When configuring mutual TLS authentication, configure the following:
 - `cert`: Certificate for client authentication.
 - `key`: Key for client authentication. 
 
+**OAuth 2**
+
+When configuring OAuth Authentication, configure the following:
+- `oauthTokenURI` - Token endpoint for your OAuth provider(Optional)
+- `scope` - space delimited oauth scopes(Optional)
+- `clientID`: the clientID  (optional)
+- `clientSecret`: the clientSecret  (optional)
```
zroubalik, line 70:
```suggestion
- `oauthTokenURI` - The OAuth Access Token URI for the OAuth provider. (Optional)
- `scope` -  A comma separated lists of OAuth scopes. (Optional)
- `clientID`: Client ID for the OAuth provider. (Optional)
- `clientSecret`: Client secret for the OAuth provider. (Optional)
```

zroubalik (commented):
`scope` should be a comma separated list, not space separated list. To be consistent with the rest of the config (eg. `authModes`)

@mingmcb could you please fix that?

zroubalik (commented):
please move this to content/docs/2.12

zroubalik (approved):
LGTM

PR #1189, opened 2023-07-17
new complexScalingLogic struct for SO definition

gauron99:
This PR is opened for https://github.com/kedacore/keda/pull/4583 new feature 

- [x] change webhook check for SO

### DIFF
--- a/content/docs/2.12/concepts/scaling-deployments.md
+++ b/content/docs/2.12/concepts/scaling-deployments.md
@@ -29,7 +29,7 @@ The only constraint is that the target `Custom Resource` must define `/scale` [s
 
 ## ScaledObject spec
 
-This specification describes the `ScaledObject` Custom Resource definition which is used to define how KEDA should scale your application and what the triggers are. The `.spec.ScaleTargetRef` section holds the reference to the target resource, ie. `Deployment`, `StatefulSet` or `Custom Resource`. 
+This specification describes the `ScaledObject` Custom Resource definition which is used to define how KEDA should scale your application and what the triggers are. The `.spec.ScaleTargetRef` section holds the reference to the target resource, ie. `Deployment`, `StatefulSet` or `Custom Resource`.
 
 [`scaledobject_types.go`](https://github.com/kedacore/keda/blob/main/apis/keda/v1alpha1/scaledobject_types.go)
 
@@ -49,7 +49,7 @@ spec:
     envSourceContainerName: {container-name}                # Optional. Default: .spec.template.spec.containers[0]
   pollingInterval:  30                                      # Optional. Default: 30 seconds
   cooldownPeriod:   300                                     # Optional. Default: 300 seconds
-  idleReplicaCount: 0                                       # Optional. Default: ignored, must be less than minReplicaCount 
+  idleReplicaCount: 0                                       # Optional. Default: ignored, must be less than minReplicaCount
   minReplicaCount:  1                                       # Optional. Default: 0
   maxReplicaCount:  100                                     # Optional. Default: 100
   fallback:                                                 # Optional. Section to specify fallback options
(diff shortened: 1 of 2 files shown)


content/docs/2.12/concepts/scaling-deployments.md
```diff
@@ -66,6 +66,17 @@ spec:
           - type: Percent
             value: 100
             periodSeconds: 15
+    complexScalingLogic:                                    # Optional. Section to specify complex scaling logic
```
JorTurFer, line 69: `complexScalingLogic` sounds weird, doesn't? WDYT about something like `modifiers` ?

content/docs/2.12/concepts/scaling-deployments.md
```diff
@@ -66,6 +66,17 @@ spec:
...
+      formula: {formula-for-fetched-metrics}                # Optional. Formula for calculation
+      target: {target-value-to-scale-on}                    # Optional. If metrics are anyhow composed together
+      externalCalculators:                                  # Optional. Section for user defined grpc calculation servers
```
JorTurFer, line 72: As above, something like `externalModifiers`?

content/docs/2.12/concepts/scaling-deployments.md
```diff
@@ -66,6 +66,17 @@ spec:
...
+        - name: {name-of-ec}                                # Mandatory if externalCalculators section is given
+          url: {url-of-ec-server}                           # Mandatory if externalCalculators section is given
+          timeout: {time-to-connect}                        # Mandatory if externalCalculators section is given
+          certDirectory: ""                                 # Optional. Defaults to "", no certificates used
+        fallback:                                           # Optional. Section to specify ec-fallback options
+          replicas: 3                                       # Mandatory if fallback section is included
+          failureThreshold: 3                               # Mandatory if fallback section is included
```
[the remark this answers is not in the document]
JorTurFer: If this fallback is for the external server, I think that it's wrong indented

JorTurFer (commented):
I ❤️  this feature!

Maybe we should add an example for each case, as this is an advanced feature, some examples could help with the adoption

content/docs/2.12/concepts/scaling-deployments.md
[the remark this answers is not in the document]
gauron99: the fallback is for external servers, if one of them fails, the fallback is applied, works the same like normal triggers. I was thinking if triggers have error, normal fallback is applied and this logic can still work, therefore if any of the server fail, that would be the fallback for it. But i agree, this could be simplified. I could simply call the same fallback function as the standard one if an error occurs in external servers

content/docs/2.12/concepts/scaling-deployments.md
gauron99: `externalModifiers` does sound pretty good. combining these two would then be 
```
modifiers:
 externalModifiers:
  - name: ...
 ```
what about just leaving the servers as `external` as they would be already in `modifiers` field?

gauron99:
> Maybe we should add an example for each case, as this is an advanced feature, some examples could help with the adoption
yes Im definitelly going to add examples, Im still drafting the actual PR as well (i shouldve probably put this as draft as well, thats my bad), i thought of more cases to test and want to cleanup unnecessary code before putting it up for review. Im happy to change the structures before that so we can discuss here

content/docs/2.12/concepts/scaling-deployments.md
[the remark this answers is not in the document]
gauron99: ive got rid of the grpc server fallback entirely and everything is handled with the default fallback as it exists currently

content/docs/2.12/concepts/scaling-deployments.md
gauron99: Ive opened https://github.com/kedacore/keda/pull/4583 for review. Im guessing this will be discussed there

zroubalik:
@gauron99 this can be closed, right?

gauron99:
yes, sorry i completely forgot about this one @zroubalik

PR #1191, opened 2023-07-20
Add blurb about standalone CRDs

atzawada:
Relates to a future PR in kedacore/keda

### Checklist


JorTurFer (commented):
Only v2.12 has to be updated because v2.10 and v2.11 won't have that option as they are already released. Could you revert changes in v2.10 and v2.11?

atzawada:
Done! I thought 2.10 and 2.11 were still active release lines for some reason. Now it should only reflect the change for 2.12

JorTurFer (approved):

PR #1192, opened 2023-07-21
Search: implement version filter and migrate algolia to v3

### LINKED ISSUE #619: Add version number to doc search results
Please add the version number to search results when using the built-in search functionality on keda.sh. Currently, if you search for an item and it appears in older versions, then there may be confusion on which version you're using. I assumed that if there are multiple results, the highest item would be the latest, though this is not the case.

image

### Use-Case

As an example scenario, navigate to keda.sh. In the search box, enter "azure pod identity". This returns several results, all of which look like they point to the same page. It's only when you navigate over the links that you can see they point to different versions of Keda. Thanks to @tomkerkhove on #618 for pointing this out to me, as I hadn't realised.

To avoid this scenario, it would be great if the version of the page could also be added into the search result.

### Specification

- [ ] Add the version of the page into the search result.

thisisobate:

### Checklist

Fixes #619

### DIFF
--- a/layouts/partials/css.html
+++ b/layouts/partials/css.html
@@ -12,4 +12,4 @@
   {{ $prodCss := $css | fingerprint -}}
   <link rel="stylesheet" href="{{ $prodCss.RelPermalink }}" integrity="{{ $prodCss.Data.Integrity }}">
 {{ end -}}
-<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/docsearch.js@2/dist/cdn/docsearch.min.css" />
+<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/@docsearch/css@3" />
--- a/layouts/partials/javascript.html
+++ b/layouts/partials/javascript.html
@@ -23,14 +23,20 @@
 {{/* Algolia search bar */}}
 <script
   type="text/javascript"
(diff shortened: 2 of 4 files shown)


thisisobate:
@tomkerkhove

thisisobate:
cc: @JorTurFer @tomkerkhove Will appreciate a review/approval.
Thanks!

JorTurFer (approved):

JorTurFer:
❤️

thisisobate:
oh wow...taking a look now

thisisobate:
I just fixed it @JorTurFer

JorTurFer:
> I just fixed it @JorTurFer

Will you open another PR or it was something in the account itself?

thisisobate:
oh yes, I had to modify the config to produce the desired result.

JorTurFer:
I have noticed that in the main page (keda.sh) the search still doesn't work:
image

I guess that it's something related with not having a version defined, in versiones pages the search works nicely, but in the main page and inside the blog (both are cases without versioning, at least in the URL) it doesn't work

thisisobate:
Oh...just noticed.
Will push a fix soon @JorTurFer

JorTurFer:
And the last thing that I have noticed, in the search bar always says `latest` even if the version is any older version. For example having v2.9 selected:

image

This is totally non-critical issue because the search works fantastic now ❤️

PR #1198, opened 2023-07-25
feat: Add new feature `formula` for SO

gauron99:
provide documentation for new feature, which is defined in SO.
here -> [new feature implementation](https://github.com/kedacore/keda/pull/4821)
### Checklist

### DIFF
--- a/content/docs/2.12/concepts/scaling-deployments.md
+++ b/content/docs/2.12/concepts/scaling-deployments.md
@@ -222,7 +222,7 @@ The `complexScalingLogic` is optional. If defined, both `target` and `formula` a
 
   `formula` composes external metrics together and allows them to be modified/manipulated with. It accepts mathematical/conditional statements using [this external project](https://github.com/antonmedv/expr). If `fallback` is applied, the `formula` will NOT be applied. Complete language definition of `expr` package can be found [here](https://expr.medv.io/docs/Language-Definition). Formula must return a single value (not boolean).
 
-**Example 1**
+**Example: compose average value**
 
 ```yaml
 advanced:
@@ -244,9 +244,9 @@ triggers:
       valueLocation: "tasks"
 ```
 
-Formula composes 2 given metrics from 2 triggers `kubernetes-workload` named `trig_one` and `metrics-api` named `trig_two` together as an average value and returns one final metric which is passed to HPA instead of 2 metrics individually.
+Formula composes 2 given metrics from 2 triggers `kubernetes-workload` named `trig_one` and `metrics-api` named `trig_two` together as an average value and returns one final metric which is used to make autoscaling decisions on.
 
-**More formula examples**
+**Example: ternary operator**
 
 ```yaml
 advanced:
@@ -254,7 +254,9 @@ advanced:
     formula: "trig_one > 2 ? trig_one + trig_two : 1"
 ```
 
-If metric value of trigger `trig_one` is more then 2, then return `trig_one` + `trig_two` otherwise return 1.
+If metric value of trigger `trig_one` is more than 2, then return `trig_one` + `trig_two` otherwise return 1.
+
+**Example: count function**
 
 ```yaml
 advanced:
@@ -263,6 +265,8 @@ advanced:
 ```
 
 If atleast 2 metrics (from the list `trig_one`,`trig_two`,`trig_three`) have value of more than 1, then return 5, otherwise return 0
+
+**Example: nested conditions and operators**
 
 ```yaml
 advanced:
--- a/content/docs/2.12/concepts/scaling-deployments.md
+++ b/content/docs/2.12/concepts/scaling-deployments.md
@@ -210,17 +210,66 @@ advanced:
     formula: {formula-for-fetched-metrics}                # Mandatory. Formula for calculation
 ```
 
-**`complexScalingLogic:`**
+**`complexScalingLogic`**
 
-The `complexScalingLogic` is optional. If defined, both `target` and `formula` are mandatory. Using this structure creates `composite-scaler-metric` for the HPA that will replace all requests for external metrics and handle then internally. With `complexScalingLogic` all external triggers' names **must** be defined, for potentially being used in the `formula` itself.
+The `complexScalingLogic` is optional and **experimental**. If defined, both `target` and `formula` are mandatory. Using this structure creates `composite-scaler-metric` for the HPA that will replace all requests for external metrics and handle them internally. With `complexScalingLogic` all external triggers' names **must** be defined, for potentially being used in the `formula` itself.
 
 **`complexScalingLogic.target`**
 
-`target` defines new target value to scale on for the composed metric. All external metrics must be of the same type in order for ScaledObject to be successfully validated.
+`target` defines new target value to scale on for the composed metric. All scaler metrics must be of the same type in order for ScaledObject to be successfully validated.
 
 **`complexScalingLogic.formula`**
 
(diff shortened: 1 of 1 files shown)


content/docs/2.12/concepts/scaling-deployments.md
```diff
@@ -201,6 +201,80 @@ Starting from Kubernetes v1.18 the autoscaling API allows scaling behavior to be
...
+
+**`complexScalingLogic.target`**
+
+`target` defines new target value to scale on for the composed metric. All external metrics must be of the same type in order for ScaledObject to be successfully validated.
+
+**`complexScalingLogic.formula`**
+
+  `formula` composes external metrics together and allows them to be modified/manipulated with. It accepts mathematical/conditional statements using [this external project](https://github.com/antonmedv/expr). If `fallback` is applied, the `formula` will NOT be applied. Complete language definition of `expr` package can be found [here](https://expr.medv.io/docs/Language-Definition). Formula must return a single value (not boolean).
+
+**Example 1**
```
tomkerkhove, line 225: I'm thinking out loud - Everything above here is a nice overview, but all the rest below should go in a dedicated section similar to https://deploy-preview-1198--keda.netlify.app/docs/2.12/concepts/scaling-deployments/#pause-autoscaling.

This is because the samples are mixed with the conceptual doc which makes the whole section a bit message/unstructured. I'd do a brief introduction here, add a dedicated section down below and just point to it here. (or move everything in a dedicated section).

The current document for this section is already _way too long_

Thoughts @zroubalik / @JorTurFer?

content/docs/2.12/concepts/scaling-deployments.md
```diff
@@ -201,6 +201,80 @@ Starting from Kubernetes v1.18 the autoscaling API allows scaling behavior to be
...
+      value: '1'
+  - type: metrics-api
+    name: trig_two
+    metadata:
+      targetValue: "2"
+      url: "https://mockbin.org/bin/336a8d99-9e09-4f1f-979d-851a6d1b1423"
+      valueLocation: "tasks"
+```
+
+Formula composes 2 given metrics from 2 triggers `kubernetes-workload` named `trig_one` and `metrics-api` named `trig_two` together as an average value and returns one final metric which is passed to HPA instead of 2 metrics individually.
```
tomkerkhove, line 247: KEDA end-users shouldn't have to care what an HPA is
```suggestion
Formula composes 2 given metrics from 2 triggers `kubernetes-workload` named `trig_one` and `metrics-api` named `trig_two` together as an average value and returns one final metric which is used to make autoscaling decisions on.
```

content/docs/2.12/concepts/scaling-deployments.md
tomkerkhove, line 225: Let's give these examples names for better skimming of doc

content/docs/2.12/concepts/scaling-deployments.md
```diff
@@ -201,6 +201,80 @@ Starting from Kubernetes v1.18 the autoscaling API allows scaling behavior to be
...
+```yaml
+advanced:
+  complexScalingLogic:                                    # Optional. Section to specify complex scaling logic
+    target: {target-value-to-scale-on}                    # Mandatory. New target if metrics are anyhow composed together
+    formula: {formula-for-fetched-metrics}                # Mandatory. Formula for calculation
+```
+
+**`complexScalingLogic:`**
+
+The `complexScalingLogic` is optional. If defined, both `target` and `formula` are mandatory. Using this structure creates `composite-scaler-metric` for the HPA that will replace all requests for external metrics and handle then internally. With `complexScalingLogic` all external triggers' names **must** be defined, for potentially being used in the `formula` itself.
```
tomkerkhove, line 215: I presume we have validation on the trigger names?


Same here, end-users don't have to care about HPA. Also, I'm removing external as KEDA end-users might/will be confused to think this is only supported with external scalers and not built-in ones.
```suggestion
The `complexScalingLogic` is optional. If defined, both `target` and `formula` are mandatory. By using formula-based scaling logic, KEDA will compute the metric to make autoscaling decisions on. With `complexScalingLogic` all triggers' names **must** be defined, for potentially being used in the `formula` itself.
```

content/docs/2.12/concepts/scaling-deployments.md
```diff
@@ -201,6 +201,80 @@ Starting from Kubernetes v1.18 the autoscaling API allows scaling behavior to be
...
+
+**`complexScalingLogic.target`**
+
+`target` defines new target value to scale on for the composed metric. All external metrics must be of the same type in order for ScaledObject to be successfully validated.
```
tomkerkhove, line 219:
```suggestion
`target` defines new target value to scale on for the composed metric. All scaler metrics must be of the same type in order for ScaledObject to be successfully validated.
```

content/docs/2.12/concepts/scaling-deployments.md
```diff
@@ -201,6 +201,80 @@ Starting from Kubernetes v1.18 the autoscaling API allows scaling behavior to be
...
+
+**`complexScalingLogic.formula`**
+
+  `formula` composes external metrics together and allows them to be modified/manipulated with. It accepts mathematical/conditional statements using [this external project](https://github.com/antonmedv/expr). If `fallback` is applied, the `formula` will NOT be applied. Complete language definition of `expr` package can be found [here](https://expr.medv.io/docs/Language-Definition). Formula must return a single value (not boolean).
```
tomkerkhove, line 223:
```suggestion
  `formula` composes scaler metrics together and allows them to be modified/manipulated. It accepts mathematical/conditional statements using [this external project](https://github.com/antonmedv/expr). If `fallback` is applied, the `formula` will NOT be applied. Complete language definition of `expr` package can be found [here](https://expr.medv.io/docs/Language-Definition). Formula must return a single value (not boolean).
```

Fallback here means the fallback scaling feature we have or? We might want to clarify that

content/docs/2.12/concepts/scaling-deployments.md
gauron99: there is validation for having the trigger names defined when `formula` is specified in yaml. Currently oyu have to specify all triggers because there is no parsing. Iam currently working on parsing the formula to extract the names and match them to trigger names so user can define trigger names to only those triggers that are used in the formula which would be more intuitive I think.

I will follow up here

content/docs/2.12/concepts/scaling-deployments.md
zroubalik: Agree!

zroubalik (commented):
We should mark this as an Experimental feature in the docs.

See this older docs for reference: https://keda.sh/docs/2.9/concepts/scaling-deployments/#caching-metrics-experimental

gauron99:
@zroubalik @tomkerkhove PTAL

zroubalik (approved):

content/docs/2.12/concepts/scaling-deployments.md
```diff
@@ -201,6 +201,29 @@ Starting from Kubernetes v1.18 the autoscaling API allows scaling behavior to be
...
+```yaml
+advanced:
+  scalingModifiers:                                    # Optional. Section to specify scaling modifiers
+    target: {target-value-to-scale-on}                    # Mandatory. New target if metrics are anyhow composed together
+    formula: {formula-for-fetched-metrics}                # Mandatory. Formula for calculation
+```
+
+**`scalingModifiers`**
+
+The `scalingModifiers` is optional and **experimental**. If defined, both `target` and `formula` are mandatory. Using this structure creates `composite-scaler-metric` for the HPA that will replace all requests for external metrics and handle them internally. With `scalingModifiers` all external triggers' names **must** be defined, for potentially being used in the `formula` itself.
```
zroubalik, line 215:
```suggestion
The `scalingModifiers` is optional and **experimental**. If defined, both `target` and `formula` are mandatory. Using this structure creates `composite-metric` for the HPA that will replace all requests for external metrics and handle them internally. With `scalingModifiers` all external triggers' names **must** be defined, for potentially being used in the `formula` itself.
```

zroubalik (approved):

JorTurFer (approved):

PR #1226, opened 2023-09-19
Refactoring of azure-pipelines.md docs

jakuboskera:
Refactors `azure-pipelines.md` docs to follow best practise for writing Markdown documentations.

### Checklist


tomkerkhove:
Thanks for the PR; but I'm not sure if this is a best practice or just a preference :)

If we go this route, then we should have linting in place and update _all_ docs to align for consistency imo

PR #1234, opened 2023-10-01
chore: Add databaseIndex to documentation of redis-cluster scaler

peschmae:

Add `databaseIndex` argument to documentation of `redis-cluster` scaler.
The text is identical to the `redis` scaler.

### Checklist

Fixes #

### DIFF
--- a/content/docs/2.12/scalers/redis-cluster-lists.md
+++ b/content/docs/2.12/scalers/redis-cluster-lists.md
@@ -22,6 +22,7 @@ triggers:
     activationListLength: "5" # optional
     enableTLS: "false" # optional
     unsafeSsl: "false" # optional
+    databaseIndex: "0" # optional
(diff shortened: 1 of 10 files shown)


JorTurFer:
Nice!
You're right, all the redis list/stream scalers use the same library under the hood, so the behaviour should be the same. Could you update other redis docs with this change too?

peschmae:
I just realised that there are mutliple version directories for the docs.
I've updated 2.12 in the previous commit. Should I also do 2.13 or are some changes synced between the versions?

peschmae:
I've squashed all doc changes for 2.12 into a single commit, and made the same changes to the docs for 2.13. 

I hope it's fine like this

JorTurFer (approved):
LGTM! thanks for the improvement ❤️

PR #1251, opened 2023-10-18
Add documentation about new AWS auth

JorTurFer:
### Checklist

Related to https://github.com/kedacore/keda/issues/4134

### DIFF
--- a/content/docs/2.13/authentication-providers/aws-eks.md
+++ b/content/docs/2.13/authentication-providers/aws-eks.md
@@ -4,6 +4,8 @@ title = "EKS Pod Identity Webhook for AWS"
 
 [**EKS Pod Identity Webhook**](https://github.com/aws/amazon-eks-pod-identity-webhook), which is described more in depth [here](https://aws.amazon.com/blogs/opensource/introducing-fine-grained-iam-roles-service-accounts/), allows you to provide the role name using an annotation on a service account associated with your pod.
 
+> ⚠️ **WARNING:** `aws-eks` auth has been deprecated and support for it will be removed from KEDA on v3. We strongly encourage the migration to [`aws` auth](./aws.md).
+
 You can tell KEDA to use EKS Pod Identity Webhook via `podIdentity.provider`.
 
 ```yaml
--- a/content/docs/2.13/authentication-providers/aws-kiam.md
+++ b/content/docs/2.13/authentication-providers/aws-kiam.md
@@ -4,6 +4,8 @@ title = "Kiam Pod Identity for AWS"
 
 [**Kiam**](https://github.com/uswitch/kiam/) lets you bind an AWS IAM Role to a pod using an annotation on the pod.
 
+> ⚠️ **WARNING:** [AWS KIAM was abandoned](https://github.com/uswitch/kiam/#-%EF%B8%8Fthis-project-is-now-being-abandoned-%EF%B8%8F-) and support for it will be removed from KEDA on v2.15. We strongly encourage the migration to [`aws` auth](./aws.md).
+
 You can tell KEDA to use Kiam via `podIdentity.provider`.
 
 ```yaml
--- a/content/docs/2.13/authentication-providers/aws.md
+++ b/content/docs/2.13/authentication-providers/aws.md
@@ -1,0 +1,78 @@
++++
+title = "Pod Identity Webhook for AWS"
++++
+
+[**AWS IRSA Pod Identity Webhook**](https://github.com/aws/amazon-eks-pod-identity-webhook), which is described more in depth [here](https://aws.amazon.com/blogs/opensource/introducing-fine-grained-iam-roles-service-accounts/), allows you to provide the role name using an annotation on a service account associated with your pod.
+
+You can tell KEDA to use AWS Pod Identity Webhook via `podIdentity.provider`.
+
+```yaml
+podIdentity:
+  provider: aws # Optional. Default: none
+  roleArn: <role-arn|workload> # Optional. Default: RoleArn from annotation on service-account.
+```
+
+AWS IRSA will give access to pods with service accounts having appropriate annotations. Refer
+to these [docs](https://aws.amazon.com/es/blogs/opensource/introducing-fine-grained-iam-roles-service-accounts/) for more information. You can set these annotations on the KEDA Operator service account. This can be done for you during deployment with Helm with the following flags: 
+
+1. `--set podIdentity.aws.irsa.enabled=true`
+2. `--set podIdentity.aws.irsa.roleArn={aws-arn-role}`
+
+You can override the role that was assigned to KEDA during installation, by specifying an `roleArn` parameter under the `podIdentity` field. This allows end-users to use different roles to access various resources which is more secure than using a single identity that has access to multiple resources. 
+To reduce the managing overhead, `podIdentity.roleArn` can be set with the value `workload` and KEDA will check the workload service account to check if IRSA annotation is there and KEDA will assume that role.
+
+## Setting up KEDA role and policy
+
+How to do a basic configuration of IRSA role [is explained in official docs](https://aws.amazon.com/es/blogs/opensource/introducing-fine-grained-iam-roles-service-accounts/) but the policy changes depending if you are using KEDA role (`podIdentity.roleArn` is not set) or workload role (`podIdentity.roleArn` sets a RoleArn or `workload`).
+
+### Using KEDA role to access infrastructure
+
+This is the easiest case and you just need to attach to KEDA's role the desired policy/policies, granting the access permissions that you want to provide.
+
+
+### Using KEDA role to assume workload role
+
+In this case, KEDA will use its own role to assume the workload role (and to use workload's role attached policies). This scenario is a bit more complex because we need to establish a trusted relationship between both roles and we need to grant to KEDA's role the permission to assume other roles.
+
+This is how KEDA's role policy could look:
+
+```json
+{
+  "Version": "2012-10-17",
+  "Statement": [
+    {
+      "Effect": "Allow",
+      "Action": "sts:AssumeRole",
+      "Resource": [
+        "arn:aws:iam::ACCOUNT:role/ROLE_NAME"
+      ]
+    }
+  ]
+}
+```
+
+This policy attached to KEDA's role will allow KEDA to assume other roles, now you have to allow other roles to being assumed by KEDA's role. To achieve this, you have to add a trusted relation to the workload role:
+
+```json
+{
+  "Version": "2012-10-17",
+  "Statement": [
+    {
+      // Your already existing relations
+      "Sid": "",
+      "Effect": "Allow",
+      // ...
+    },
+    {
+      "Sid": "",
+      "Effect": "Allow",
+      "Principal": {
+        "AWS": "arn:aws:iam::ACCOUNT:role/KEDA_ROLE_NAME"
+      },
+      "Action": "sts:AssumeRole"
+    }
+  ]
+}
+```
+
+After these actions, you are ready to use KEDA's role to assume the workload role, providing it explictitly `podIdentity.roleArn: AWS_ROLE_AWS` or using the workload role (if the workload has IRSA enabled) `podIdentity.roleArn: workload`.


content/docs/2.13/authentication-providers/aws.md
```diff
@@ -0,0 +1,78 @@
++++
+title = "Pod Identity Webhook for AWS"
++++
+
+[**AWS IRSA Pod Identity Webhook**](https://github.com/aws/amazon-eks-pod-identity-webhook), which is described more in depth [here](https://aws.amazon.com/blogs/opensource/introducing-fine-grained-iam-roles-service-accounts/), allows you to provide the role name using an annotation on a service account associated with your pod.
```
tomkerkhove, line 5: Let's be clear on what IRSA stands for
```suggestion
[**AWS IAM Roles for Service Accounts (IRSA) Pod Identity Webhook**](https://github.com/aws/amazon-eks-pod-identity-webhook) ([details](https://aws.amazon.com/blogs/opensource/introducing-fine-grained-iam-roles-service-accounts/)) allows you to provide the role name using an annotation on a service account associated with your pod.
```

content/docs/2.13/authentication-providers/aws.md
```diff
@@ -0,0 +1,78 @@
...
+
+You can tell KEDA to use AWS Pod Identity Webhook via `podIdentity.provider`.
+
+```yaml
+podIdentity:
+  provider: aws # Optional. Default: none
+  roleArn: <role-arn|workload> # Optional. Default: RoleArn from annotation on service-account.
```
tomkerkhove, line 12: What does roleArn mean? Given this is also an option, I think we might need to consider renaming it to `type`?

content/docs/2.13/authentication-providers/aws.md
```diff
@@ -0,0 +1,78 @@
++++
+title = "Pod Identity Webhook for AWS"
```
tomkerkhove, line 2:
```suggestion
title = "Pod Identity Webhook for AWS by using IAM Roles for Service Accounts (IRSA)"
```

content/docs/2.13/authentication-providers/aws.md
```diff
@@ -0,0 +1,78 @@
...
+You can tell KEDA to use AWS Pod Identity Webhook via `podIdentity.provider`.
+
+```yaml
+podIdentity:
+  provider: aws # Optional. Default: none
+  roleArn: <role-arn|workload> # Optional. Default: RoleArn from annotation on service-account.
+```
+
+AWS IRSA will give access to pods with service accounts having appropriate annotations. Refer
+to these [docs](https://aws.amazon.com/es/blogs/opensource/introducing-fine-grained-iam-roles-service-accounts/) for more information. You can set these annotations on the KEDA Operator service account. This can be done for you during deployment with Helm with the following flags:
```
tomkerkhove, line 16:
```suggestion
AWS IRSA will give access to pods with service accounts having appropriate annotations. ([official docs](https://aws.amazon.com/es/blogs/opensource/introducing-fine-grained-iam-roles-service-accounts/)) You can set these annotations on the KEDA Operator service account.

This can be done for you during deployment with Helm with the following flags: 
```

content/docs/2.13/authentication-providers/aws.md
```diff
@@ -0,0 +1,78 @@
...
+```
+
+AWS IRSA will give access to pods with service accounts having appropriate annotations. Refer
+to these [docs](https://aws.amazon.com/es/blogs/opensource/introducing-fine-grained-iam-roles-service-accounts/) for more information. You can set these annotations on the KEDA Operator service account. This can be done for you during deployment with Helm with the following flags: 
+
+1. `--set podIdentity.aws.irsa.enabled=true`
+2. `--set podIdentity.aws.irsa.roleArn={aws-arn-role}`
+
+You can override the role that was assigned to KEDA during installation, by specifying an `roleArn` parameter under the `podIdentity` field. This allows end-users to use different roles to access various resources which is more secure than using a single identity that has access to multiple resources. 
+To reduce the managing overhead, `podIdentity.roleArn` can be set with the value `workload` and KEDA will check the workload service account to check if IRSA annotation is there and KEDA will assume that role.
```
tomkerkhove, line 22: New line for better rendering to make it more digestable
image


```suggestion
You can override the role that was assigned to KEDA during installation, by specifying an `roleArn` parameter under the `podIdentity` field. This allows end-users to use different roles to access various resources which is more secure than using a single identity that has access to multiple resources.

To reduce the managing overhead, `podIdentity.roleArn` can be set with the value `workload` and KEDA will check the workload service account to check if IRSA annotation is there and KEDA will assume that role.
```

content/docs/2.13/authentication-providers/aws.md
```diff
@@ -0,0 +1,78 @@
...
+
+## Setting up KEDA role and policy
+
+How to do a basic configuration of IRSA role [is explained in official docs](https://aws.amazon.com/es/blogs/opensource/introducing-fine-grained-iam-roles-service-accounts/) but the policy changes depending if you are using KEDA role (`podIdentity.roleArn` is not set) or workload role (`podIdentity.roleArn` sets a RoleArn or `workload`).
+
+### Using KEDA role to access infrastructure
+
+This is the easiest case and you just need to attach to KEDA's role the desired policy/policies, granting the access permissions that you want to provide.
+
```
tomkerkhove, line 31: Nit
```suggestion
```

content/docs/2.13/authentication-providers/aws.md
```diff
@@ -0,0 +1,78 @@
...
+1. `--set podIdentity.aws.irsa.enabled=true`
+2. `--set podIdentity.aws.irsa.roleArn={aws-arn-role}`
+
+You can override the role that was assigned to KEDA during installation, by specifying an `roleArn` parameter under the `podIdentity` field. This allows end-users to use different roles to access various resources which is more secure than using a single identity that has access to multiple resources. 
+To reduce the managing overhead, `podIdentity.roleArn` can be set with the value `workload` and KEDA will check the workload service account to check if IRSA annotation is there and KEDA will assume that role.
+
+## Setting up KEDA role and policy
+
+How to do a basic configuration of IRSA role [is explained in official docs](https://aws.amazon.com/es/blogs/opensource/introducing-fine-grained-iam-roles-service-accounts/) but the policy changes depending if you are using KEDA role (`podIdentity.roleArn` is not set) or workload role (`podIdentity.roleArn` sets a RoleArn or `workload`).
```
tomkerkhove, line 26:
```suggestion
The [official docs](https://aws.amazon.com/es/blogs/opensource/introducing-fine-grained-iam-roles-service-accounts/) explain how to set up a a basic configuration for an IRSA role but the policy changes depending if you are using the KEDA role (`podIdentity.roleArn` is not set) or workload role (`podIdentity.roleArn` sets a RoleArn or `workload`).
```

But after reading it, I am not sure I get the difference

content/docs/2.13/authentication-providers/aws.md
```diff
@@ -0,0 +1,78 @@
...
+
+You can tell KEDA to use AWS Pod Identity Webhook via `podIdentity.provider`.
+
+```yaml
+podIdentity:
+  provider: aws # Optional. Default: none
+  roleArn: <role-arn|workload> # Optional. Default: RoleArn from annotation on service-account.
+```
+
+AWS IRSA will give access to pods with service accounts having appropriate annotations. Refer
```
tomkerkhove, line 15: I think we should clarify (better) with the different roleArn options are

content/docs/2.13/authentication-providers/aws.md
JorTurFer: This is related with how AWS IRSA works. In Azure, when a user overrides the clientID, KEDA uses the new clientID, and the MSI must be federated too with KEDA's service account. In AWS in opposition, if you want to use other role (Azure's MSI), you use KEDA's role to assume the other role. Instead of federating both roles with KEDA's service account, you only have to federate KEDA's role and KEDA will use always its role, using the attached policy (permissions) to access resources or to operate in representation of other role assuming it (and using other role policies, not KEDA's permissions).

Is it clear now?

content/docs/2.13/authentication-providers/aws.md
blakepettersson: An option could be to have a separate `type` field with `roleArn|workload` (with the default set to `roleArn`). If `workload` is set, `roleArn` is ignored.

content/docs/2.13/authentication-providers/aws.md
```diff
@@ -0,0 +1,79 @@
...
++++
+
+[**AWS IAM Roles for Service Accounts (IRSA) Pod Identity Webhook**](https://github.com/aws/amazon-eks-pod-identity-webhook) ([details](https://aws.amazon.com/blogs/opensource/introducing-fine-grained-iam-roles-service-accounts/)) allows you to provide the role name using an annotation on a service account associated with your pod.
+
+You can tell KEDA to use AWS Pod Identity Webhook via `podIdentity.provider`.
+
+```yaml
+podIdentity:
+  provider: aws # Optional. Default: none
+  roleArn: <role-arn|workload> # Optional. Default: RoleArn from annotation on service-account.
```
blakepettersson, line 12: >Default: RoleArn from annotation on service-account.

This would imply that the default is `workload` right?

content/docs/2.13/authentication-providers/aws.md
tomkerkhove: That is what I was thinking as well

content/docs/2.13/authentication-providers/aws.md
```diff
@@ -0,0 +1,79 @@
...
+
+### Using KEDA role to access infrastructure
+
+This is the easiest case and you just need to attach to KEDA's role the desired policy/policies, granting the access permissions that you want to provide.
+
+### Using KEDA role to assume workload role
+
+In this case, KEDA will use its own role to assume the workload role (and to use workload's role attached policies). This scenario is a bit more complex because we need to establish a trusted relationship between both roles and we need to grant to KEDA's role the permission to assume other roles.
+
+This is how KEDA's role policy could look:
```
blakepettersson, line 38:
```suggestion
This is an example of how KEDA's role policy could look like:
```

content/docs/2.13/authentication-providers/aws.md
```diff
@@ -0,0 +1,79 @@
...
+      "Effect": "Allow",
+      "Action": "sts:AssumeRole",
+      "Resource": [
+        "arn:aws:iam::ACCOUNT:role/ROLE_NAME"
+      ]
+    }
+  ]
+}
+```
+
```
blakepettersson, line 54: I think it would also be good to show a use case that the KEDA operator can be allowed to assume multiple roles, something like 

```suggestion
This can be extended so that KEDA can assume multiple workload roles, either as an explicit array of role ARNs, or with a wildcard.
```

content/docs/2.13/authentication-providers/aws.md
```diff
@@ -0,0 +1,79 @@
...
+This policy attached to KEDA's role will allow KEDA to assume other roles, now you have to allow other roles to being assumed by KEDA's role. To achieve this, you have to add a trusted relation to the workload role:
```
blakepettersson, line 55:
```suggestion
This policy attached to KEDA's role will allow KEDA to assume other roles, now you have to allow the workload roles you want to use all allow to being assumed by the KEDA operator role. To achieve this, you have to add a trusted relation to the workload role:
```

blakepettersson, line 55:
```suggestion
This policy attached to KEDA's role will allow KEDA to assume other roles, now you have to allow the workload roles you want to use all allow to being assumed by the KEDA operator role. To achieve this, you have to add a trusted relation to the workload role:
```

blakepettersson (commented):
Lookinh

blakepettersson:
Another thing, for all AWS scalers it would be good to clarify that all of the `roleArns` on said scalers are deprecated and should not be used (or perhaps even an error if used with the `aws` TriggerAuthentication?)

content/docs/2.13/authentication-providers/aws.md
```diff
@@ -0,0 +1,79 @@
...
+AWS IRSA will give access to pods with service accounts having appropriate annotations. ([official docs](https://aws.amazon.com/es/blogs/opensource/introducing-fine-grained-iam-roles-service-accounts/)) You can set these annotations on the KEDA Operator service account.
+
+This can be done for you during deployment with Helm with the following flags: 
+
+1. `--set podIdentity.aws.irsa.enabled=true`
+2. `--set podIdentity.aws.irsa.roleArn={aws-arn-role}`
+
+You can override the role that was assigned to KEDA during installation, by specifying an `roleArn` parameter under the `podIdentity` field. This allows end-users to use different roles to access various resources which is more secure than using a single identity that has access to multiple resources.
+
+To reduce the managing overhead, `podIdentity.roleArn` can be set with the value `workload` and KEDA will check the workload service account to check if IRSA annotation is there and KEDA will assume that role.
```
blakepettersson, line 24: I'm not sure having IRSA on both the KEDA operator and KEDA workload is _less_ overhead 😄 

```suggestion
If you would like to use the same IAM credentials as your workload is currently using, `podIdentity.roleArn` can be set with the value `workload` and KEDA will check the workload service account to check if IRSA annotation is there and KEDA will assume that role.
```

content/docs/2.13/authentication-providers/aws.md
blakepettersson: There is only really one when using IRSA, the gist of it being that `eks.amazonaws.com/role-arn` should be set as an annotation on the service account which requires access to AWS services.

I added a HOWTO in argoproj/argo-cd#14187, perhaps a sequence of steps clarifying how to setup IRSA + KEDA would be useful here as well?

content/docs/2.13/authentication-providers/aws.md
zroubalik: agree with the suggestion

content/docs/2.13/authentication-providers/aws.md
zroubalik:
```suggestion
If you would like to use the same IAM credentials as your workload is currently using, `podIdentity.roleArn` can be set with the value `workload` and KEDA will inspect the workload service account to check if IRSA annotation is there and KEDA will assume that role.
```
just a nit, so we don't have "check" twice in the sentece :)

content/docs/2.13/authentication-providers/aws.md
```diff
@@ -0,0 +1,79 @@
...
+
+## Setting up KEDA role and policy
+
+The [official docs](https://aws.amazon.com/es/blogs/opensource/introducing-fine-grained-iam-roles-service-accounts/) explain how to set up a a basic configuration for an IRSA role but the policy changes depending if you are using the KEDA role (`podIdentity.roleArn` is not set) or workload role (`podIdentity.roleArn` sets a RoleArn or `workload`).
```
zroubalik, line 28:
```suggestion
The [official AWS docs](https://aws.amazon.com/es/blogs/opensource/introducing-fine-grained-iam-roles-service-accounts/) explain how to set up a a basic configuration for an IRSA role. The policy changes depend if you are using the KEDA role (`podIdentity.roleArn` is not set) or workload role (`podIdentity.roleArn` sets a RoleArn or `workload`).
```

content/docs/2.13/authentication-providers/aws.md
```diff
@@ -0,0 +1,79 @@
...
+
+### Using KEDA role to access infrastructure
+
+This is the easiest case and you just need to attach to KEDA's role the desired policy/policies, granting the access permissions that you want to provide.
```
zroubalik, line 32: Would be great to show an example, or a part of the config.

content/docs/2.13/authentication-providers/aws.md
zroubalik: +1000

content/docs/2.13/authentication-providers/aws.md
JorTurFer: oh, I misunderstood it :/ 
What about something like `useWorkloadIdentity`? I'm thinking that using workload identity can be ported to Azure podIdentity and it can be useful, so I prefer to not couple it to AWS. (or maybe `useTargetRefIdentity` or so) 
I mean, we could have 2 fields, one `roleArn|clientId` (AWS and Azure respectively) and other `useWorkloadIdentity`.
If `useWorkloadIdentity: true`, we will ignore the other and if it's false, we will take into account the other.
WDYT?

content/docs/2.13/authentication-providers/aws.md
JorTurFer: No no, default is KEDA's role

content/docs/2.13/authentication-providers/aws.md
JorTurFer: I'll check it tomorrow for adding it

content/docs/2.13/authentication-providers/aws.md
JorTurFer: What do you mean, an example about how to grant permissions over a resource?

content/docs/2.13/authentication-providers/aws.md
```diff
@@ -0,0 +1,79 @@
...
+  "Statement": [
+    {
+      "Effect": "Allow",
+      "Action": "sts:AssumeRole",
+      "Resource": [
+        "arn:aws:iam::ACCOUNT:role/ROLE_NAME"
+      ]
+    }
+  ]
+}
```
JorTurFer, line 52:
```suggestion
{
  "Version": "2012-10-17",
  "Statement": [
    {
      "Effect": "Allow",
      "Action": "sts:AssumeRole",
      "Resource": [
        "arn:aws:iam::ACCOUNT_1:role/ROLE_NAME"
      ]
    },
    {
      "Effect": "Allow",
      "Action": "sts:AssumeRole",
      "Resource": [
        "arn:aws:iam::ACCOUNT_2:role/*"
      ]
    }
  ]
}
```

content/docs/2.13/authentication-providers/aws.md
JorTurFer: I have added a second item on the policy to adding an example of wildcard, is it okey now? I can add another example explicitly setting it with a wildcard instead, if you think it's better

JorTurFer:
> Another thing, for all AWS scalers it would be good to clarify that all of the roleArns on said scalers are deprecated and should not be used (or perhaps even an error if used with the aws TriggerAuthentication?)

This is a good point and I missed it, tomorrow I'll add a note on them

content/docs/2.13/authentication-providers/aws.md
tomkerkhove: useWorkloadIdentity removes our flexibility in case there is a new thing in the future that superceeds workload identity.

I'm also not sure how Azure comes in to play here for an AWS trigger auth resource?

content/docs/2.13/authentication-providers/aws.md
blakepettersson: Then it's probably more clear to say something like

```suggestion
  roleArn: <role-arn|workload> # Optional. If not set the IAM role which the KEDA operator uses will be used
```

content/docs/2.13/authentication-providers/aws.md
zroubalik: yeah just a suggestion, basically that we have the specific example for each approach. Disregard, if it doesn't make sense to you.

content/docs/2.13/authentication-providers/aws.md
JorTurFer: Because the spec is shared as it's part of triggerauthentication, so I prefer to use general terms wherever is possible. I mean, `clientId` or `roleArn` are vendor locked, but `provider` is generic. 
If I could use a generic term for using the workload assigned identity, we can use it for all the cloud providers and for user operation is easier IMHO (but I can be overthinking the things)

content/docs/2.13/authentication-providers/aws.md
JorTurFer: yeah, I can add it, np

content/docs/2.13/authentication-providers/aws.md
blakepettersson: I think the issue with having something "vendor neutral" for this is that it's not terminology that's familiar to someone not already steeped in it. E.g "workload identity" is not something that's immediately apparent to someone who has not used GKE or AKS. To be fair, EKS is the odd one out for calling it IRSA instead of Workload Identity, but still...

content/docs/2.13/authentication-providers/aws.md
blakepettersson: LGTM 👍

content/docs/2.13/authentication-providers/aws.md
tomkerkhove: I think we should standardize as much as we can, but still provide the best end-user experience which means removing cognitive overload and just call it however the provider calls it.

We can use other names internally but I think we should just call it whatever the vendor calls it. A lot of people using "managed identities" don't know the background of how it works or what the project is called, they just want it to work

That's why I think the introduction of `provider: aws` is good! But I'm not a fan of `roleArn: <role-arn|workload>`

JorTurFer: > But I'm not a fan of roleArn: <role-arn|workload>

roleArn is part of the vendor naming (it's the AWS' for managed identity), so we should use it. My question is, having this scenario, how would you organize the trigger?

- we allow to use the already set roleArn during KEDA installation
- we allow to override the roleArn directly using the TriggerAuthenticationm
- we allow to override the roleArn, getting the value from the scaled workload

Currently, TriggerAuthentication (for azure) allows only the first 2 options, but current AWS auth (getting info also from scaler metadata) supports all of them, and we should support all of them IMO, and eventually allow this extra approach on azure side. 
That's why I'm trying to find the correct way to name the option of using the TrigerAuthentication provide value or the workloads's value. 
Currently, we are already using `identityOwner` as part of AWS Scalers' spec, so maybe we can introduce this concept to the TriggerAuthentication, something like that

JorTurFer: I think that this will be the most painful point of the new feature... xD

Maybe we could schedule a call for discussing it, or talk about this during the next community standup.

blakepettersson: When is the next community standup?

JorTurFer: November 7th at 5PM CEST
You have all the info here: https://keda.sh/community/

JorTurFer: I've created a new field `identityOwner` with 3 valid values: `keda`, `workload` or unset (same as `keda`) to choose this option

JorTurFer:
PTAL @zroubalik , @tomkerkhove  @blakepettersson  @thasami
I've been working on this feature and it's ready to review. I've addressed all the comments here and also in the issue (or at least I hope so)

content/docs/2.13/authentication-providers/aws.md
JorTurFer: done

content/docs/2.13/authentication-providers/aws.md
JorTurFer: done

content/docs/2.13/authentication-providers/aws.md
```diff
@@ -0,0 +1,141 @@
...
+```yaml
+podIdentity:
+  provider: aws
+  roleArn: <role-arn> # Optional. 
+  identityOwner: keda|workload # Optional. If not set, 'keda' is default value. Mutually exclusive with 'roleArn' (if set)
+```
+
+**Parameter list:**
+
+- `roleArn` - Role ARN no be used by KEDA. If not set the IAM role which the KEDA operator uses will be used. Mutually exclusive with 'identityOwner: workload'
```
tomkerkhove, line 18: What does "Role ARN no" mean?

content/docs/2.13/authentication-providers/aws.md
```diff
@@ -0,0 +1,141 @@
...
+
+[**AWS IAM Roles for Service Accounts (IRSA) Pod Identity Webhook**](https://github.com/aws/amazon-eks-pod-identity-webhook) ([details](https://aws.amazon.com/blogs/opensource/introducing-fine-grained-iam-roles-service-accounts/)) allows you to provide the role name using an annotation on a service account associated with your pod.
+
+You can tell KEDA to use AWS Pod Identity Webhook via `podIdentity.provider`.
+
+```yaml
+podIdentity:
+  provider: aws
+  roleArn: <role-arn> # Optional. 
+  identityOwner: keda|workload # Optional. If not set, 'keda' is default value. Mutually exclusive with 'roleArn' (if set)
```
tomkerkhove, line 13: Is this for non-role ARN scenarios (and thus pod identity?)? Based on the doc it is not clear so maybe we need a "this has 2 auth options and this is how they are different" section

tomkerkhove:
Can you please remind me why 1 auth provider is removed in KEDA v2.15 and the other one in KEDA v3?

content/docs/2.13/authentication-providers/aws.md
JorTurFer: `Role ARN no` it's a typo, it should have been `Role ARN to`. ARN means Amazon Resource Names and it's their naming xD

JorTurFer:
```suggestion
- `roleArn` - Role ARN to be used by KEDA. If not set the IAM role which the KEDA operator uses will be used. Mutually exclusive with 'identityOwner: workload'
```

JorTurFer:
> Can you please remind me why 1 auth provider is removed in KEDA v2.15 and the other one in KEDA v3?

It's because KIAM is deprecated, so we can remove `aws-kiam` after the proper announce (and 2 versiones) and IRSA still works (`aws-eks` uses IRSA), but we decided to create another auth (although it uses IRSA too) to redesign the current approach, unifying the concept with Azure/GCP approaches instead of current approach.  This is because `aws-eks` uses information from TriggerAuthentication but also from trigger.metadata for authentication stuff (for example, if you want to override the roleArn, `aws-eks` takes it from trigger.metadata instead of TriggerAuthentication)
As `aws-eks` doesn't use a deprecated technology, we decided to keep it available until KEDA v3, where we can introduce breaking changes with freedom.

content/docs/2.13/authentication-providers/aws.md
JorTurFer: roleARN is like clientId in Azure, so KEDA always uses roleArn, but there are 3 different options for providing it:
- Not setting `roleArn` nor `identityOwner` or `identityOwner: keda`: KEDA will use its own role
- Setting `roleArn`: KEDA will use provided roleArn
- Setting `identityOwner: workload`: KEDA will inspect workload service account for getting its role and will use it

Maybe, instead of `identityOwner: workload` we can just use something like `useWorkloadIdentity` and that's all. Maybe it's more clear. My afraid is that `useWorkloadIdentity` can be even more confusing, as WorkloadIdentity has its own meaning in Azure and GCP. 

Currently, we don't support this feature (`identityOwner`) in azure pod identity, but I think that it's something interesting that we can support too

content/docs/2.13/authentication-providers/aws-eks.md
```diff
@@ -4,6 +4,8 @@ title = "EKS Pod Identity Webhook for AWS"
 
 [**EKS Pod Identity Webhook**](https://github.com/aws/amazon-eks-pod-identity-webhook), which is described more in depth [here](https://aws.amazon.com/blogs/opensource/introducing-fine-grained-iam-roles-service-accounts/), allows you to provide the role name using an annotation on a service account associated with your pod.
 
+> ⚠️ **WARNING:** `aws-eks` auth has been deprecated and support for it will be removed from KEDA on v3. We strongly encourage the migration to [`aws` auth](./aws.md).
```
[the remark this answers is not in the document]
JorTurFer: There isn't any discussion because we didn't merge the feature yet, IIRC the discussion has to be created during the release process, hasn't?
I've already created the issues, I can link them (or I can create the discussion now and add them)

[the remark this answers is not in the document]
JorTurFer:
```suggestion
> ⚠️ **WARNING:** [`aws-eks` auth has been deprecated](https://github.com/kedacore/keda/issues/5086) and support for it will be removed from KEDA on v3. We strongly encourage the migration to [`aws` auth](./aws.md).
```

content/docs/2.13/authentication-providers/aws-kiam.md
```diff
@@ -4,6 +4,8 @@ title = "Kiam Pod Identity for AWS"
 
 [**Kiam**](https://github.com/uswitch/kiam/) lets you bind an AWS IAM Role to a pod using an annotation on the pod.
 
+> ⚠️ **WARNING:** [AWS KIAM was abandoned](https://github.com/uswitch/kiam/#-%EF%B8%8Fthis-project-is-now-being-abandoned-%EF%B8%8F-) and support for it will be removed from KEDA on v2.15. We strongly encourage the migration to [`aws` auth](./aws.md).
```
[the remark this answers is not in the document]
JorTurFer:
```suggestion
> ⚠️ **WARNING:** `aws-kiam` auth has been deprecated given [AWS KIAM is no longer maintained](https://github.com/uswitch/kiam/#-%EF%B8%8Fthis-project-is-now-being-abandoned-%EF%B8%8F-). As a result, [support for it will be removed from KEDA on v2.15](https://github.com/kedacore/keda/issues/5085). We strongly encourage the migration to [`aws` auth](./aws.md).
```

content/docs/2.13/authentication-providers/aws-eks.md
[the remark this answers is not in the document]
tomkerkhove: > There isn't any discussion because we didn't merge the feature yet, IIRC the discussion has to be created during the release process, hasn't?

Not specifically, as part of the deprecation as that is sooner

content/docs/2.13/authentication-providers/aws-kiam.md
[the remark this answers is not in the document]
tomkerkhove: See above, let's use discussion instead

content/docs/2.13/authentication-providers/aws.md
tomkerkhove: Oh OK then it makes sense :)

content/docs/2.13/authentication-providers/aws-eks.md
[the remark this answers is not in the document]
JorTurFer: Okey, I miss understood the process and I haven't created them yet, only the issues.
Let me create them too and update the links

content/docs/2.13/authentication-providers/aws.md
JorTurFer: Yeah, we talked about adding a new field instead of using roleArn field for both. We suggested 'type', but that can be confusing too... ( because we talked about using it as enum too, like this)
That's why I used identityOwner as it's the current name in AWS scalers. It's agnóstico from the vendor, it already in use for aws, and can be exaplained for other vendors

content/docs/2.13/authentication-providers/aws-kiam.md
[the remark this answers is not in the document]
JorTurFer: yeah

content/docs/2.13/authentication-providers/aws-eks.md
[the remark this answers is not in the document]
JorTurFer: done

content/docs/2.13/authentication-providers/aws-kiam.md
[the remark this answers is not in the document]
JorTurFer: done

content/docs/2.13/authentication-providers/aws.md
```diff
@@ -0,0 +1,140 @@
++++
+title = "Pod Identity Webhook for AWS by using IAM Roles for Service Accounts (IRSA)"
```
zroubalik, line 2:
```suggestion
title = "EKS Pod Identity Webhook for AWS by using IAM Roles for Service Accounts (IRSA)"
```

Just a nit to group all AWS/EKS related auth providers together

content/docs/2.13/authentication-providers/aws-kiam.md
```diff
@@ -4,6 +4,8 @@ title = "Kiam Pod Identity for AWS"
 
 [**Kiam**](https://github.com/uswitch/kiam/) lets you bind an AWS IAM Role to a pod using an annotation on the pod.
 
+> ⚠️ **WARNING:** `aws-kiam` auth has been deprecated given [AWS KIAM is no longer maintained](https://github.com/uswitch/kiam/#-%EF%B8%8Fthis-project-is-now-being-abandoned-%EF%B8%8F-). As a result, [support for it will be removed from KEDA on v2.15](https://github.com/kedacore/keda/discussions/5342). We strongly encourage the migration to [`aws` auth](./aws.md).
```
zroubalik, line 7: Please fix the title in this file to contain EKS prefix:
```yaml
EKS Kiam Pod Identity for AWS
```

Just a nit to group all AWS/EKS related auth providers together

content/docs/2.13/authentication-providers/aws.md
```diff
@@ -0,0 +1,140 @@
...
++++
+
+[**AWS IAM Roles for Service Accounts (IRSA) Pod Identity Webhook**](https://github.com/aws/amazon-eks-pod-identity-webhook) ([details](https://aws.amazon.com/blogs/opensource/introducing-fine-grained-iam-roles-service-accounts/)) allows you to provide the role name using an annotation on a service account associated with your pod.
+
+You can tell KEDA to use AWS Pod Identity Webhook via `podIdentity.provider`.
```
zroubalik, line 7: Might be worth adding a line on where should users set this (which resource).

content/docs/2.13/authentication-providers/aws.md
```diff
@@ -0,0 +1,140 @@
...
+```yaml
+podIdentity:
+  provider: aws
+  roleArn: <role-arn> # Optional. 
+  identityOwner: keda|workload # Optional. If not set, 'keda' is default value. Mutually exclusive with 'roleArn' (if set)
+```
+
+**Parameter list:**
+
+- `roleArn` - Role ARN to be used by KEDA. If not set the IAM role which the KEDA operator uses will be used. Mutually exclusive with 'identityOwner: workload'
```
zroubalik, line 18:
```suggestion
- `roleArn` - Role ARN to be used by KEDA. If not set, the IAM role which the KEDA operator uses will be used. Mutually exclusive with `identityOwner: workload`
```

content/docs/2.13/authentication-providers/aws.md
```diff
@@ -0,0 +1,140 @@
++++
+title = "Pod Identity Webhook for AWS by using IAM Roles for Service Accounts (IRSA)"
++++
+
+[**AWS IAM Roles for Service Accounts (IRSA) Pod Identity Webhook**](https://github.com/aws/amazon-eks-pod-identity-webhook) ([details](https://aws.amazon.com/blogs/opensource/introducing-fine-grained-iam-roles-service-accounts/)) allows you to provide the role name using an annotation on a service account associated with your pod.
```
zroubalik, line 5:
```suggestion
[**AWS IAM Roles for Service Accounts (IRSA) Pod Identity Webhook**](https://github.com/aws/amazon-eks-pod-identity-webhook) ([documentation](https://aws.amazon.com/blogs/opensource/introducing-fine-grained-iam-roles-service-accounts/)) allows you to provide the role name using an annotation on a service account associated with your pod.
```

content/docs/2.13/authentication-providers/aws.md
zroubalik: Theoretically, can there be a different `identityOwner` than `keda` or `workload`? If not, then mabye the boolean is a better option?
I'd be okay with both options though.

content/docs/2.13/authentication-providers/aws.md
```diff
@@ -0,0 +1,140 @@
...
+
+If you would like to use the same IAM credentials as your workload is currently using, `podIdentity.identityOwner` can be set with the value `workload` and KEDA will inspect the workload service account to check if IRSA annotation is there and KEDA will assume that role.
+
+## AssumeRole or AssumeRoleWithWebIdentity?
+
+This authentication uses automatically both, doing a fallback from [AssumeRoleWithWebIdentity](https://docs.aws.amazon.com/STS/latest/APIReference/API_AssumeRoleWithWebIdentity.html) to [AssumeRole](https://docs.aws.amazon.com/STS/latest/APIReference/API_AssumeRole.html) if the first one fails. This extends the capabilities because KEDA doesn't need sts:AssumeRole permission if you are already working with [WebIdentities](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_providers_oidc.html), you just need to add KEDA service account to the trusted relations of the role.
```
zroubalik, line 38:
```suggestion
This authentication uses automatically both, doing a fallback from [AssumeRoleWithWebIdentity](https://docs.aws.amazon.com/STS/latest/APIReference/API_AssumeRoleWithWebIdentity.html) to [AssumeRole](https://docs.aws.amazon.com/STS/latest/APIReference/API_AssumeRole.html) if the first one fails. This extends the capabilities because KEDA doesn't need `sts:AssumeRole` permission if you are already working with [WebIdentities](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_providers_oidc.html), you just need to add KEDA service account to the trusted relations of the role.
```

zroubalik (commented):
Great job!

content/docs/2.13/authentication-providers/aws.md
JorTurFer: No, it can be `keda` or `workload` (and values are validated by OpenAPI schema). My problem here is how to name it in a clear and concise way because the "concept" of identity is attached to vendor, so useWorkloadIdentity can mean different things on different vendors. As AWS uses "roles", workload identity is quite clear, KEDA'll use the workload role, but for Azure or GCP where Workload Identity is the authentication way itself, this can be more confusing... Am I using Workload Identity (the authentication way) or the workload identity (retrieved using Workload identity authentication way). I'm open to suggestions with the naming and if we find a good naming for a boolean value, I'm more than happy changing it

JorTurFer: We should also take into account that by omission, KEDA has to be the default value, so useKedaIdentity sounds weird when I have to explicitly set it to false if I want to use podIdentity.roleArn (and podIdentity.clientId in the future) because they are mutually exclusive... 
Maybe something like `useScaledWorkloadIdentity` can be clear enough, but it's a huge name for a key...

content/docs/2.13/authentication-providers/aws.md
JorTurFer: This works outside EKS too, so I prefer to not include EKS prefix. Let me rename AWS auths from "XXXXX for AWS" to "AWS XXXX"

JorTurFer: Now they are grouped: 
image

content/docs/2.13/authentication-providers/aws.md
JorTurFer: I don't get your point, this is an authentication, so TriggerAuthentication is the place. I mean, we aren't doing it either for other auths

content/docs/2.13/authentication-providers/aws.md
JorTurFer: WDYT @zroubalik ?

content/docs/2.13/authentication-providers/aws.md
tomkerkhove: > As AWS uses "roles", workload identity is quite clear, KEDA'll use the workload role, but for Azure or GCP where Workload Identity is the authentication way itself, this can be more confusing... Am I using Workload Identity (the authentication way) or the workload identity (retrieved using Workload identity authentication way).

I honestly don't know if we should go to a common thing at the cost of less "optimized UX" per cloud provider. The majority of users will only use one cloud, so I'd prefer to make it dead-simple to configure auth for that cloud.

If the approach per cloud provider is different per provider, then it's up to us to clearly help in docs (or separate section) and clarify that.

> Maybe something like `useScaledWorkloadIdentity` can be clear enough, but it's a huge name for a key...

Size does not matter, but booleans limit flexibility in the future. I think the current `identityOwner` is OK and fairly clear.

tomkerkhove: > For KEDA v3, I think that we should refactor this to nested structs for each auth (IMHO it will improve user experience), but until that, I think that less fields are better if we can summarize and reuse them.

I don't know if that is a good way if it means we lose cloud-specific naming/concepts; end-users don't care about code re-use but do care about simplicity.

content/docs/2.13/authentication-providers/aws.md
tomkerkhove: Agreed

content/docs/2.13/authentication-providers/aws.md
tomkerkhove: I'm OK with the current approach, however, `identityOwner: workload` makes me expect it uses something from the workload but mostlikely it is the role ARN in the trigger spec? If so, this is confusing IMO and should be `autoscaling` or `autoscalingRule` or so imo

JorTurFer: `identityOwner: workload` means that KEDA will use the workload roleArn, inspecting the workload's service account looking for the roleArn annotation on it (not in trigger spec). 

> If so, this is confusing IMO and should be autoscaling or autoscalingRule or so imo

I don't get you, sorry :(

tomkerkhove: This is OK, :shipit:

tomkerkhove (approved):
:shipit:

PR #1259, opened 2023-11-14
use correct apiVersion for redis lists example

nadnerb:
Fix the documentation for Redis lists trigger to use the correct apiVersion. The version in the example does not work.

### Checklist


zroubalik (commented):
KEDA 1.4 used this apiVersion

PR #1262, opened 2023-11-21
Vault: Allow TriggerAuthentication Defaults via ENV variables

BojanZelic:
This PR sets vault defaults if specified via ENV variables. When configuring the operator, you can set ENV variables so users don't have to redefine vault settings every time they use vault authentication;

### Checklist

Fixes # https://github.com/kedacore/keda/issues/4965

Relates To: https://github.com/kedacore/keda/pull/5196

### DIFF
--- a/content/docs/2.13/authentication-providers/hashicorp-vault.md
+++ b/content/docs/2.13/authentication-providers/hashicorp-vault.md
@@ -13,14 +13,14 @@ You can pull one or more Hashicorp Vault secrets into the trigger by defining th
 
 ```yaml
 hashiCorpVault:                                               # Optional.
-  address: {hashicorp-vault-address}                          # Required.
-  namespace: {hashicorp-vault-namespace}                      # Optional. Default is root namespace. Useful for Vault Enterprise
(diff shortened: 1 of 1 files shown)


content/docs/2.13/authentication-providers/hashicorp-vault.md
```diff
@@ -13,14 +13,14 @@ You can pull one or more Hashicorp Vault secrets into the trigger by defining th
 
 ```yaml
 hashiCorpVault:                                               # Optional.
-  address: {hashicorp-vault-address}                          # Required.
-  namespace: {hashicorp-vault-namespace}                      # Optional. Default is root namespace. Useful for Vault Enterprise
-  authentication: token | kubernetes                          # Required.
-  role: {hashicorp-vault-role}                                # Optional.
-  mount: {hashicorp-vault-mount}                              # Optional.
+  address: {hashicorp-vault-address}                          # Optional. (env - $VAULT_ADDR)
```
tomkerkhove, line 16: @JorTurFer Is this the format we usually use?

PR #1266, opened 2023-11-27
Clarify that autoscaling.keda.sh/paused annotation value is not important. Only prescence is checked

nappelson:
Essentially, as a new Keda user I spent a long time trying to figure out why `autoscaling.keda.sh/paused: false` wasn't working as expected.

I opened this https://github.com/kedacore/keda/issues/5215 in `keda-core`. Until a resolution is decided there, I think this documentation update would be beneficial to avoid end-user communication.

### Checklist

Fixes #
Potentially fixes - https://github.com/kedacore/keda/issues/5215

### DIFF
--- a/content/docs/2.11/concepts/scaling-jobs.md
+++ b/content/docs/2.11/concepts/scaling-jobs.md
@@ -28,7 +28,7 @@ kind: ScaledJob
 metadata:
   name: {scaled-job-name}
   annotations:
-    autoscaling.keda.sh/paused: true          # Optional. Use to pause autoscaling of Jobs
+    autoscaling.keda.sh/paused: any-value          # Optional. Use to pause autoscaling of Jobs.
 spec:
   jobTargetRef:
     parallelism: 1                            # [max number of desired pods](https://kubernetes.io/docs/concepts/workloads/controllers/job/#controlling-parallelism)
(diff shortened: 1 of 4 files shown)


content/docs/2.11/concepts/scaling-jobs.md
```diff
@@ -28,7 +28,7 @@ kind: ScaledJob
 metadata:
   name: {scaled-job-name}
   annotations:
-    autoscaling.keda.sh/paused: true          # Optional. Use to pause autoscaling of Jobs
+    autoscaling.keda.sh/paused: any-value          # Optional. Use to pause autoscaling of Jobs.
```
nappelson, line 31:
```suggestion
    autoscaling.keda.sh/paused: any-value          # Optional. Use to pause autoscaling of Jobs
```

tomkerkhove:
Thanks! It is not important indeed, but I personally think the current example value still makes more sense though.

We can add the note you suggested, but I'd prefer to stick with `true`

zroubalik:
As discussed here, let's stay with `true` and add `false` once it is implemented.

https://github.com/kedacore/keda/issues/5215

tomkerkhove:
Closing because of the above, but thanks for proposing anyway!

PR #1269, opened 2023-12-01
docs: Introduce troubleshooting guide for Azure Service Bus

tomkerkhove:
Based on https://github.com/kedacore/keda/issues/5227, I had a quick idea to start introducing troubleshooting guides to reduce the load on maintainers/community and document recurring issues.

Currently this is added to the same page as the scaler, but we should integrate scalers in to our existing ["troubleshooting" section](https://keda.sh/docs/2.12/troubleshooting/) over time.

### Checklist

Relates to https://github.com/kedacore/keda/issues/5227


tomkerkhove:
What do you think about this idea @zroubalik / @JorTurFer

content/docs/2.12/scalers/azure-service-bus.md
```diff
@@ -107,3 +107,14 @@ spec:
...
+
+### Troubleshooting
+
+#### KEDA is unable to get queue information due to "invalid queue runtime properties: no CountDetails element"
+
+When KEDA logs show errors similar to `invalid queue runtime properties: no CountDetails element` it usually is caused because of throttling by Azure Service Bus.
+
+Consider applying one of the following mitigations:
+- Scaling the Azure Service Bus namespace to a higher SKU, or use premium
+- Reduce the polling interval of the ScaledObject/ScaledJob
```
JorTurFer, line 119: This actually means increasing the pollingInterval to reduce the frecuency (ro reduce the amount of calls)

JorTurFer:
Maybe we can create a new secction for these stuff, and include FAQ and troubleshooting pages. Troubleshooting info looks quite fragmentated between FAQ, troubleshooring, and now in page

JorTurFer:
Just to clarify, I'd move to a versioned section, because fixes related with v2.X can apply to v2.Y or not

tomkerkhove:
I agree and that is also why I added this:
> Currently this is added to the same page as the scaler, but we should integrate scalers in to our existing ["troubleshooting" section](https://keda.sh/docs/2.12/troubleshooting/) over time.

However, I am not sure if we should do it now already or in the future. Currently I think we should just add this to scalers for now and once we have more scenarios we can look at how we can support it and what the needs are (instead of assuming)

JorTurFer:
https://github.com/kedacore/keda-docs/issues/1270

zroubalik (commented):
Is this relevant only to 2.12?

JorTurFer:
> Is this relevant only to 2.12?

I think that it's interesting in general

tomkerkhove:
@zroubalik @JorTurFer as mentioned above, wanted to know if you think it's the right place to start before applying it everywhere.

I think that is not fully answered yet, once it is I will expand it to all versions

tomkerkhove:
Shall we merge this or do we look for alternative ways @zroubalik @JorTurFer ?

JorTurFer:
We can merge it IMO (but add it to v2.13 too)

tomkerkhove:
Ugh I love GitHub codespaces messing up DCO

content/docs/2.13/scalers/azure-service-bus.md
```diff
@@ -107,3 +107,15 @@ spec:
...
+
+#### KEDA is unable to get queue information due to "invalid queue runtime properties: no CountDetails element"
+
+When KEDA logs show errors similar to `invalid queue runtime properties: no CountDetails element` it usually is caused because of throttling by Azure Service Bus.
+
+Consider applying one of the following mitigations:
+- Scaling the Azure Service Bus namespace to a higher SKU, or use premium
+- Increase the polling interval of the ScaledObject/ScaledJob
+- Use [caching of metrics](./../concepts/scaling-deployments/#caching-metrics)
+
```
tomkerkhove, line 121:
```suggestion
```

zroubalik (approved):

PR #1272, opened 2023-12-06
Added note to the deploy.md (on all versions) to make people aware th…

HemeraOne:
…at only a single metric server can handle external.metrics.k8s.io per cluster

### Checklist

Fixes #


tomkerkhove:
Hi, this is already documented in the FAQ

JorTurFer:
FAQ is not totally visible. I think that we can merge the note as users usually don't check the FAQ until something is not working, but they will see the note before installing

tomkerkhove:
If we merge it, then we have 2 places to maintain IMO. We should fix FAQ discoverability instead or change the note to just check FAQ to learn more about metric server limitations

JorTurFer:
The problems with FAQ is that it doesn't allow linking it. I mean, this note could be just something like:

*NOTE: Check api server limitations here: FAQ_SECTION_LINK*

But we can't link those items, that's why I'd merge this, at least until we improve the faq

tomkerkhove:
Well, we can just fix that @JorTurFer  🤷‍♂️

Every item has an achor assigned and can be linked to: `https://keda.sh/docs/faq/#can-i-run-multiple-metric-servers-serving-external-metrics-in-the-same-cluster`. Please update the docs to just point to this instead please.

I have opened https://github.com/kedacore/keda-docs/pull/1273 to do the same for the categories and https://github.com/kedacore/keda-docs/issues/1274 to provide a button to copy.

JorTurFer:
is there any option to open the[ linked anchor](https://keda.sh/docs/faq/#can-i-run-multiple-metric-servers-serving-external-metrics-in-the-same-cluster)?
Default behaviour is browsing to it closed and that's a bit confusing: 
image

PR #1282, opened 2023-12-21
fix: Provide valid syntax for PostgreSQL queries

M4nihere:
There was a missing ";" in the triggers.metadata.query 

### Checklist

Fixes #
Just had to add a semicolon ";" at the end of the query.
image


JorTurFer:
Hello @M4nihere ,
Thanks for your contribution! 🙇 

I have a question, is the semicolon required or just for styling? I'm checking current e2e test and it doesn't have it
image

M4nihere:
Yes, there is always semicolon required at the end of a postgresql query, if you don't use semicolon the query will not return any value.

M4nihere:
As you can see here below screenshot, query without semicolon is not giving any output.
image


Screenshot with semicolon!!

image

JorTurFer:
lol, I don't know how KEDA does the magic but it works without it (but I agree with adding it if it's the standard).
Could you update other versions as well with the semicolon?

M4nihere:
I myself have tested it on Keda ScaledObject with airflow.
So to thing was that the query was not returning any value so that it can scale the Statefulsets.
While troubleshooting I found that the semicolon is required.

Thanks !!

JorTurFer:
Yeah, don't worry. I've tested with latest KEDA version and maybe something has changed, but in any case it works too, so we can add the semicolon to docs, np
Could you update other versions as well with the semicolon?

M4nihere:
Yes, I can do that but as you have mentioned in some Keda version it worked without semicolon, however I'm not sure how did that worked but I won't be able to test every Keda Version.

Because I'm using This version that's why I've found that bug.

JorTurFer:
Are you using KEDA v2.12, right?
I ask for executing the e2e test using your same version to double-check it

zroubalik (commented):
@M4nihere thanks! Could you please fix this also for the other versions (including 2.13)?

M4nihere:
@JorTurFer Just updated the 2.13 !!

M4nihere:
@JorTurFer Are you going to merge this pull request ?

JorTurFer:
> @M4nihere thanks! Could you please fix this also for the other versions (including 2.13)?

Could you update older versions please? Then we can merge this :)

M4nihere:
Sure @JorTurFer

tomkerkhove (changes requested):
Thank you! Mind updating all versions please?

M4nihere:
Hi @JorTurFer 
I have updated all the previous version with correct postgresql command.
You can merge it now.

zroubalik (approved):
Thanks 👍

ashb:
This is just stylistic btw. A semi colon is required when using `psql`, but not needed when using a DB API client library to issue queries.

And you could make the argument that it's possible harmful in KEDA as it might mistakenly lead users to think they could issue multiple queries when only one will work.

PR #1287, opened 2024-01-10
Fix Elasticsearch example

kevinnoel-be:

### Checklist

Fixes #


JorTurFer:
nice catch!
Could you update other versions as well? I think that this happens on all of them

kevinnoel-be:
> nice catch! Could you update other versions as well? I think that this happens on all of them

Sure, did it for 2.5+

JorTurFer (approved):
Thanks! ❤️

PR #1295, opened 2024-01-16
update docs on metric value needing to be greater than activation value

DP19:
Related to https://github.com/kedacore/keda/issues/5383. Updating the activation docs to make it more clear on the value needing to be greater than this value, not equal to

### Checklist

Fixes https://github.com/kedacore/keda/issues/5383


JorTurFer (commented):
Thanks for the clarification! Could you update versions above 2.8 as well? We introduced the activation concept on v2.8 so adding this clarification on all the affected version would be nice

DP19:
> Thanks for the clarification! Could you update versions above 2.8 as well? We introduced the activation concept on v2.8 so adding this clarification on all the affected version would be nice

@JorTurFer 👍🏻  - Also updated 2.14 since 2.13 was just released

PR #1303, opened 2024-02-02
Feat: Copy to Clipboard for FAQ

### LINKED ISSUE #1274: Provide "Copy Link" button for FAQ questions
FAQ questions have anchors to which can be linked, but there is no simple way to copy the URL to it.

We should provide "Copy Link" button to make that simpler, similar to the categories:
image

thisisobate:
**Todos**
- [x] POC
- [x] Fix style
- [x] Clean up code

Fixes #1274

### DIFF
--- a/assets/sass/style.sass
+++ b/assets/sass/style.sass
@@ -266,13 +266,20 @@ hr
 
 .copyBtn 
   padding: 10px
-  background: #5784f5
+  background: #326DE6
+  height: 50%
+  margin-right: 10px
+  margin-top: 5px
   color: #ffffff
   font-size: 18px
   border: none
   outline: none
   border-radius: 10px
   cursor: pointer
+  &:focus
+      outline: 2px solid #5784f5
+  &:hover
+      background: #5784f5
   &:active
       background: #809ce2
   &:before
--- a/assets/sass/style.sass
+++ b/assets/sass/style.sass
@@ -264,6 +264,41 @@ hr
 .section 
 	padding: 4rem 1.5rem
 
+.copyBtn 
+  padding: 10px
+  background: #5784f5
+  color: #ffffff
+  font-size: 18px
+  border: none
+  outline: none
+  border-radius: 10px
+  cursor: pointer
+  &:active
+      background: #809ce2
+  &:before
+      content: "Copied"
+      position: absolute
+      top: -45px
+      right: 0px
+      background: #5c81dc
+      padding: 8px 10px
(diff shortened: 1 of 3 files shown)


tomkerkhove (changes requested):
Thanks a ton!

Just a few suggestions:


- Can we make the button less explicit? It's pretty invasive but it might be me, @zroubalik/@JorTurFer ?

image

- When copying a link and going to the URL, the item does not expand so I still have to look for it, can we auto-expand them?

- Can we do the same for troubleshooting?
image

JorTurFer:
I agree with all your points 😄

zroubalik (approved):
Agree with @tomkerkhove

tomkerkhove:
> @tomkerkhove I suggest we do the troubleshooting part in a separate PR.

Sounds good!

tomkerkhove:
The expanding is great! However, if I wouldn't now there was a button I would not see it since it requires a hover:
image

Can we maybe just get rid of the full blue background and always show a subtle icon?

thisisobate:
@tomkerkhove How about this?

Screenshot 2024-02-07 at 15 01 57

zroubalik (commented):
This icon looks better, though it might be just me but the icon gives me impression that I am about to copy some piece of text or code, not a link. Do we have an icon like this?
Screenshot 2024-02-07 at 15 10 26

thisisobate:
@zroubalik Yes, we do.
I agree with you on the icons.
Will substitute the **clone** icon with the **link** icon.

tomkerkhove:
Works like a charm, thanks! However, it no longer auto-expands which it used to do though :(

thisisobate:
@tomkerkhove Actually, it does expand. It looks like you tried loading the FAQ link in the same tab. 
To test appropriately, copy the link and load it in a separate tab. That should work.

tomkerkhove:
Hm, why would that not work? It's not a blocker but curious though

tomkerkhove (approved):

thisisobate:
> Hm, why would that not work? It's not a blocker but curious though

So here's a bit of an explanation of why it won't work:, 
By default, pages are cached when loaded in the browser tab. Now, when you try to navigate to an internal link (FAQ link) in this case, it won't cause a reload or rerender because it's an internal link; thus the reason why you can't see a change when you load the link in the same tab. 

Hope that helps.

tomkerkhove:
Thanks!

PR #1330, opened 2024-03-07
Make the example for Cron Scaler match the screenshot

ecerulm:
The example now match the screenshot 

### Checklist

Fixes #

### DIFF
--- a/content/docs/1.5/scalers/cron.md
+++ b/content/docs/1.5/scalers/cron.md
@@ -16,8 +16,8 @@ triggers:
   metadata:
     # Required
     timezone: Asia/Kolkata  # The acceptable values would be a value from the IANA Time Zone Database.
-    start: 30 * * * *       # Every hour on the 30th minute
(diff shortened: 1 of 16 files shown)


zroubalik (commented):
I like it, great, could you please also fix other versions? Thanks

ecerulm:
> I like it, great, could you please also fix **other versions**? Thanks

I'm not sure I understand what you meant,  I update the other example the one about having same `start` and `end`, but I have the feel that that is probably not what you meant by "**other versions**".

zroubalik:
You fixed this in version 2.13, would be great if we can fix this also in other version (2.14, 2.12,...)

ecerulm:
> You fixed this in version 2.13, would be great if we can fix this also in other version (2.14, 2.12,...)

Got you now,  I added the changes in all other versions now.

zroubalik (approved):
Thanks!

PR #1341, opened 2024-03-22
Update apache-kafka.md

sameerjoshinice:
The value of boolean should be in single quotes , otherwise its not accepting.

### Checklist

Fixes #


JorTurFer (commented):
nice catch! Could you update other versions as well?

sameerjoshinice:
> nice catch! Could you update other versions as well?

Updated other versions where the flag was not in quotes.

JorTurFer (approved):
LGTM! Thanks a lot!

PR #1360, opened 2024-04-12
Add Dynatrace scaler docs

cyrilico:
Add documentation on how to use new Dynatrace scaler.

### Checklist

Relates to https://github.com/kedacore/keda/pull/5685


cyrilico:
Not sure why adding only to `2.14` fails the build (it seems to try to point to `2.13`?), but I kept it there since it seems there won't any more `2.13.x` releases

content/docs/2.14/scalers/dynatrace.md
```diff
@@ -0,0 +1,95 @@
...
+
+### Trigger Specification
+
+This specification describes the `dynatrace` trigger that scales based on Dynatrace metric data points.
+
+```yaml
+triggers:
+  - type: dynatrace
+    metadata:
+      host: https://dummy-instance.live.dynatrace.com/
```
tomkerkhove, line 17: Is this host always using `live.dynatrace.com` or not? Otherwise, we should just use this:
```suggestion
      organization: dummy-instance
```

content/docs/2.14/scalers/dynatrace.md
```diff
@@ -0,0 +1,95 @@
...
+      token: "NRAK-xxxxxxxxxxxxxxxxxxxxxxxxxxx"
```
tomkerkhove, line 18: This should be coming from trigger authentication and no longer through inlined secrets, we no longer introduce new ones sorry!

content/docs/2.14/scalers/dynatrace.md
```diff
@@ -0,0 +1,95 @@
...
+- `activationThreshold` - Target value for activating the scaler. Learn more about activation [here](./../concepts/scaling-deployments.md#activating-and-scaling-thresholds). (Optional, default `0`, can be a float)
+
+### Authentication Parameters
+
+You can use `TriggerAuthentication` CRD to configure authentication the `host` and `token` parameters.
+
+**Authentication:**
+
+- `host` - The Dynatrace instance to query against
+
```
tomkerkhove, line 48: Nit
```suggestion
```

content/docs/2.14/scalers/dynatrace.md
cyrilico: Sounds like a good practice to me, should it not be allowed at all or just not encouraged?

content/docs/2.14/scalers/dynatrace.md
cyrilico: It depends on the [deployment type](https://docs.dynatrace.com/docs/get-started/monitoring-environment#melocation), so not necessarily

content/docs/2.14/scalers/dynatrace.md
cyrilico: Updated on the code side to not fetch from trigger metadata

content/docs/2.14/scalers/dynatrace.md
tomkerkhove: In that case, why not introduce `deploymentType` or so which is cloud by default and uses `organization`. If it is not, then host is rquired to be filled in?

It feels like a nicer UX if `organization` is used by default.

cyrilico: The `organization` terminology isn't really used in Dynatrace afaik, so I'm not so sure about it. Plus it would require to configure 2 parameters instead of 1, if I understood correctly. I tried to follow a similar logic to other scalers, which usually have a `host` parameter (admittedly the naming is anything but consistent across scalers).

tomkerkhove: Fair enough, but it would have made the usability easier though

cyrilico:
Moved @JorTurFer 👍 I'm not sure how to tackle the CI errors, they seem to be trying to look for a previous page version, which does not make sense in this context:

```
3:51:59 PM: docs/2.15/scalers/dynatrace/index.html
3:51:59 PM:   target does not exist --- docs/2.15/scalers/dynatrace/index.html --> /docs/2.14/scalers/dynatrace/
3:52:00 PM: Failed during stage 'building site': Build script returned non-zero exit code: 2 (https://ntl.fyi/exit-code-2)
3:52:00 PM: ========================================================================
3:52:00 PM: ✘✘✘ failed in 1.338598592s
3:52:00 PM: 1 errors in 1371 documents
3:52:00 PM: make: *** [Makefile:17: check-links] Error 1
```

JorTurFer:
Yeah, I forgot to explain it, sorry
You need to include the link as excluded verification links here -> https://github.com/kedacore/keda-docs/blob/main/.htmltest.yml
Splunk scaler is an example about how to add it

content/docs/2.15/scalers/dynatrace.md
```diff
@@ -0,0 +1,93 @@
++++
+title = "Dynatrace"
+availability = "2.15+"
+maintainer = "Community"
+description = "Scale applications based on Dynatrace metric data points"
```
JorTurFer, line 5:
```suggestion
availability = "2.15+"
maintainer = "Community"
category = "Metrics"
description = "Scale applications based on Dynatrace metric data points"
```

content/docs/2.15/scalers/dynatrace.md
```diff
@@ -0,0 +1,93 @@
...
+      from: now-2d
+      threshold: "10"
+      # Optional
+      activationThreshold: "5"
+```
+
+**Parameter list:**
+
+- `host` - The Dynatrace instance to query against
+- `token` - The API key that will be leveraged to connect to Dynatrace and make requests ([official documentation](https://docs.dynatrace.com/docs/dynatrace-api/basics/dynatrace-api-authentication)). Requires the `metrics.read` scope. Must be provided via Trigger Authentication (see [Authentication Parameters](#authentication-parameters))
```
JorTurFer, line 29: The token can't be provided via metadata (the code is correct)
```suggestion
- `token` - The API key that will be leveraged to connect to Dynatrace and make requests ([official documentation](https://docs.dynatrace.com/docs/dynatrace-api/basics/dynatrace-api-authentication)). Requires the `metrics.read` scope. Must be provided via Trigger Authentication (see [Authentication Parameters](#authentication-parameters))
```

JorTurFer (commented):
Looking nice! just 2 quick suggestions inline

content/docs/2.15/scalers/dynatrace.md
cyrilico: correct, and that is what it says 👀

JorTurFer: Yeah, but we define the parameters only where they can be used (that's why host in twice for example). As `token` can be set only via TriggerAuthentication, it should be defined only in the section `### Authentication Parameters`

JorTurFer: And beingtotally honest, I didn't read the full message, mb 100%

cyrilico: gotcha, removed 👍

JorTurFer (approved):
❤️

zroubalik (approved):

PR #1395, opened 2024-05-20
Prometheus: Mark queryParameters as an optional field instead of an required field

### LINKED ISSUE #1394: Prometheus: `queryParameters` is an optional parameter
### Discussed in https://github.com/kedacore/keda-docs/discussions/1391

<div type='discussions-op-text'>

<sup>Originally posted by **ahinh43** May 15, 2024</sup>
I'm working on upgrading our Keda deployment from 2.11 to 2.14, and one of the changes introduced in 2.13 is the [queryParameters field](https://github.com/kedacore/keda/issues/4962) for the prometheus scaler. However, it's documented as a required field yet in the [source code the field is marked optional](https://github.com/kedacore/keda/blob/main/pkg/scalers/prometheus_scaler.go#L58). 

My initial impression of reading the [trigger specifications](https://keda.sh/docs/2.14/scalers/prometheus/#trigger-specification)
is that missing fields under required fields would result in the request being rejected during an apply but I could be misunderstanding this.

Is there any implications for not setting queryParameters in a prometheus trigger's metadata field? </div>

ahinh43:
Marks the `queryParameters` field as an optional field instead of a required field, as it's actually not mandatory for this field to be configured when deploying new Prometheus triggers.

### Checklist

Fixes #1394

### DIFF
--- a/content/docs/2.13/scalers/prometheus.md
+++ b/content/docs/2.13/scalers/prometheus.md
@@ -17,7 +17,6 @@ triggers:
     # Required fields:
     serverAddress: http://<prometheus-host>:9090
     query: sum(rate(http_requests_total{deployment="my-deployment"}[2m])) # Note: query must return a vector/scalar single element response
-    queryParameters: key-1=value-1,key-2=value-2
     threshold: '100.50'
     activationThreshold: '5.5'
     # Optional fields:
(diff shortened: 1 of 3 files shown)


JorTurFer (commented):
nice catch ! 
Could you update other versions as well?

ahinh43:
> nice catch ! Could you update other versions as well?

Thanks for catching that! The other versions should be updated with this change.

zroubalik (approved):

PR #1393, opened 2024-05-17
update postgres docs for nulls and coalesce

robpickerill:
Updates Postgres scaler docs to add a note about handling nulls in the query.

### Checklist

- [ ] Commits are signed with Developer Certificate of Origin (DCO)

Fixes https://github.com/kedacore/keda/issues/5686 for postgres, will present separate PRs for other RDBMS scalers

PR #1396, opened 2024-05-21
add note for null and coalesce for mysql

robpickerill:
Adds a note about the use of coalesce for null values for MySQL. Follows on from: https://github.com/kedacore/keda-docs/pull/1393


zroubalik (approved):
Thanks!

PR #1397, opened 2024-05-21
add note for null and coalesce for mssql

robpickerill:
Adds a note about the use of coalesce for null values for MSSQL. Follows on from: https://github.com/kedacore/keda-docs/pull/1393


zroubalik (approved):
Thanks!

PR #1400, opened 2024-05-28
Feat: Provide filtering capabilities for scalers

### LINKED ISSUE #412: Improve scaler catalog to be more user-friendly

thisisobate:
Todo
- [x] provide filter button on mobile
- [x] Annotate existing scalers with metadata for easier filtering

Relates to #412


thisisobate:
cc: @tomkerkhove

tomkerkhove:
I was waiting with reviewing because there was a TODO :) Doing that now!

thisisobate:
> This takes up a big chunck of the scaler area

This is because bulma spacing selectors automatically allocate that much space for columns. I can try using vanilla css instead to reduce the spacing so the scaler cards can fit comfortably.

thisisobate:
@tomkerkhove any updates on [this](https://github.com/kedacore/keda-docs/issues/1401)?

tomkerkhove:
Btw, can we align the external vs built-in filter in to this new model please? Now we have 2 flavors

thisisobate:
> No not for now, but I don't think this should block this PR, does it?

I agree. It doesn't block it. In that case, I think we should merge and close this issue.

tomkerkhove:
> Btw, can we align the external vs built-in filter in to this new model please? Now we have 2 flavors

We need to do this first though

thisisobate:
> Btw, can we align the external vs built-in filter in to this new model please? Now we have 2 flavors

What exactly do you mean? The "maintainer" filter option already supports the built-in and external scalers. E.g, to filter the maintainer "JensWalter", you need to switch to the `external` tab.

If you are saying we should add a new "built-in scaler" and "external scaler" filter options in the new filter pane, then I don't think that would work. Moreover, I don't think it's necessary given we already designed the ux and data architecture to work for the existing UI structure.

tomkerkhove:
> If you are saying we should add a new "built-in scaler" and "external scaler" filter options in the new filter pane, then I don't think that would work. Moreover, I don't think it's necessary given we already designed the ux and data architecture to work for the existing UI structure.

I tend to disagree though, we currently have 2 UX for filtering scalers which is confusing. Certainly if we'll start adding more filters

thisisobate:
> we currently have 2 UX for filtering scalers

You have a point. If we must go the route you suggest, then it would warrant a complete overhaul of the data architecture which is an additional workload and frankly, it's slightly out of scope. 

I will have to discuss with @nate-double-u internally and see what we can do about this. In the meantime, I would advise you open a new issue for this request.

Thank you!

tomkerkhove:
Let's keep this open, it's better to keep the current experience rather than introducing two different UXes.

Why would this alignment cause a complete data architecture overhaul?

thisisobate:
> Why would this alignment cause a complete data architecture overhaul?


Because for this to work, we would have to generate one central
datastore/index to query both in-built and external scalers which might be
tricky to solve.

We can keep this open. I will try and see if I can get something working;
will keep you posted.

cc: @nate-double-u

thisisobate:
@tomkerkhove PTAL

tomkerkhove:
A lot better, thank you!

image

One last ask: Can we change the order so that we show, "Type", "Category", "Maintainer" please?
Any other remarks @kedacore/keda-maintainers?

thisisobate:
@tomkerkhove I think we're good to go 😀

tomkerkhove (approved):

tomkerkhove:
Thank you!

PR #1403, opened 2024-06-06
clearify confusing example

AlonGluz:
- **align examples after looking at the code**
- **align examples after looking at the code**

### Checklist

Fixes #

### DIFF
--- a/content/docs/2.14/concepts/scaling-deployments.md
+++ b/content/docs/2.14/concepts/scaling-deployments.md
@@ -40,7 +40,7 @@ metadata:
   name: {scaled-object-name}
   annotations:
     scaledobject.keda.sh/transfer-hpa-ownership: "true"     # Optional. Use to transfer an existing HPA ownership to this ScaledObject
-    validations.keda.sh/hpa-ownership: "false"               # Optional. Use to disable HPA ownership validation on this ScaledObject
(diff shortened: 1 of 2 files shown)


tomkerkhove:
Hey, may I ask why you are changing this? I believe true is the default?

AlonGluz:
@tomkerkhove Simply because it is confusing. The values and the comments don't match the example, `Optional. Use to disable HPA ownership validation on this ScaledObject`.
Also when looking at the code it skips the validation is the value is false
https://github.com/kedacore/keda/blob/fd4fcc3dcb6eb86e745c310f71c93526ed99ba61/apis/keda/v1alpha1/scaledobject_webhook.go#L224

AlonGluz:
@tomkerkhove WDYT?

tomkerkhove:
> @tomkerkhove Simply because it is confusing. The values and the comments don't match the example, `Optional. Use to disable HPA ownership validation on this ScaledObject`. Also when looking at the code it skips the validation is the value is false [kedacore/keda@`fd4fcc3`/apis/keda/v1alpha1/scaledobject_webhook.go#L224](https://github.com/kedacore/keda/blob/fd4fcc3dcb6eb86e745c310f71c93526ed99ba61/apis/keda/v1alpha1/scaledobject_webhook.go#L224)

This is the criteria for disabling which is different from the default value though, which is true

PR #1413, opened 2024-06-23
feat: splunk scaler docs

circa10a:
Added `splunk` scaler documentation for potentially v2.15 release. Discussion issue: https://github.com/kedacore/keda/issues/5904

### Checklist

Relates to: https://github.com/kedacore/keda/pull/5905

### DIFF
--- a/content/docs/2.15/scalers/splunk.md
+++ b/content/docs/2.15/scalers/splunk.md
@@ -1,0 +1,292 @@
++++
+title = "Splunk"
+availability = "v2.15+"
+maintainer = "Community"
+category = "Data & Storage"
+description = "Scale applications based on Splunk saved search results."
+go_file = "splunk_scaler"
++++
+
+### Trigger Specification
+
+This specification describes the `splunk` trigger that scales based on the result of a [saved search](https://docs.splunk.com/Documentation/Splunk/9.2.1/Search/Savingsearches).
+
+The trigger always requires the following information:
+
+```yaml
+triggers:
+  - type: splunk
+    metadata:
+      host: https://splunk.default.svc.cluster.local:8089
+      targetValue: "1"
+      activationValue: "10"
+      savedSearchName: my-saved-search-name
+      valueField: count
+```
+
+**Parameter list:**
+
+- `host` - Search API host and port. Example: `https://localhost:8089`.
+- `unsafeSsl` - Whether to trust invalid certificates or not. (Values: `"true"`, `"false"`, Default: `"false"`, Optional)
+- `targetValue` - Value to reach to start scaling (This value can be a integer or float).
+- `activationValue` - Target value for activating the scaler. Learn more about activation [here](./../concepts/scaling-deployments.md#activating-and-scaling-thresholds).
+- `savedSearchName` - Name of saved search that returns metric data for scaling.
+- `valueField` - The name of the field in the search results containing the metric value. Example: `index=_internal | tail | stats count`, the `valueField` is `count`.
+
+### Authentication Parameters
+
+You can authenticate by using a username/password or an API token. You will need to use `TriggerAuthentication` CRD to configure the authentication.
+
+> **Note:**
+>
+> `TriggerAuthentication` is required to use this scaler due to the hard requirement of providing a `username` for the Splunk API.
+
+**Parameter list:**
+
+- `username` - Splunk username authorized to access the search API.
+- `apiToken` - Splunk API token for supplied `username`. Conflicts with `password`.
+- `password` - Password for supplied `username`. Conflicts with `apiToken`.
+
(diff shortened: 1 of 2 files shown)


circa10a:
> 9:45:18 AM: docs/2.15/scalers/splunk/index.html
9:45:18 AM:   target does not exist --- docs/2.15/scalers/splunk/index.html --> /docs/2.14/scalers/splunk/
9:45:19 AM: Failed during stage 'building site': Build script returned non-zero exit code: 2 

Not sure if this is expected as the scaler wouldn't exist in v2.14 of KEDA

content/docs/2.15/scalers/splunk.md
```diff
@@ -0,0 +1,279 @@
...
+      username: admin
+      targetValue: "1"
+      activationValue: "10"
+      savedSearchName: my-saved-search-name
+      valueField: count
+```
+
+**Parameter list:**
+
+- `username` - Splunk username authorized to access the search API.
```
zroubalik, line 30: I think we should probably place the `username` next to the `password` in the trigger auth. I suppose it will only be used together and the username without a password is not gonna work, right?

CC @kedacore/keda-maintainers

circa10a: For auth, a password or API token can be supplied and a username is required for either. I kept username in trigger metadata since it's not technically a secret and is required for both methods of authentication.

FWIW, the username is required in the API route when making requests, this is another reason why its a required field in trigger metadata.

zroubalik: okay, makese sense. let's keep it this way.

circa10a: I can change the scaler to always source username from trigger auth if that's preferred.

zroubalik: Let's see what other think, thanks 🙏

tomkerkhove: I do think Zbynek has a point though, do we at least support passing username as part of trigger authentication? I do not see it so would prefer to add it  there.

Otherwise, end-users need to change 2 places if they want to change identities nor cannot re-use it across scalers and have to duplicate it everywhere (although the secret can be shared which defeats purpose)

circa10a: Fair point. I updated https://github.com/kedacore/keda/pull/5905 and the docs here to source `username` from auth params

zroubalik: @circa10a and to be clear, the scaler can't be used withouth `username` ie never without TriggerAuthentication, is that right? (it is not a problem, just a clarification we might want to add to the docs).

circa10a: Correct. The operator will also error stating that username is supplied as well. Where is the most applicable place in the docs to specify this requirement?

content/docs/2.15/scalers/splunk.md
```diff
@@ -0,0 +1,288 @@
...
+- `unsafeSsl` - Whether to trust invalid certificates or not. (Values: `"true"`, `"false"`, Default: `"false"`, Optional)
+- `targetValue` - Value to reach to start scaling (This value can be a integer or float).
+- `activationValue` - Target value for activating the scaler. Learn more about activation [here](./../concepts/scaling-deployments.md#activating-and-scaling-thresholds).
+- `savedSearchName` - Name of saved search that returns metric data for scaling.
+- `valueField` - The name of the field in the search results containing the metric value. Example: `index=_internal | tail | stats count`, the `valueField` is `count`.
+
+### Authentication Parameters
+
+You can authenticate by using a username/password or an API token. You will need to use `TriggerAuthentication` CRD to configure the authentication.
+
```
davogen, line 39:
```suggestion

> **Note:**
>
> I would put the note here

```

content/docs/2.15/scalers/splunk.md
davogen: I am followin the conversation and I would suggest here: https://github.com/kedacore/keda-docs/pull/1413#discussion_r1661395761

WDYT @zroubalik ?

content/docs/2.15/scalers/splunk.md
zroubalik: I like this

content/docs/2.15/scalers/splunk.md
zroubalik: Agree, something around the lines, that user should be aware that the scaler always requires `username` set

content/docs/2.15/scalers/splunk.md
circa10a: Thank you :)

Added the blurb

zroubalik (approved):
LGTM, thanks!

PR #1424, opened 2024-07-09
Update github-runner.md DISABLE_RUNNER_UPDATE -> DISABLE_AUTO_UPDATE

YoranSys:
The correct value from : https://github.com/myoung34/docker-github-actions-runner is : DISABLE_AUTO_UPDATE


JorTurFer (commented):
Thanks for the improvement! Could you update other versions as well?

PR #1434, opened 2024-07-24
remove metrics server leader election env vars

lantingchiang:
The documentation originally mentions the possibility to set leader election parameters through environment variables for the keda-metrics-server. However, it was confirmed in https://github.com/kedacore/keda/issues/5959 that the keda-metrics-server does not use leader election, and these parameters should not be there. 

### Checklist

Fixes #

### DIFF
--- a/content/docs/2.15/operate/cluster.md
+++ b/content/docs/2.15/operate/cluster.md
@@ -158,7 +158,7 @@ To modify this properties you can set environment variables on both KEDA Operato
 
 ## Configure Leader Election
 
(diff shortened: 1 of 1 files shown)


JorTurFer (commented):
Checking the code, those envs are still in use and they should be removed from the code before. This means that we should remove them only from v2.15 version and remove them from the code too -> https://github.com/kedacore/keda/blob/main/cmd/adapter/main.go#L89-L105

Could you update this PR and create another PR removing the code? I've reviewed the adapter and nothing is using the leader election API, so it's safe removing it

Issue #1368, opened 2024-04-15
Write a Procedure for Setting Up a Scaler


dwelsch-esi:
Write a procedure for setting up a scaler for KEDA. Include all steps that are common to any scaler. Supplemental instructions specific to particular scalers can be elsewhere.

This procedure is fundamental to using KEDA and should go at or near the beginning of the Operator's Guide.

### Use-Case

Setting up a scaler seems to be largely a matter of installing the scaler and providing parameters in a configuration file. While the configurations are provided for all the various scalers, there doesn't seem to be a description of the procedure for doing the basic setup. This should go at the top of the Operator guide. 

Users should be able to follow the procedure for any (or at least most) scalers. Any scaler that requires special instructions other than the configuration file should have its own procedure page, listing the extra steps required.

### Specification

Write complete, step-by-step instructions. Don't assume the user knows anything. For example, name the configuration file and tell users where to put it, even if this is common knowledge. This is the generic template they'll use to set up any scaler, including their first.

shubhusion:
Hi @dwelsch-esi and @tomkerkhove,

Could you please assign this issue to me? I would like to contribute to it. I have applied for the TUF documentation project through the Linux Foundation.

Thank you!

nate-double-u:
Hi @shubhusion, I appreciate your enthusiasm, but the selections haven't been made yet for LFX Mentorship 2024 Term 2. Assigning someone before the announcements go out would be premature.

zroubalik:
@shubhusion we would be happy to assign this issue to you, if you are still interested.

shubhusion:
> @shubhusion we would be happy to assign this issue to you, if you are still interested.

Yeah I am interested. Can you also provide me resources to get started.

shubhusion:
@zroubalik Procedure needs to be updates in version 2.15 or 2.14 ?

zroubalik:
@shubhusion I think that the content will be the same for both versions, so it's a low effort to update both versions, thanks :)

shubhusion:
> @shubhusion I think that the content will be the same for both versions, so it's a low effort to update both versions, thanks :)

Hi @zroubalik, Could you please guide me a little? I am new to this repository 
I have updated version 2.14 rn which other version do I need to update?

shubhusion:
@zroubalik @tomkerkhove can you please assign this issue to me ? I have already made a PR

nate-double-u:
[...]

PR #1415, opened 2024-06-28
Created Procedure for Setting Up a Scaler 

shubhusion:
### Description of Changes

Created a procedure for setting up a scaler in KEDA.

### Checklist
- [x] Commits are signed with Developer Certificate of Origin (DCO)

### Issue Fixed
Fixes [kedacore/keda-docs#1368](https://github.com/kedacore/keda-docs/issues/1368)

PR #1446, opened 2024-08-05
Created Procedure for Setting Up Autoscaling with KEDA

shubhusion:
### Description of Changes

Created a procedure for setting up a scaler in KEDA.

### Checklist

### Issue Fixed
Fixes [kedacore/keda-docs#1368](https://github.com/kedacore/keda-docs/issues/1368)


shubhusion:
Hi @JorTurFer,

Could you please review the finalized PR for #1368? The changes you specified in the previous PR #1415 have been implemented here. However, due to an issue with rebasing, I had to close the earlier PR.

Thank you!

JorTurFer (commented):
Looks good to me! The only point is that we should include it also as part of v2.16 (if you rebase the branch the folder will be there).

There is another question that comes to my mind, maybe we could transform this into a getting started that we place in the getting started section instead of a versioned documentation. WDYT @tomkerkhove @zroubalik ? 
Maybe we can do it in a following PR just refactoring a bit this one (which is nice, tbh)

shubhusion:
> Looks good to me! The only point is that we should include it also as part of v2.16 (if you rebase the branch the folder will be there).
> 
> There is another question that comes to my mind, maybe we could transform this into a getting started that we place in the getting started section instead of a versioned documentation. WDYT @tomkerkhove @zroubalik ? Maybe we can do it in a following PR just refactoring a bit this one (which is nice, tbh)

@JorTurFer Included in v2.16 also.

zroubalik (approved):
LGTM

@JorTurFer agree on the Getting Started Guide, I think this is something we can do as a follow up.

PR #1470, opened 2024-09-18
add documentation for cpu/memory activation feature

kunwooy:
Add documentation for cpu/memory scaler activation feature.

### Checklist

Fixes #


zroubalik (commented):
I think we whould be more verbose here and explain that this is not a typical de/activation as it works in a combination with other scaler etc.

PR #1477, opened 2024-09-26
Documentation for custom KEDA scaler for Splunk Observability Cloud 

sschimper-splunk:
With this pull request, I would like to submit documentation on a custom KEDA scaler for Splunk Observability Cloud, which I have developed and which at the moment of this submission resides [here](https://github.com/sschimper-splunk/keda/tree/feature/splunk-observability-scaler). Once I create my pull request for the scaler itself, I will reference this PR for the docs.

### Checklist

### DIFF
--- a/content/docs/2.16/scalers/splunk_observability.md
+++ b/content/docs/2.16/scalers/splunk_observability.md
@@ -1,0 +1,131 @@
++++
+title = "Splunk Observability"
+availability = "v2.15+"
+maintainer = "Community"
+category = "Data & Storage"
+description = "Scale applications based on Splunk Observability Cloud metrics."
+go_file = "splunk_observability_scaler"
++++
+
+### Trigger Specification
+
+This specification describes the `splunk-observability` trigger that scales based on the result of a metric series queried from the Splunk Observability Cloud platform with a [SignalFlow query](https://dev.splunk.com/observability/docs/signalflow/).
+
+The trigger always requires the following information:
+
+```yaml
+triggers:
+  - type: splunk-observability
+    metricType: Value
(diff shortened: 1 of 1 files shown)


content/docs/2.16/scalers/splunk_observability.md
```diff
@@ -0,0 +1,131 @@
++++
+title = "Splunk Observability"
+availability = "v2.15+"
```
circa10a, line 3: This would be 2.16+ assuming the new scaler is merged.

content/docs/2.16/scalers/splunk_observability.md
```diff
@@ -0,0 +1,131 @@
...
+#### Authentication
+
+```yaml
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: splunk-secrets
+data:
+  accessToken: dGVzdF9hY2Nlc3NfdG9rZW4= # base64-encrypted example access token
```
circa10a, line 87: base64 `encoded` since base64 isn't encryption

content/docs/2.16/scalers/splunk_observability.md
circa10a: @sschimper-splunk looks like you put 2.6 instead of 2.16 by mistake

circa10a: Since the version isn't available yet, the build will fail. So you need to explicitly ignore the URL like this: https://github.com/kedacore/keda-docs/pull/1413/files#diff-9ad64044531d7d8c033064f11141586478b535147312374c24f8d2d997722889R7

PR #1483, opened 2024-10-09
Clarify pollingInterval and Caching Metrics

### LINKED ISSUE #1482: Clarify pollingInterval and Caching Metrics
I think the documentation on [pollingInterval](https://keda.sh/docs/2.15/reference/scaledobject-spec/#pollinginterval) could be improved. Described: “This is the interval to check each trigger on. By default, KEDA will check each trigger source on every ScaledObject every 30 seconds.” However, pollingInterval is only relevant from 0<->1 scaling.
The [Datadog scaler documentation](https://keda.sh/docs/2.15/scalers/datadog/#polling-intervals-and-datadog-rate-limiting) describes better how pollingInterval works. I would like to clarify the text for pollingInterval. I would also like to add a line with a reference to ‘[Caching Metrics](https://keda.sh/docs/2.15/concepts/scaling-deployments/#caching-metrics)’ so if you want to respect the pollingInterval to the external metric for 1 <-> N scaling.

In the text about [Caching Metrics](https://keda.sh/docs/2.15/concepts/scaling-deployments/#caching-metrics) I would like to add a reference to the [triggers](https://keda.sh/docs/2.15/reference/scaledobject-spec/#triggers). This because ‘caching metrics’ give information about ‘This feature enables’, but the not which one.

I think that with these updates the documentation and clarity on pollingInterval and Caching Metrics will be improved.

### Use-Case

Clarify pollingInterval and Caching Metrics

rickbrouwer:
- Clarified pollingInterval.
- Added a reference to the triggers page for Caching Metrics to clarify which feature can be used.

### Checklist

Fixes #1482


zroubalik (approved):
Thanks, I like this a lot!

PR #1490, opened 2024-10-21
rfac: Replace links for direct links at Operate page

rickbrouwer:
Pages like the [Reference page](https://keda.sh/docs/2.15/reference/) use direct links. The [Operate page](https://keda.sh/docs/2.15/operate/) however uses separate 'link' links. This PR is used to keep it consistent.
If there is a reason to present the links for this page in this way, the PR may be closed.

### Checklist


zroubalik (approved):
Thanks!

PR #1507, opened 2024-11-30
Add docs for bound service account token trigger authentication

maxcao13:
Adds docs for `BoundServiceAccountToken` trigger auth provider/source.

### Checklist

Depends on: https://github.com/kedacore/keda/pull/6272

### DIFF
--- a/content/docs/2.17/authentication-providers/bound-service-account-token.md
+++ b/content/docs/2.17/authentication-providers/bound-service-account-token.md
@@ -2,13 +2,100 @@
 title = "Bound service account token"
 +++
 
-You can pull a service account token into the trigger by defining the `serviceAccountName` of the Kubernetes ServiceAccount and token `expiry` duration.
+You can pull one or more service account tokens into the trigger by defining the `serviceAccountName` of the Kubernetes service account.
 
 ```yaml
-boundServiceAccountToken:                        # Optional.
-  - parameter: connectionString                  # Required - Defined by the scale trigger
-    serviceAccountName: my-keda-service-account  # Required.
-    expiry: 1h                                   # Required.
+boundServiceAccountToken:                   # Optional.
+  - parameter: connectionString             # Required - Defined by the scale trigger
+    serviceAccountName: my-service-account  # Required.
 ```
 
 **Assumptions:** `namespace` is in the same resource as referenced by `scaleTargetRef.name` in the ScaledObject, unless specified otherwise.
+
+## Permissions for KEDA to request service account tokens
+
+By default, the KEDA operator does not have the necessary permissions to request service account tokens from an arbitrary service account. This is to prevent a privilege escalation where a bad actor could use KEDA to request tokens on behalf of any service account in the cluster.
+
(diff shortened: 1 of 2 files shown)


content/docs/2.17/authentication-providers/bound-service-account-token.md
```diff
@@ -0,0 +1,14 @@
...
+title = "Bound service account token"
++++
+
+You can pull a service account token into the trigger by defining the `serviceAccountName` of the Kubernetes ServiceAccount and token `expiry` duration.
+
+```yaml
+boundServiceAccountToken:                        # Optional.
+  - parameter: connectionString                  # Required - Defined by the scale trigger
+    serviceAccountName: my-keda-service-account  # Required.
+    expiry: 1h                                   # Required.
```
JorTurFer, line 11: This parameter isn't supported in the code, I think that it could be interesting to support this override as optional, keeping the current approach of a global value which can be overridden via an optional field here. WDYT?

JorTurFer (commented):
As default, KEDA doesn't have permission in the RBAC to create these tokens (and not granting it as default is worth IMHO), so maybe we could add an example of the required extra roles that needs to be added to KEDA operator service account (or event better, and example of all the required RBAC manifests)

content/docs/2.17/authentication-providers/bound-service-account-token.md
maxcao13: This was written before we went through the review changes in https://github.com/kedacore/keda/pull/6272, which still had the configurable expiry per trigger. I should just remove the `expiry` here. WDYT about leaving the optional expiry feature as an enhancement in a future PR? 

> As default, KEDA doesn't have permission in the RBAC to create these tokens (and not granting it as default is worth IMHO), so maybe we could add an example of the required extra roles that needs to be added to KEDA operator service account (or event better, and example of all the required RBAC manifests)
And as a response to this, yes I believe that would be a great addition to add the required example rbacs to make this work. I will add that in the next commits, thanks!

maxcao13: Changed in https://github.com/kedacore/keda-docs/pull/1507/commits/e0133f0479578af903c601709ef21596f4cd3c81.

content/docs/2.17/authentication-providers/bound-service-account-token.md
```diff
@@ -0,0 +1,54 @@
...
+boundServiceAccountToken:                   # Optional.
+  - parameter: connectionString             # Required - Defined by the scale trigger
+    serviceAccountName: my-service-account  # Required.
+```
+
+**Assumptions:** `namespace` is in the same resource as referenced by `scaleTargetRef.name` in the ScaledObject, unless specified otherwise.
+
+## Permissions for KEDA to request service account tokens
+
+By default, the KEDA operator does not have the necessary permissions to request service account tokens from an arbitrary service account. This is to prevent a privilege escalation where a bad actor could use KEDA to request tokens on behalf of any service account in the cluster.
```
zroubalik, line 17: We should document here the way how to enable this via Helm Chart, once we have the final implemenetation there

JorTurFer (approved):

PR #1594, opened 2025-07-16
fix: replace typographic quotes in scaler descriptions

rickbrouwer:
Remove markdownify from the description field in the scaler tiles JSON (list.lunr.json).
The description is plain text from front matter and does not require markdownify.
Using markdownify caused visual issues in the UI, so this prevents single quotes (') from being converted to HTML entities like `&rsquo;` and `&lsquo;`.

Example:

image

image

### Checklist


JorTurFer (approved):

JorTurFer:
awesome!

PR #1611, opened 2025-08-14
Add clarification about scale to zero

Matt-STN:
As referenced in this issue : https://github.com/kedacore/keda/issues/6863, added an example and some clarification about the possibility of scaling deployments to 0 while having other triggers.
Also added a small step by step TL;DR for people wanting to add a scale-to-zero feature to an already created scaler.

### Checklist

### DIFF
--- a/content/docs/2.17/authentication-providers/aws.md
+++ b/content/docs/2.17/authentication-providers/aws.md
@@ -33,6 +33,43 @@ You can override the default KEDA operator IAM role by specifying an `roleArn` p
 
 If you would like to use the same IAM credentials as your workload is currently using, `podIdentity.identityOwner` can be set with the value `workload` and KEDA will inspect the workload service account to check if IRSA annotation is there and KEDA will assume that role.
 
+**Regional STS endpoints**
+
+- If you’re using **`identityOwner: keda`** (the operator’s IRSA role), annotate the **`keda-operator`** ServiceAccount in the `keda` namespace.  
+- If you’re using **`identityOwner: workload`** (the default), annotate **your workload’s** ServiceAccount (the one your Deployment uses).
+
+```yaml
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: <SERVICE_ACCOUNT_NAME>
+  namespace: <NAMESPACE>
+  annotations:
+    eks.amazonaws.com/role-arn: <YOUR_IRSA_ROLE_ARN>
+    # Required for regional STS endpoints (e.g. eu-west-1)
+    eks.amazonaws.com/sts-regional-endpoints: "true"
+ ```
+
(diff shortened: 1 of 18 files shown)


rickbrouwer:
Thanks for the contribution! 🙏🏼 

You indicate that the deployment will scale to 0 between 20:00 and 06:00, and will have between 1 and 4 replicas between 06:00 and 20:00. But what happens when the CPU spikes between 20:00 and 06:00?

content/docs/2.0/scalers/cron.md
```diff
@@ -60,7 +60,37 @@ scaling down will happen 5 minutes after the cron schedule `end` parameter.
 It's almost always an error to try to do the other way
 around, i.e. set `desiredReplicas: 0` in the cron trigger.
 
-#### Example: scale up to 10 replicas from 6AM to 8PM and scale down to 0 replicas otherwise
+> 💡 **NOTE**: When you set `desiredReplicas` in a `cron` trigger, you are defining the **minimum number of replicas** that should run between `start` and `end`. If you also configure other triggers (e.g., `cpu`), the number of replicas during that time can grow from `desiredReplicas` up to `maxReplicaCount` based on those triggers.
```
JorTurFer: I'm not 100% sure about the point of "**minimum number of replicas**". Technically it's true, but I'd describe how the HPA Controller works  instead of saying that it's the minimum replicas.

WDYT about something like:
> NOTE: As the HPA Controller will evaluate all the metrics at once and will take the one which requires more instances (`max(metrics)`), the value set for `desiredReplicas` technically acts as a "dynamic" minimum replicas. For example, if you have other trigger like CPU, during the time between `start` and `end` you will have AT LEAST `desiredReplicas` because of that `max(metrics)`

rickbrouwer:
> > But what happens when the CPU spikes between 20:00 and 06:00?
> 
> Only external metrics can be used for scaling from/to zero. When CPU/Memory scalers are used to scale from/to zero, another external metric is required. In this scenario, as soon as there isn't any external metric active, the workload is scaled to zero ignoring the CPU/memory value (and when it's scaled to zero, the CPU and memory can't have spikes)

Yeah, true. But the only concern I have with this example is that someone might see the CPU trigger as a second trigger (as a scaler), and pass the logic on to other scalers and possibly thinks he's creating a window via the cron scaler. But maybe I'm thinking too complicated ;)

content/docs/2.0/scalers/cron.md
Matt-STN: That looks good ! Thanks

JorTurFer (approved):
Nice improvement! ❤️

JorTurFer:
Sorry because I think that we missed the PR :( 
Thanks for the contribution!

PR #1612, opened 2025-08-18
fix: timeout should be of type string | unless it would not be applie…

darmond157:
fix: timeout value in prometheus trigger should be of type string. unless it would not be applied in manifests using kubectl.

prometheus doc of directory content/docs/2.17/scalers/prometheus.md has been updated.
`timeout: 1000` cannot be applied by kubectl. it should be `timeout: "1000"`


JorTurFer (approved):
Thanks a lot! 🙇

PR #1620, opened 2025-08-27
fix minor indent in temporal docs

Deferrent:
Add an indent to the `authenticationRef.name` field in the temporal docs

### Checklist

Fixes #


JorTurFer:
Thanks for the fix!

PR #1625, opened 2025-09-16
Use an image which exposes metrics in "Getting Started" tutorial

creydr:
Currently the "Getting Started" tutorial, uses the `hashicorp/http-echo` which doesn't expose any metrics (like the later used `http_requests_total`) and relies on the metrics from the LoadBalancer. Users which go through this tutorial might not have to a LoadBalancer exposing those metrics, which results in the pods not getting scaled as expected.
This PR addresses it and uses an image ([stefanprodan/podinfo](https://github.com/stefanprodan/podinfo)), which exposes metrics (like the `http_requests_total`) which can get scraped by prometheus and thus make the scaling work via the http scaler work.

### Checklist

### DIFF
--- a/content/docs/2.18/deploy.md
+++ b/content/docs/2.18/deploy.md
@@ -292,11 +292,18 @@ We'll deploy a simple application that responds to HTTP requests. For this examp
          spec:
             containers:
             - name: http-app
-              image: hashicorp/http-echo
-              args:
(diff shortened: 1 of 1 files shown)


JorTurFer (commented):
WDTY @kedacore/keda-core-contributors @kedacore/keda-core-maintainers ?

PR #1484, opened 2024-10-11
Announce removal of deprecated variable in CPU Memory Scaler

rickbrouwer:
The parameter `type` is already from release 2.7 deprecated. I think it's time to announce  and remove the deprecated code in version 2.18.

### Checklist

- [X] Commits are signed with Developer Certificate of Origin (DCO)

Related https://github.com/kedacore/keda/pull/6229

PR #1643, opened 2025-10-14
Update legacy parameter name type instead of metricType

CharlesB2:
The code was updated a while ago after #1484 but the docs change wasn't complete

### Checklist

Fixes #


JorTurFer:
Hello
Thanks for the contribution! The deprecation message is correct because it's not a renaming of the parameter but removing the parameter from the metadata section. Current `trigger.metadata.type` isn't valid and `trigger.metricType` must be used instead.

As soon as we remove the parameter, we will remove the whole documentation section for it too

PR #1661, opened 2025-12-02
fix(doc): updated kubernetes-workload.md

isagarparmar:

### Checklist

Fixes #

- Numbers are usually provided as actual numbers, not strings. 
- YAML will accept both, but numeric values should be unquoted unless the scaler explicitly requires strings (this one doesn’t).
- Removing quotes avoids confusion about whether the value is being parsed as a string or number.
- Mixed quoting rules can confuse newcomers.

### DIFF
--- a/content/docs/2.10/scalers/kubernetes-workload.md
+++ b/content/docs/2.10/scalers/kubernetes-workload.md
@@ -13,8 +13,8 @@ triggers:
 - type: kubernetes-workload
   metadata:
     podSelector: app=backend
-    value: 0.5
-    activationValue: 3.1
+    value: "0.5"
+    activationValue: "3.1"
 ```
 
 **Parameter list:**
@@ -44,6 +44,6 @@ spec:
   triggers:
   - type: kubernetes-workload
     metadata:
-      podSelector: app=backend,deploy notin (critical,monolith)
-      value: 3
+      podSelector: 'app=backend, deploy notin (critical, monolith)'
+      value: "3"
 ```


content/docs/2.10/scalers/kubernetes-workload.md
```diff
@@ -12,24 +12,24 @@ go_file = "kubernetes_workload_scaler"
 triggers:
 - type: kubernetes-workload
   metadata:
-    podSelector: 'app=backend'
-    value: '0.5'
-    activationValue: '3.1'
+    podSelector: app=backend
+    value: 0.5
+    activationValue: 3.1
```
wozniakjan: this is not correct, yaml parser would defer types as `float` for these values and the `ScaledObject` will be rejected because trigger metadata are `map[string]string`. This has to be explicitly put in quotes to enforce `string` as a type

```suggestion
    value: '0.5'
    activationValue: '3.1'
```

content/docs/2.10/scalers/kubernetes-workload.md
```diff
@@ -44,6 +44,6 @@ spec:
   triggers:
   - type: kubernetes-workload
     metadata:
-      podSelector: 'app=backend, deploy notin (critical, monolith)'
-      value: '3'
+      podSelector: app=backend,deploy notin (critical,monolith)
+      value: 3
```
wozniakjan: same here, this would break the types
```suggestion
      value: '3'
```

content/docs/2.10/scalers/kubernetes-workload.md
isagarparmar: You're right, without quoting those values, the YAML parser interprets them as floats, which breaks the map[string]string requirement for trigger metadata. I appreciate you pointing this out. I've updated the PR to wrap the values in quotes to ensure they're treated as strings.

content/docs/2.10/scalers/kubernetes-workload.md
isagarparmar: Thanks again for the careful review!

PR #1663, opened 2025-12-03
Fix/kafka lagthreshold description

pmady:

### Checklist

Fixes #


content/docs/2.18/scalers/apache-kafka.md
```diff
@@ -47,7 +47,7 @@ triggers:
 - `bootstrapServers` - Comma separated list of Kafka brokers "hostname:port" to connect to for bootstrap.
 - `consumerGroup` - Name of the consumer group used for checking the offset on the topic and processing the related lag.
 - `topic` - Name of the topic on which processing the offset lag. (Optional, see note below)
-- `lagThreshold` - Target value for the total lag (sum of all partition lags) to trigger scaling actions. (Default: `10`, Optional)
+- `lagThreshold` - Average target value to trigger scaling actions. (Default: `10`, Optional)
```
JorTurFer, line 50: Hello
Could you elaborate this? Although AverageValue is the default metric type (set at scaler level), it can also be `Value`. 

Checking the current scaler code, it sums all the lags per partition and return the total value, so the HPA controller calculates avg or not depending on metric type

PR #1679, opened 2026-01-02
Add unreleased versions parameter for documentation

rickbrouwer:
Add unreleased versions parameter for documentation:

image

I've called it `unreleased` now, although I prefer `upcoming`.

Which do you prefer, `unreleased` or `upcoming`?

### Checklist

Fixes #https://github.com/kedacore/governance/issues/96


rickbrouwer:
Then we go for `unreleased`.

I'd prefer to keep them separate.
Otherwise, there's a chance, conversely, that someone might think the version has already been added and not post the next version. And then it suddenly shows up as unreleased.

Technically, it's possible, of course.
If you'd like, I could see if I can create a CI check, but that would be heading towards overkill.

PR #1683, opened 2026-01-06
mongodb: clarify `queryValue` documentation

VannTen:
The current documentation for queryValue (for the mongoDB scaler) does not really hint that it should be used at the scaling divisor

### Checklist

Fixes #

### DIFF
--- a/content/docs/2.16/scalers/mongodb.md
+++ b/content/docs/2.16/scalers/mongodb.md
@@ -64,7 +64,7 @@ The `mongodb` trigger always requires the following information:
 - `dbName` - Name of the database.
 - `collection` - Name of the collection.
 - `query` - A MongoDB query that should return single numeric value.
-- `queryValue` - A threshold that will define when scaling should occur.
+- `queryValue` - The divisor to apply to the query result to obtain the final number of pods / jobs.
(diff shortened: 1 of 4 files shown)


content/docs/2.16/scalers/mongodb.md
```diff
@@ -64,7 +64,7 @@ The `mongodb` trigger always requires the following information:
 - `dbName` - Name of the database.
 - `collection` - Name of the collection.
 - `query` - A MongoDB query that should return single numeric value.
-- `queryValue` - A threshold that will define when scaling should occur.
+- `queryValue` - The divisor to apply to the query result to obtain the final number of pods / jobs.
```
JorTurFer, line 67: This is not correct. `queryValue` is the target value for the HPA, but there are different algorithms on HPA Controller side and this change only applies when the metric type is AverageValue. 

I agree with you about the confusing description, reading it I'd not be 100% about what's this metric for, but maybe we can rephrase it like: 

`Target value for the HPA. How this value is used depends on the metric type` and maybe a link to metric type docs. WDYT?

VannTen: That would probably work, but I can't find the docs on metric type (Which are `AverageValue`, `Value` or `Utilisation` right ?)

PR #1698, opened 2026-01-31
Support empty string value for `paused-replicas` annotation

### LINKED ISSUE #7263: Support paused-replicas annotation as empty string value
### Proposal

The `paused-replicas` needs to be an integer otherwise it causes a `ScaledObjectCheckFailed`. The issue is when using ArgoCD in particular you may want to override this annotation in order to fix the replica scaling and then have it be indicated as "out of sync" by Argo.

This isn't possible if the field is not explicitly defined in the manifest and as `null` fields don't count, there is no way of supporting this. If we could treat empty string as "no value" then this would enable setting it in git and allowing Argo to notice the diff when changes are manually applied.

An alternative would be to support a value like `false` to make it more explicit than empty string.

### Use-Case

The Argo use case is the big one, but I could imagine there are other CD concepts that use the same idea of diffing manifests to determine if drift has occured.

### Is this a feature you are interested in implementing yourself?

Maybe

### Anything else?

_No response_

rickbrouwer:
Documentation related to pull request https://github.com/kedacore/keda/pull/7264

### Checklist

Fixes https://github.com/kedacore/keda/issues/7263

### DIFF
--- a/content/docs/2.19/concepts/scaling-deployments.md
+++ b/content/docs/2.19/concepts/scaling-deployments.md
@@ -45,12 +45,11 @@ This feature is not supported for `cpu`, `memory` or `cron` scaler.
 
 ## Pausing autoscaling
 
-It can be useful to instruct KEDA to pause the autoscaling of objects, to do to cluster maintenance or to avoid resource starvation by removing non-mission-critical workloads.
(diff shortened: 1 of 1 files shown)


rickbrouwer:
I am closing this PR. PR on the Keda side is closed due to stale.

PR #1733, opened 2026-04-11
Clarify URL format when using aggregateFromKubeServiceEndpoints

rickbrouwer:
The `url` parameter description only mentions that the port must be the pod port, but doesn't specify the required hostname format. When `aggregateFromKubeServiceEndpoints: true`, the URL must follow Kubernetes DNS format (`<service>.<namespace>[.svc.cluster.local]:<pod-port>/path`).
Without this, the scaler will fail to parse the service name and namespace from the hostname.

### Checklist

Related: https://github.com/kedacore/keda/pull/7642

### DIFF
--- a/content/docs/2.19/scalers/metrics-api.md
+++ b/content/docs/2.19/scalers/metrics-api.md
@@ -29,8 +29,9 @@ triggers:
 **Parameter list:**
 
 - `url` - Full URL of the API operation to call to get the metric value (eg. `http://app:1317/api/v1/stats`).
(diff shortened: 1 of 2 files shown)


wozniakjan (approved):
lgtm, this matches the implementation, although might be worth considering improving the UX and doing the port translation as part of the scaler?

rickbrouwer:
> although might be worth considering improving the UX and doing the port translation as part of the scaler?

Hmm, I don't know. I find it clear myself, and it might save some code that has to do some "magic." Or not?

PR #1760, opened 2026-05-06
fix: reduce unused whitespace and allow scrollable tables

linkvt:
The actual space for the docs is currently quite small due do the sidebar and a lot of unused whitespace. Font size is also quite large IMO but let's keep that at least for now.
In larger tables content is also truncated, see https://keda.sh/docs/2.19/reference/events/ .

For the reviewer: changes in layouts/partials/docs/hero.html are mostly whitespace, might make sense to review with the following: 
image

### Before and After

image

image

### Checklist

### DIFF
--- a/assets/sass/content.sass
+++ b/assets/sass/content.sass
@@ -1,3 +1,9 @@
+.doc-hero
+  margin-bottom: 1.5rem
+
+  .title sup
+    line-height: 0
+
 .content
   a sup
     margin-left: 0.1rem
@@ -7,6 +13,12 @@
 
   .highlight
     margin-bottom: 1rem
+
+  // Allow horizontal scrolling for larger tables
+  table
+    display: block
+    overflow-x: auto
+    max-width: 100%
 
   &.is-constrained
     max-width: 90ch
--- a/assets/sass/nav.sass
+++ b/assets/sass/nav.sass
@@ -10,7 +10,6 @@
   justify-content: space-between
 
 .nav
-  padding-top: 5rem
   padding-bottom: 2rem
 
   &-subsection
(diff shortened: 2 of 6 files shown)


JorTurFer:
I like this change!

JorTurFer (commented):
I have just tested how it looks, but in terms of code I can't say anything, sorry

linkvt:
Quick explanation of the changes that got no comment:
- layouts/partials/content.html
    - mostly whitespace changes, make sure to review with whitespace ignored as documented in the PR description
	- removed the section class, we apply this now on the article.html to not add padding to the content while ignoring the hero (the title and version selector)
		- basically: add whitespace around title+content, not individually
	- remove the div.container - it had no effect
- layouts/partials/article.html
	- uses -desktop classes for the columns that are only used on large screens (1/4 for nav, 3/4 for content), small screens (=named touch by bulma) only have a single column and get weird whitespace which is meant for the nav/content split only
	- adds section to the container to surround the inner content with whitespace (otherwise text would start at screen border)
- layouts/partials/docs/hero.html
	- mostly whitespace as above
	- removed two wrapping elements which didnt provide any value (no visual change before and after)
	- added `if` to not render the empty `<div class="field"...` which would automatically get whitespace above and below despite being empty (this is the whitespace above the version selector)
- assets/sass/style.sass
	- removing these styles lets us use the default values of bulma (CSS "framework") which are smaller -> less unused whitespace
- assets/sass/nav.sass
	- moves the navbar up, the top whitespace is now added via the section class in layouts/partials/article.html
- assets/sass/content.sass
	- margin-bottom creates the gap between version selector and the docs content
	- line-height: 0 solves a weird bug where the latest badge behind the docs title got pushed down causing a misalignment between the top of the navbar and the top of the docs content
	- table changes as commented

rickbrouwer (approved):

PR #1788, opened 2026-06-09
feat(clickhouse): add `clickhouse` scaler

izaakschroeder:
Adds docs for the new scaler.

Refs: https://github.com/kedacore/keda/pull/7850
Refs: https://github.com/kedacore/keda/issues/7418
Refs: https://github.com/kedacore/keda/pull/7404

### Checklist

### DIFF
--- a/content/docs/2.21/scalers/clickhouse.md
+++ b/content/docs/2.21/scalers/clickhouse.md
@@ -1,0 +1,186 @@
++++
+title = "ClickHouse"
+availability = "v2.21+"
+maintainer = "Community"
+category = "Data & Storage"
+description = "Scale applications based on ClickHouse query results."
+go_file = "clickhouse_scaler"
++++
+
+### Trigger Specification
+
+This specification describes the `clickhouse` trigger that scales based on the results of a ClickHouse query.
+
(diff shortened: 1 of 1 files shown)


rickbrouwer:
This one is still draft, but here are some points to get you started:

Should we clarify that host is required unless connectionString is set? It is listed under Trigger Specification without an “Optional” marker or default, unlike the surrounding fields, which makes its conditional nature a bit unclear.

The connection pool options (maxIdleConns, maxOpenConns, connMaxLifetime) appear twice, once under Trigger Specification and once under Authentication Parameters. Is the duplication intentional, or could one of the two be removed?

Would it be worth adding a short TLS example? The tls/cert/key/ca parameters are listed, but there is no YAML snippet showing them in use, unlike the connection-string and password flows.

In the individual-parameters example, the host is clickhouse-svc.namespace.cluster.local. Should that be ...svc.cluster.local to match a Kubernetes service FQDN (and the e2e test)?

rickbrouwer:
Could you take a look at my comments @izaakschroeder? The code has already been merged, but the documentation is still missing. Also, this PR is still a draft.

PR #1797, opened 2026-06-30
blog: add Autoscaling Wasm on K8s with KEDA and wasmCloud

ericgregory:
This PR adds a blog post on autoscaling WebAssembly deployments on Kubernetes with KEDA and wasmCloud.


ericgregory:
Bumping as not stale.

PR #1798, opened 2026-06-30
docs: add wasmCloud integration page to 2.20 and 2.21

ericgregory:
Adds wasmCloud integration pages to 2.20 and 2.21 docs versions.


ericgregory:
@JorTurFer Those kind links are updated -- just let me know if you need any other changes here!

PR #1809, opened 2026-07-11
docs: fix broken links in scaler docs

mailnike:
Several documentation links return HTTP 404. This updates each to its current working location:

- `content/docs/2.21/scalers/kubernetes-workload.md` (1 link)
- `content/docs/2.21/scalers/rabbitmq-queue.md` (1 link)
- `content/docs/2.21/scalers/aws-kinesis.md` (1 link)

Docs-only change. Each replacement URL was verified to return HTTP 200.

### DIFF
--- a/content/docs/2.21/scalers/aws-kinesis.md
+++ b/content/docs/2.21/scalers/aws-kinesis.md
@@ -40,7 +40,7 @@ triggers:
 
 ### Authentication Parameters
 
-You can use `TriggerAuthentication` CRD to configure the authenticate by providing either a role ARN or a set of IAM credentials, or use other [KEDA supported authentication methods](https://keda.sh/concepts/authentication).
+You can use `TriggerAuthentication` CRD to configure the authenticate by providing either a role ARN or a set of IAM credentials, or use other [KEDA supported authentication methods](https://keda.sh/docs/2.21/concepts/authentication).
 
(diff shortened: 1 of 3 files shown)


rickbrouwer:
Thanks!

What do you think, would it be better to go to `latest` instead of the mentioned versions? That way, it will at least always find the latest version, and we won't have to keep adjusting that in the future.

rickbrouwer:
By the way, you only modified version 2.21. Could you include the other versions as well?

mailnike:
Thanks @rickbrouwer, good call on both. I've:

- switched the keda.sh links to `/docs/latest/` (and fixed the `oauth` -> `oauth2` slug), and
- used the version-agnostic `kubernetes-api/workload-resources/pod-v1/#PodStatus` reference for the Kubernetes link instead of a pinned API version,

so they won't go stale, and applied the fixes across every doc version that contained each broken link (aws-kinesis: 2.4-2.21, rabbitmq oauth: 2.20-2.21, kubernetes-workload: 2.6-2.21). Let me know if you'd like anything scoped differently.

mailnike:
Sorry, I accidentally closed this while updating the branch. Reopened as #1810 with your feedback incorporated (using `/docs/latest/` and covering all affected versions). Continuing there.

PR #1810, opened 2026-07-16
docs: fix broken links in scaler docs across all versions

mailnike:
### Problem

Several scaler docs contain dead links. This fixes them using version-agnostic targets so they will not go stale, applied across all doc versions that contain each broken link:

- `keda.sh` links now point at `/docs/latest/` (and the `oauth` -> `oauth2` slug is corrected)
- the Kubernetes `PodStatus` reference now uses the stable `kubernetes-api/workload-resources/pod-v1/#PodStatus` page instead of a pinned API version

Affected: `aws-kinesis` (2.4-2.21), `rabbitmq-queue` (2.20-2.21), `kubernetes-workload` (2.6-2.21).

> This supersedes #1809, which I accidentally closed while updating the branch. It incorporates @rickbrouwer's review feedback there (use `latest`, and cover all versions, not just 2.21).

### DIFF
--- a/content/docs/2.10/scalers/aws-kinesis.md
+++ b/content/docs/2.10/scalers/aws-kinesis.md
@@ -40,7 +40,7 @@ triggers:
 
 > These parameters are relevant only when `identityOwner` is set to `pod`.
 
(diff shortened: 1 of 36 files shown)


rickbrouwer:
Thanks!

In your previous pull request, you also fixed a link in the rabbitmq scaler. Could you include that in this pull request as well?

JorTurFer (approved):
thanks for the fix!

PR #1811, opened 2026-07-19
docs: document valid ranges for liiklus lagThreshold and activationLagThreshold

### LINKED ISSUE #7917: liiklus scaler: integer divide-by-zero panic when lagThreshold is 0
### Report

The liiklus scaler panics with `integer divide by zero` when a liiklus trigger sets `lagThreshold: "0"`.

This is the same class of bug as #3366, which was fixed for the Kafka scaler by adding a `Validate()` that rejects a non-positive `lagThreshold`. The liiklus scaler was never given the equivalent guard. `liiklusMetadata` has no `Validate()` method, so the value is used unchecked in `GetMetricsAndActivity` (`pkg/scalers/liiklus_scaler.go`):

```go
if totalLag/uint64(s.metadata.LagThreshold) > uint64(len(lags)) {
```

When `LagThreshold` is `0`, `uint64(0)` is the divisor and the metric evaluation panics. The field default of `10` only applies when the key is absent, so an explicit `"0"` reaches the division. The same value also feeds an unchecked `uint64` multiply on the next line, and a negative `activationLagThreshold` becomes a huge `uint64` that keeps the scaler from ever activating.

The kafka and apache-kafka scalers already reject a non-positive `lagThreshold` and a negative `activationLagThreshold` in their `Validate()`.

### Expected Behavior

A liiklus trigger with `lagThreshold: "0"` (or a negative value) is rejected at parse time with a clear error, the same as the kafka scaler, instead of panicking.

### Actual Behavior

The trigger is accepted and the liiklus scaler panics with `integer divide by zero` on the metric evaluation.

### Steps to Reproduce the Problem

A brand-new ScaledObject may not reach the scaler, because the HPA it creates can be rejected first for a non-positive target. Two reliable paths:

1. Create a ScaledJob with a liiklus trigger and `lagThreshold: "0"`. ScaledJobs read scaler metrics directly, with no HPA target validation in front, so the metric evaluation divides by zero and panics. This matches the original #3366 stack, which came through the ScaledJob path.
2. Or create a valid liiklus ScaledObject first, let the scale loop start, then update `lagThreshold` to `"0"`. The generation change rebuilds the scaler with the new metadata and the next poll hits the divide-by-zero.

The scaler cache calls `GetMetricsAndActivity` with no panic recovery, so the panic propagates.

### KEDA Version

main

### Scaler Details

liiklus

thc1006:
Documents the valid ranges for the liiklus scaler thresholds, matching the parse-time validation added in kedacore/keda#7918 (fixes kedacore/keda#7917): `lagThreshold` must be greater than `0` and `activationLagThreshold` must be greater than or equal to `0`.

### DIFF
--- a/content/docs/2.21/scalers/liiklus-topic.md
+++ b/content/docs/2.21/scalers/liiklus-topic.md
@@ -31,8 +31,8 @@ triggers:
 - `address` - Address of the gRPC liiklus API endpoint.
 - `group` - Name of consumer group.
 - `topic` - Topic to monitor and scale based on `lagThreshold`.
-- `lagThreshold` - Value to trigger scaling actions for. (Default: `10`, Optional)
-- `activationLagThreshold` - Target value for activating the scaler. Learn more about activation [here](./../concepts/scaling-deployments.md#activating-and-scaling-thresholds). (Default: `0`, Optional)
+- `lagThreshold` - Value to trigger scaling actions for. Must be greater than `0`. (Default: `10`, Optional)
+- `activationLagThreshold` - Target value for activating the scaler. Must be greater than or equal to `0`. Learn more about activation [here](./../concepts/scaling-deployments.md#activating-and-scaling-thresholds). (Default: `0`, Optional)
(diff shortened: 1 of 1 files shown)


rickbrouwer:
Because we are deprecating this scaler (https://github.com/kedacore/keda/issues/7929) and will remove it later, we are stopping development on it.

PR #1831, opened 2026-07-29
docs(http-add-on): note queue tick duration also bounds /queue fetches

Fedosin:
## Summary
- Document that `KEDA_HTTP_QUEUE_TICK_DURATION` also bounds each scaler request to an interceptor pod's `/queue` endpoint (clamped to a minimum of `250ms`), matching [kedacore/http-add-on#1729](https://github.com/kedacore/http-add-on/pull/1729).
- Scoped to the unreleased `0.16` docs; this behavior is not in v0.15.0.

## Test plan
- [x] Confirm the updated description in [Environment Variables](https://keda.sh/docs/http-add-on/0.16/reference/environment-variables/) renders correctly in the table.
- [x] Spot-check that `0.14`/`0.15` entries are unchanged.


linkvt (approved):
LGTM

PR #1832, opened 2026-07-29
docs(http-add-on): document direct pod routing default

Fedosin:
## Summary
- Document `KEDA_HTTP_DIRECT_POD_ROUTING` (default `true`) in the 0.16 environment variables reference and Configure Interceptor guide, matching [kedacore/http-add-on#1697](https://github.com/kedacore/http-add-on/pull/1697) / [#1585](https://github.com/kedacore/http-add-on/pull/1585).
- Correct architecture text that still described forwarding only through the Service ClusterIP.
- Scoped to the unreleased `0.16` docs; this feature is not in v0.15.0.

## Test plan
- [x] Confirm the new env var row renders in [Environment Variables](https://keda.sh/docs/http-add-on/0.16/reference/environment-variables/).
- [x] Confirm the Direct pod routing section and helm example in [Configure the Interceptor](https://keda.sh/docs/http-add-on/0.16/operations/configure-interceptor/).
- [x] Confirm the architecture interceptor section no longer claims ClusterIP-only forwarding.
- [x] Spot-check that `0.14`/`0.15` pages are unchanged.


linkvt (approved):
LGTM

PR #1863, opened 2026-09-08
docs: fix broken scaler schema links on the Schema page

PragalvaXFREZ:
The Schema page points at two files in the keda repository that do not exist. Both links 404 today:

```
$ curl -sIL -o /dev/null -w '%{http_code}\n' https://github.com/kedacore/keda/blob/main/schema/generated/scalers-metadata-schema.yaml
404
$ curl -sIL -o /dev/null -w '%{http_code}\n' https://github.com/kedacore/keda/blob/main/schema/generated/scalers-schema.yaml
200
```

The generated schema directory contains exactly two files:

```
$ ls schema/generated/
scalers-schema.json
scalers-schema.yaml
```

This corrects both the URL and the link text. Applied across 2.18, 2.19, 2.20 and 2.21, which are all the versions whose Schema page carries the stale name. The four files are byte-identical, so it is the same one-line change in each.

`.htmltest.yml` sets `CheckExternal: false`, so the link-check workflow never resolved these URLs, which is why they went stale unnoticed.

Live today at https://keda.sh/docs/2.20/operate/schema/

### DIFF
--- a/content/docs/2.18/operate/schema.md
+++ b/content/docs/2.18/operate/schema.md
@@ -6,7 +6,7 @@ weight = 100
 
 ## Scaler Schema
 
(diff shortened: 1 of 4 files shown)


JorTurFer (approved):

JorTurFer:
nice!

PR #1871, opened 2026-09-13
docs for cache-miss-to-direct-client

dttung2905:
Related to this https://github.com/kedacore/charts/pull/902. Let me know if we want to make changes to past versions too (v2.16 +) or just ship with later version

### Checklist


JorTurFer:
I guess that at least 2-3 versions should be covered or even all where this is supported

